PR #10, opened 2026-07-02
One box: all generated UI renders in the sandbox

yousefh409:
## One box: all generated UI renders in the sandbox

Consolidates the render surface to a single output invariant: **everything the AI generates renders in the egress-jailed sandbox, through one tool (`render_view`).** Pre-built components are building blocks used *inside* a generated view, not a separate output path. Removes the two paths that bypassed the box.

Stacked on #8 (Tier 2.5) — this PR is the follow-up simplification we discussed: one box, not a pile of render paths.

### Before → after
| | Before | After |
|---|---|---|
| Single pre-built component | rendered **in-process** (`render_ui`) | a one-node **generated view** in the sandbox |
| Composed / novel UI | generated view in the sandbox | same |
| Arbitrary interactive app (e.g. dino game) | **raw HTML** in a *loose* iframe (`HtmlApp`, could phone home) | a **novel generated component** in the *jailed* sandbox |
| Connect (OAuth) card | `render_ui name:"Connect"` | `request_connect` (host-rendered; the one principled exception) |

### What changed
- **`render_ui` removed** — `render_view` is the only UI-content tool. Engine/barrel/shell updated; `examples/basic` + default instructions de-referenced.
- **`request_connect` added** — the Connect OAuth card can't run in the sandbox (needs host popup/chat privileges), so it gets its own narrow, explicit affordance. It emits the exact node the existing host-side Connect branch renders.
- **`render-node` collapsed** to two routes: Connect (host) and `kind:"generated"` (sandbox). The in-process component renderer and the raw-HTML `App` branch are gone; **`HtmlApp.tsx` deleted**.
- **Demo prompt rewritten** to the single-tool model: emit one `render_view` payload; use catalog components + layout primitives as blocks; write novel generated components (React.createElement, network-jailed, `props.flowlet.dispatch`) for anything custom incl. games; `request_connect` for connecting toolkits. Raw-HTML/`App` guidance removed.
- **Cleanup**: dead `render_ui` policy entry removed; stale comments fixed.

### Why (the point)
The old design had three render paths, two of them running AI content — one in a *loose* box, one in a *tight* box — for no principled reason, just history. This makes the model match the mental model: **one locked box for all generated UI; the trusted catalog is vocabulary inside it; OAuth Connect is the sole host exception because it genuinely needs host powers the box denies.** It also closes the last unjailed egress path (the raw-HTML iframe) by removing it entirely.

### Verification
- `pnpm build` 10/10; `pnpm test` green except the pre-existing unrelated `orders.test.ts`; browser gates **22 passed / 1 skipped / 0 failed**.
- Visual proof: a pre-built host `Card` renders **inside the sandbox** via a generated view (previously it would have rendered in-process).
- Final holistic review confirmed the invariant is enforced end-to-end: only `render_view` + `request_connect` emit UI; render-node routes only box + Connect; grep confirms no bypass path survives.

### Follow-up (out of scope)
Host-side JSX compilation so novel components author as naturally as raw HTML did — the one ergonomic thing the removed HTML path had over generated components.

https://claude.ai/code/session_01HMRHgPTUVXatBaVdj4Ak2F


yousefh409:
Landed in main via #11 (the stack was merged as one integration after resolving the main-divergence conflict at the tip). All one-box commits are in main as of ce348e0c.

PR #14, opened 2026-07-02
Contracts freeze: manifest schema + five runtime seams

yousefh409:
Implements sequencing step 2 of the [platform architecture](docs/superpowers/specs/2026-07-01-flowlet-platform-architecture-design.md) (Decisions 1, 3, 6): the frozen contracts the parallel tracks (ENG-197/198/202/188) build against. **Additive only** — no refactors, no runtime carve-out, `flowlet-agent` untouched (ENG-202 works there in parallel).

## What's frozen

**Manifest schema** (`packages/flowlet-core/src/manifest/`, docs in [docs/contracts/manifest.md](docs/contracts/manifest.md)):
- `theme.json` — structurally identical to `BrandTokens` v1; a conformance test in flowlet-components fails CI on divergence
- `tools.json` — tool descriptors with **required** `{mutating, dangerous}` annotations (MCP hint mapping documented) and a minimal `http` binding (discriminated union = extension point), plus host event declarations for automation triggers
- component entries (JSON image of `PrewiredDescriptor`), the published-manifest unit, and `ManifestRef` (immutable registry rows; sessions bind to a ref)
- Language-neutral JSON Schema artifacts in `packages/flowlet-core/schemas/`, generated from the zod source (`pnpm generate:schemas`); a sync test fails CI on drift, ajv-validated

**Five seam interfaces** (`packages/flowlet-core/src/seams/`, docs in [docs/contracts/seams.md](docs/contracts/seams.md)): Store, CredentialBroker, Executor, Scheduler, Channels — embedded-vs-cloud mapping from the architecture table captured in TSDoc, plus an in-memory stub test (the embedded/CI guarantee in miniature).

## Deliberately deferred

- Automation `spec` is `unknown` at the Store seam — ENG-188 owns the DSL (its JSONata step-graph decisions are intentionally NOT imported here)
- No `memory` member on Store until ENG-189 (additive later)
- Realtime voice excluded from `Channels.deliver` — its own contract at ENG-185
- Query/body parameter mapping in http bindings — convention until a real case demands it

## Review trail

All 13 open contract points were reviewed mid-flight by Yousef and locked (every recommendation accepted): [docs/superpowers/specs/2026-07-01-contracts-freeze-open-questions.md](docs/superpowers/specs/2026-07-01-contracts-freeze-open-questions.md).

## Verification

[...]

PR #15, opened 2026-07-02
ENG-183: Persistent & saved flowlets — persistence layer

yousefh409:
## What

Phase 1 of ENG-183 (persistence layer only — the library UI surface is paused at the UI gate per the standing rules):

- **Bound data queries** (`@flowlet/core` + `@flowlet/agent`): `GeneratedPayload.queries` declares which policy-governed tool call produced each `data` subtree (`{path, tool, input}`, validated, capped at 16). `render_view` accepts and ships them; the demo agent's prompt teaches the contract (tool result verbatim at the path, reshape at render time).
- **Extended `Flowlet` record** (`@flowlet/shell` store seam): `prompt` (originating user prompt), `pinned`, `createdAt` — additive, matching architecture Decision 6 (tree + bound query + prompt + name/pin).
- **Real embedded store**: `createWebStorage` — a `FlowletStore` over localStorage (namespaced, injectable Storage/clock, loud on quota/unavailable, newest-first list). demo-bank now passes it, so the loud in-memory fallback warning is gone.
- **Reopen flow**: new `RunQuery` host seam + `refreshFlowletNode`/`useReopenFlowlet`. Reopening renders the snapshot instantly, re-runs the declared queries through the normal tool path (same policy), patches results in via `applyPointerPatch`, and the stage's existing data-delta path updates in place. Fully-live refreshes write back to the store; failed queries keep the snapshot (Linear default: live re-run with graceful fallback).
- **demo-bank wiring**: `runQuery` is one fetch to the existing policy-governed `/api/flowlet/action` route (zero new server surface); the `/flowlet` page tabs now persist through the store (and the auto-save filter now includes `generated` nodes — the old `component`-only filter saved nothing post-ENG-200). No visual changes.

## Pre-existing issues surfaced (fixed here, flagged for review)

- **SandboxStage hung at "loading" forever in dev**: the mounted-ref guard never reset under StrictMode's mount→cleanup→mount, so `loadSources()` resolved into a dead closure. Reproduces on main.
- **Built React shim dropped `export * from "react"`** (CJS interop), so the components bundle's named imports (`PureComponent`, `createPortal` via recharts) failed to link on any freshly built checkout. Named exports are now explicit. Reproduces on main.
- **Lint**: `public/flowlet/**` build artifacts are now eslint-ignored (lint failed on any built checkout). Note: 4 unrelated `react-hooks` lint errors remain on main in untouched files (`transactions/*`, `FlowletPoller`, `filters-bar`) — reported, not fixed here.

## Open design points for Yousef (also in the spec)

1. `queries` lives inside `GeneratedPayload` (self-contained artifact, additive to genui/v1) — vs. keeping provenance only on the `Flowlet` record.
2. localStorage (not IndexedDB) as the embedded reference store.
3. Refreshed data silently writes back to the store on a fully-live reopen.
4. ⚠️ Reopen executes whatever tool a saved payload declares — policy is the only guard, and the demo policy allows all in-process tools (verified: a saved view declaring `set_rule` as a query executed it on reopen). Suggest constraining reopen to read-annotated tools once ENG-202 descriptors land.

## Verification

- `pnpm typecheck` / `pnpm test` (13 tasks) green; new unit coverage in core/agent/shell/demo-bank.
- Browser-verified via `pnpm demo` + Playwright: saved tab survives reload; opening it fires POST `/api/flowlet/action` re-running `get_transactions`; the pane renders fresh merchants replacing seeded stale values; write-back bumps `updatedAt`; a failing query (unknown tool) keeps the snapshot untouched with no crash. Screenshots delivered to Yousef via the session (gh can't attach images); happy to re-drop them here.
- Not exercised live: the save-on-chat-turn path (Anthropic API credits exhausted mid-verification) — covered by unit tests on `deriveSavedDrafts`; the reopen path was driven with a store record byte-identical to what a save writes.

Spec: `docs/superpowers/specs/2026-07-01-flowlet-eng183-saved-flowlets-design.md` · Plan: `docs/superpowers/plans/2026-07-01-eng183-saved-flowlets.md`

Per standing rules: **do not merge — awaiting Yousef.** Library UX proposal follows separately at the UI gate.

https://claude.ai/code/session_01NMVoBXj41pmA9aP3tFx1rz


yousefh409:

**All 8 findings accepted and fixed in 79ec18a4 — none rejected.** Branch rebased onto main (post-#14 contracts freeze).

| Finding | Verdict | Fix |
|---|---|---|
| [P1, consensus] node id collision across requests | **Real** — confirmed `createRenderViewTool` is built per `run()`; live repro: today's session saved `view-1` again | ids now `view-N-<uuid8>`; test covers two tool instances |
| [P1, Codex] replay unconstrained to read-only | **Real** — accepted with Yousef's reads-only ruling; pulled forward from the follow-up | explicit allowlist in demo `runQuery` (throws before network), reads-only contract documented on the `RunQuery` seam; manifest `{mutating}` annotations are the upgrade path when ENG-202 merges |
| [P2, Codex] `refreshing` sticks true after cancelled refresh | **Real** | query-less opens reset it; regression test |
| [P2, Codex] root `''` query can corrupt `payload.data` | **Real** | non-object root results count as failures (snapshot kept); test |
| [P2, Opus] unhandled rejection in page save pipeline | **Real** | `.catch` + loud log; next items-change retries |
| [P2, Opus] write-back reverts rename/pin | **Real** — must-fix given approved library UI | write-back load-merges the current record; also skips deleted records (no resurrection); test covers rename-during-refresh and delete-during-refresh |
| [P2, Opus] no schema version on records | **Real** | `{v:1, record}` envelope; unknown versions skipped with warning, legacy bare records still read; test |
| [P2, Opus] shim exports `findDOMNode` (React 19 removed) | **Real** — nothing imports it (checked the built components bundle) | dropped; tracked shim artifact rebuilt to match source (it was stale in the earlier commit) |

Also since review: live save-on-chat-turn verified with real credits — the model declared `queries: [{path: "/transactions", tool: "get_transactions", input: {limit: 40}}]` unprompted and the view auto-saved with its prompt. Full suite + typecheck green post-rebase.

Merging per the new pipeline (dual review + triage complete).

https://claude.ai/code/session_01NMVoBXj41pmA9aP3tFx1rz

PR #16, opened 2026-07-02
ENG-202 (embedded): Their API as the agent's tools — act as the user

yousefh409:
## Summary

The embedded half of [ENG-202](https://linear.app/runvendo/issue/ENG-202/their-apiclimcp-as-the-agents-tools-act-as-the-user), per the locked platform architecture (Decisions 2 & 4, topology B): the host company's own REST API becomes the agent's toolset, gated by the existing policy layer + approval cards, with every interactive call executing **in the user's browser on their existing session**. User credentials never leave the browser; the loop only sees tool results.

- **`@flowlet/core`** — `openApiToHostTools` (OpenAPI 3.x → host tool definitions with mutating/dangerous annotations, flat JSON-Schema inputs, HTTP call metadata) and `executeHostToolCall` (the browser call executor; HTTP errors are structured data for the model).
- **`@flowlet/agent`** (additive only) — `hostToolset` feeds definitions through the previously-unexercised **caller seam** (`RunInput.tools`); `wrapClientTool` enforces policy at `needsApproval` for no-execute client tools (deny fails closed by erroring the turn); `buildToolset` routes marked tools there and still fails closed on unmarked no-execute tools. Engine untouched.
- **`@flowlet/react`** — `hostTools` on `FlowletProvider`: a runner executes settled host tool calls once each (un-gated, or after the user answers the existing approval card) and a host-aware `sendAutomaticallyWhen` holds auto-resubmission until approved host tools carry outputs.
- **demo-bank (testbed)** — a real OpenAPI 3.1 spec for its existing API (16 reads + the `createOrder` write), kept honest by a spec↔route-files sync test; policy decides client-executed tools by annotations; both sides consume one derivation of the spec.

Key SDK finding (probed against ai@6.0.28, drove the design): the tool call streams **before** its approval request, so an `onToolCall`-based executor would bypass the gate — the runner keys off settled tool-part state instead. Full notes in [`docs/superpowers/specs/2026-07-02-eng-202-embedded-host-tools-findings.md`](docs/superpowers/specs/2026-07-02-eng-202-embedded-host-tools-findings.md).

## Live end-to-end (verified in a real browser)

1. **Read, auto-allowed:** "How much money is in my accounts?" → `listAccounts` runs from the browser (network log: `GET /api/accounts`), no card.
2. **Mutation, gated → approved:** DoorDash order → approval card (existing UI) → Approve → browser issues `POST /api/orders` → transaction lands in the store (`DOORDASH*TACOS`, −$32.00).
3. **Mutation, declined:** second order — card shown even though the user said "no need to confirm"; Decline → no `POST /api/orders`, agent acknowledges the block.

| Read (no card) | Approval card | Executed | Declined |
|---|---|---|---|
| read | card | executed | declined |

## Verification

- `turbo run test --force`: 13/13 packages green (61 new tests across core/agent/react/demo-bank).
- `pnpm typecheck`: 15/15 green. `pnpm build`: green.
- `pnpm lint`: fails **on a clean tree too** — pre-existing `react-hooks` errors in the generated `public/flowlet/components-sandbox.js` bundle (documented in the findings note, not touched here).

## Out of scope / known quirks

- Duplicate activity chips after a client-tool round-trip (pre-existing thread-grouping behavior, cosmetic).
- No new UI was needed — the existing ApprovalCard and activity chips cover host tools.
- Cloud port (ENG-198 merge), CLI/SDK/MCP adapters, and token-exchange for automations are the other halves of ENG-202, not this PR.

https://claude.ai/code/session_016VFZDnpmj3sLfJumbSe5An


yousefh409:

Rebased onto main (contracts freeze #14 in) and verified conformance with the frozen `httpBindingSchema` — the executor and adapter now enforce the same host-relative path regex.

| Finding | Verdict | Action |
|---|---|---|
| **P1 (consensus)** `provider.tsx` — `hostToolNames` keyed on `hostTools` object identity; inline literal from FlowletRoot rebuilds the Chat every parent render, wiping message/approval state mid-turn | **Real — fixed** | Memoized on `hostTools?.definitions`; regression test renders the provider 3× with fresh inline config objects and asserts one Chat instance |
| **Codex P2** executor validates only path params; a required write body silently omitted executes host defaults | **Real — fixed** | Schema-driven guard: any name in `inputSchema.required` (path/query/body uniformly) missing ⇒ throw before fetch, with tests |
| **Codex P2** path-item-level `parameters` ignored | **Real — fixed** | Path-level params merged into every operation, operation-level wins on (name, in), with tests. (Demo-bank's spec declares params at operation level only, which is why the live e2e worked.) |
| **Opus P2** double-exec guard is in-memory only | **Real — documented loudly** (per reviewer's own either/or) | Findings note now has a hard constraint: no persistence/resume of client-tool threads before an idempotency key ships. Sound today: `toolCallId`s are unique per call and a reload drops the in-memory thread entirely |
| **Opus P2** side-effecting-GET trust assumption | **Real — documented** | Spec-authoring requirement (`x-flowlet-dangerous: true` override) recorded in the findings note + adapter docstring; extractor lint suggested for ENG-197 |
| **Opus P2** deny path unit-tested, not e2e-verified | **Accepted, no code** | Already flagged in the findings note; deny is unreachable with the annotation policy and stays fail-closed by construction |

Nothing rejected — all six findings were real; four fixed in code, two documented as binding constraints.

**Verification after rebase + fixes:** `turbo run test --force` 13/13, `typecheck` 15/15, `build` 10/10 (uncached).

https://claude.ai/code/session_016VFZDnpmj3sLfJumbSe5An

PR #17, opened 2026-07-02
ENG-184: fix the broken + unstyled tiers of generated UI (audit + P0/P1 fixes)

yousefh409:
## What

Phase 1 of ENG-184 (brand-native UI surface): the full audit (`audit/FINDINGS.md` + evidence screenshots), then the Yousef-approved fixes for the **broken** and **unstyled** tiers only. The brand tier (prompt guidance, chart palette, primitives, host components, catalog expansion) is deliberately untouched — Yousef drives it next.

## Fixes (each TDD'd: gate red → fix → green)

1. **React shim named exports** *(applied verbatim from PR #15 with attribution — they hit the same bug)*: `export * from "react"` was dropped by the CJS interop, so the real components bundle failed at module link (`PureComponent`) and **every render_view rendered a blank stage**.
2. **New `gate-real-bundle`**: browser gates now build + load the REAL `@flowlet/components` sandbox bundle (a test:browser pre-step), so this class of bug can't hide behind sample bundles again.
3. **Stage auto-size**: the runtime's `resize` messages are now consumed in `createStage`; previously nothing consumed them and views clipped at the UA default iframe height. The test fixture's compensating listener is removed so `gate-autosize` pins the production path.
4. **OpenUI base CSS ships in the sandbox bundle** (`?inline` + injected `<style>`): catalog components rendered as bare HTML before; the iframe cannot load external stylesheets (CSP `default-src 'none'`).
5. **Srcdoc baseline styles**: the sandbox document now consumes the injected `--flowlet-*` vars (font/fg, margin reset). Bare generated markup no longer renders Times-New-Roman-1996. New `gate-baseline-styles` + unit test.
6. **Concrete brand font token**: `mapleBrand.fontFamily` contained `var(--font-inter)`, which can't resolve in the iframe and invalidated the whole declaration. Now a concrete stack; the host keeps real Inter via a host-scope delivery override in `FlowletRoot` (next/font's mangled family name only exists there).
7. *(from PR #15, verbatim)*: SandboxStage StrictMode mount-guard hang fix + eslint-ignore for generated `public/flowlet` bundles — both needed to run and verify live.

## Verification

- `pnpm typecheck`, `pnpm test`, and the full stage browser suite green (30 gates, incl. the live-LLM e2e gate, now passing with the topped-up key).
- **Live model-driven runs** (screenshots in `audit/`): `after-live-dashboard.png` — a real render_view dashboard (chart + table) renders styled and auto-sized in the Vendo thread; `after-live-novel-widget.png` — a Tier 2.5 model-written component renders. Before/after harness pairs: `05/06/09` vs `after-*`.

## Coordination

- Three hunks applied **verbatim from PR #15** (ENG-183) with attribution in each commit message — identical text merges cleanly whichever branch lands first.
- BrandTokens schema NOT changed (frozen-adjacent): recommend tightening `fontFamily` to reject `var(` — flagged for the orchestrator/ENG-197.
- Not fixed, reported: shell markdown renders `$` amounts as LaTeX math (flowlet-shell); explicit "show me" prompts sometimes answer text-only without render_view (F7/brand tier); 5 pre-existing lint errors in untouched demo-bank files (react-hooks v6 rules).

## What Yousef reviews next (brand tier, not in this PR)

F7 prompt brand guidance · F8 chart palette/responsiveness · F9 richer primitives · F10 host-component registration · F11 catalog gaps — per `audit/FINDINGS.md`.

https://claude.ai/code/session_011zQZV3DF23MHVTHqkmqMhm


yousefh409:
Review triage:

- **Fixed (consensus P1):** resize heights clamped — finite/positive validation, host-owned max (`opts.maxStageHeight`, default 8192), rAF-coalesced applies. New gate 6b fires 1e9/-50/Infinity from a real generated component and asserts the iframe never exceeds the max. (c7354f76)
- **Fixed (consensus P2):** `createStage` returns `dispose()` (listener + pending frame); stage-adapter calls it on unmount. Unit test pins the wiring. (c7354f76)
- **Not fixed, acknowledged (Opus informational, pre-existing):** `runtime.ts` `injectTheme` string-concatenates theme values into CSS unescaped, and `BrandTokens.fontFamily` is a free string — a value containing `}` escapes into arbitrary CSS. Host-trusted input today, not sandbox-exploitable; orchestrator is routing a hardening requirement to ENG-197 (extractor rejects `var(`/`}` in fontFamily) with possible schema tightening later.

Full verify after fixes: typecheck, all package tests, 31 browser gates green (incl. the live-LLM e2e earlier).

PR #19, opened 2026-07-02
ENG-183 phase 2: saved-flowlet library, reads-only live refresh, undo delete

yousefh409:
## What

Phase 2 of ENG-183, built to Yousef's UX-gate decisions and **approved as-is from screenshots** (toast stays 6s):

- **Library on the new-tab page** (placement A): the empty thread state is now the library — greeting, hero composer on top, then pinned cards first and recent after. `FlowGallery` becomes a card grid (name, prompt subtitle, quiet mono `updated …`), with hover actions: pin/unpin, inline rename (Enter commits, Esc cancels), delete.
- **Delete = undo toast** (no confirm): new `FlowletToast`; undo restores the exact record (original timestamps). 6s window.
- **Reads-only reopen** (pulled into #15's triage, contract documented on the `RunQuery` seam): mutating/unknown tools are refused before any network call → snapshot + note. Manifest `{mutating}` annotations (frozen in #14) are the upgrade path once ENG-202 lands.
- **Live refresh while open**: `useReopenFlowlet` re-runs the declared reads-only queries on a host-configurable cadence (`refreshIntervalMs` on the shell provider, default 60s, `0` disables) — only while the document is visible, dropping data-equal ticks (no stage churn, no `updatedAt` creep), stopping after 3 consecutive full failures. No websockets/event infra (host-event-driven refresh comes later with automations).
- **Stale-data note**: quiet mono `showing saved data — live refresh unavailable` above a reopened view when any query couldn't re-run.

## Screenshots (reviewed + approved by Yousef)

| | |
|---|---|
| **Empty new-tab page** — hero composer on top | **Populated library** — pinned first, then recent |
| empty | populated |
| **Inline rename + undo toast** | **Stale note after reads-only refusal** |
| rename-undo | stale |

(The four PNGs are force-added at the repo root — root `*.png` is gitignored — because the body needs hosted images; happy to drop the commit post-review if unwanted.)

## Verification

- Full suite + typecheck green (15 turbo tasks); new tests: FlowGallery interactions, FlowletToast, live-refresh tick/pause-hidden/stop-on-failures, write-back rename/delete races.
- Browser-verified on `pnpm demo`: real agent turn auto-saved with model-declared `queries`; rename persisted across card/tab/store; delete→undo restored record + card; `set_rule`-declared query refused with zero execution and the stale note shown; **live-refresh tick observed at exactly t+60s** in the resource timeline with the data-equal result dropped.

Per the pipeline: dual review (fresh Codex + Opus) runs next; I'll triage findings and self-merge after.

https://claude.ai/code/session_01NMVoBXj41pmA9aP3tFx1rz


yousefh409:

**All 6 findings accepted — fixed coherently in fe1170a6 around the shared root cause (state read from stale sources instead of the store / last-good node). None rejected.**

| Finding | Verdict | Fix |
|---|---|---|
| [Opus P1] single-slot undo buffer orphans a second delete | **Real** | deletes queue FIFO; one toast per view (keyed by id) with its own full 6s window; browser-verified with a rapid double delete → two sequential toasts, both undos restored |
| [Codex P1] ticks patch from the ORIGINAL snapshot | **Real** | refresh base is the last-good node, adopted on every non-full-failure result; test: a partially-failed tick keeps the prior tick's fresh value instead of rolling back |
| [Codex P1] data-equal recovery ticks dropped → stale note sticks | **Real** | change detection compares status + error count alongside data; test: fail → recover with identical data → note clears |
| [Codex P1 + Opus P2] gallery mutations save from the stale `flow` prop | **Real** | rename/pin are read-modify-write against `store.load(id)`; delete captures the current record so undo restores fresh data; the hook's write-back already merges only `node` |
| [Opus P2] toast countdown resets on parent re-render | **Real** | countdown anchored to the instance via a callback ref (fires the latest callback, on the original schedule); regression test |
| [Opus P2, scope] composer hoist leaked to all surfaces + remount loses focus/draft | **Real** | hoist is now opt-in via `heroComposer` — only the demo library page sets it; overlay/slot verified unchanged in-browser. Residual (accepted, noted): on hero surfaces the composer still remounts once on the first send — at that moment the draft is empty by definition; called out rather than engineered around |

Verified: full suite + typecheck green (15 tasks), and in-browser on `pnpm demo` — hero composer on the library page only, Cmd-K overlay composer untouched, double-delete undo queue end-to-end.

Merging per the pipeline.

https://claude.ai/code/session_01NMVoBXj41pmA9aP3tFx1rz

PR #21, opened 2026-07-02
ENG-188: automations engine — DSL + interpreter + grants, demo-parity snitch as data, AutomationCard

yousefh409:
## Summary

ENG-188 front half: the automations engine as pure library code in `@flowlet/agent`, wired into demo-bank so the Slack snitch is **pure data** (demo parity was the acceptance bar), with chat authoring behind an always-approval-gated `create_automation` and the new AutomationCard (approved by Yousef as-is).

Spec: [`docs/superpowers/specs/2026-07-01-flowlet-automations-proposal.md`](docs/superpowers/specs/2026-07-01-flowlet-automations-proposal.md) (includes the post-review amendments + contracts-freeze alignment) · Plan: [`docs/superpowers/plans/2026-07-01-flowlet-automations-engine.md`](docs/superpowers/plans/2026-07-01-flowlet-automations-engine.md) · Linear: ENG-188

### Engine (`packages/flowlet-agent/src/automations/`)
- **DSL** — zod schema (discriminated triggers, `steps`/`agent` execution union, 4 node types, snake_case unique ids, hard ceilings); the spec doc's four worked examples are test fixtures.
- **Expressions** — JSONata safe profile: AST-validated (no `$eval`, no lambdas), length/time/output caps, `{{ }}` interpolation with whole-value raw resolution, closed scope (`trigger`/`steps`/`run`/`user`).
- **Interpreter** — policy consulted per step (`allow` / `approve`+scope-hashed-grant / `deny`), pause + checkpointed resume that replays nothing, branch/for_each with iteration-scoped outputs, evaluated-input validation against tool schemas, retry only on idempotent-hinted tools with deterministic idempotency keys, dry-run simulation, wall-clock guard, gated tools inside agent runs.
- **Grants** — scope-hashed over (tool descriptor, trigger, guard, input mapping); never in the compiler-emitted DSL, re-collected on every version, drift ⇒ pause.
- **Store/runner/scheduler** — built on the FROZEN core seams (no parallel definitions): `AutomationEngineStore` extends core `AutomationStore` (Principal-scoped, store-owned identity/timestamps; coarse statuses kept exhaustive with additive `outcome` + `disabledReason`); deterministic firing ids dedupe redelivery; visible firing-cap drops; park after 5 consecutive failures; serialized per-automation firings; `InProcessScheduler` implements core `Scheduler` (Principal replayed per `AutomationFiring`); host events are ingest paths, not Scheduler paths. Retain-everything per ruling (retention = marked TODO).
[...]

PR #22, opened 2026-07-02
Runtime carve-out: extract @flowlet/runtime (portable loop + five seams)

yousefh409:
Wave 2, track 1 — the cloud-wave linchpin. Extracts `@flowlet/runtime` (new `packages/flowlet-runtime`) out of `flowlet-agent` per platform architecture Decisions 1 and 7. Behavior-preserving refactor: no new features, no UI changes.

## What's here

1. **The carved package** — loop, tool-calling, policy, UI-gen tools, and the ENG-188 automations engine, moved wholesale. The first commit (`0b932faf`) is a **pure `git mv`** — 65 files, 0 insertions, 0 deletions — so every file is a clean rename.
2. **In-memory implementations of all five frozen seams** (`src/embedded/`): `createInMemoryStore` (threads + saved flowlets + audit + the existing `InMemoryAutomationStore`), `InProcessCredentialBroker`, `InProcessExecutor`, `InAppChannels`, plus re-exports of the existing `InProcessScheduler`. 17 new tests.
3. **demo-bank on the carved runtime, zero behavior change** — import swap only; every runtime logic file is byte-identical to main (verified: `diff <(git show main:...agent/src/engine.ts) ...runtime/src/engine.ts` → empty, same for render-view-tool, wrap-tool, toolset, automations/{tools,runner}).
4. **The permanent CI guard** — `src/dependency-guard.test.ts`: (a) package.json deps must equal a reviewed allowlist, (b) no dep/devDep may be a db/queue/http server, (c) an import scan over `src/` bans db drivers, queues, http frameworks, and Node server builtins (incl. bare side-effect imports). **Mutation-checked**: adding `pg` to deps fails 2 tests; adding a `node:http` side-effect import fails the scan; clean tree passes.

## Decision flagged for Yousef: flowlet-agent is ABSORBED, not shimmed

`@flowlet/agent` is gone; no compat re-export package. Rationale: the architecture doc says flowlet-runtime "evolves flowlet-agent"; every consumer is in-repo (demo-bank, examples/basic — both rewired here); a permanent shim would give one library two names and force the dependency guard to police both. Shout if you want the shim instead — it's a 10-line package.

## Coordination note for eng-184-brand-native (rebase onto this)

Your files moved but their **contents are untouched**: `engine.ts`, `render-view-tool.ts`, `compile-component.ts` now live at `packages/flowlet-runtime/src/` (same relative paths). The move commit is a pure rename, so `git rebase` will follow it automatically. flowlet-components was not touched at all.

## Verification

- `pnpm build` / `typecheck` / `test` — all green (fresh run: 258 passed | 4 skipped in @flowlet/runtime; full workspace 14 test tasks green).
- **Browser gate** (screenshots in `verification/runtime-carve-out/`): spending chart via render_view in the sandbox; approval-gated createOrder (card → Approve → order posted); request_connect Slack card + connect flow; automation authoring card (trigger/guard/steps/permissions) → approved → **fired on a real late-night order** — run history shows `succeeded` + Slack post `succeeded`.

render_view chart

## Pre-existing issues observed (NOT fixed here, per standing rules)

- `pnpm lint` fails on demo-bank: 4 react-hooks errors in `app/transactions/[id]/page.tsx`, `app/transactions/page.tsx`, `components/flowlet/FlowletPoller.tsx`, `components/transactions/filters-bar.tsx` — all files untouched by this branch (last modified `b66072ad`, pre-main-merge).
- UX edge: typing a new chat message while the connect-flow's automatic resubmit is in flight yields the ai-SDK `tool_use without tool_result` error and a stuck "Something went wrong / Retry" alert. Reproduced during verification; runtime logic is byte-identical to main, so this exists on main too. Worth a follow-up ticket.

https://claude.ai/code/session_01UACA474phNbCadE5iv5res


yousefh409:
## Dual-review triage (fresh Codex CLI + fresh Opus)

Both reviewers: **no findings in the carve-out itself** — pure rename confirmed, no stale imports, no duplicate/missing exports, public API preserved, no demo-bank behavior delta. All findings were in the NEW embedded/guard code. Triage:

**Fixed in `fe7800cf`** (each reproduced red-first with a regression test):
1. Scope-key delimiter collision (`{a, b::c}` vs `{a::b, c}`) → Principal ownership now compares tenantId/subject as fields.
2. Aliasing: thread messages and saved-flowlet `uiTree`/`query` shared references with caller state → `structuredClone` at every store boundary (matches the copy semantics the cloud Postgres store gets from serialization).
3. Audit log's public mutable array → private array, cloned on append AND read; append-only for real.
4. `InAppChannels.onDeliver` not awaited → now `void | Promise<void>`, awaited; a failed host delivery rejects `deliver` and is not recorded.
5. Guard bypass via computed dynamic imports / `createRequire` → new rule bans non-literal `import(...)` in src (mutation-checked), `module`/`node:module` added to the denylist, scan extended to `.js/.jsx/.mjs/.cjs`.

**Rejected with reasoning:**
- "Parse imports instead of regex" (false-positive concern): a TS-parser devDep for one lint test isn't worth it; the raw-source trade-off is documented in the test header and the one live false positive (a doc comment) was caught during the mutation checks. Determined-evasion resistance is explicitly out of scope for a static lint — the guard exists to catch honest mistakes in CI.
- `getMessages` returning `[]` on miss vs `get` returning `undefined`: the frozen `ThreadStore` contract types `getMessages` as `Promise<FlowletUIMessage[]>` — `[]` is the only lawful miss value.
- Banning http *clients* (undici/axios): mission scope is db/queue/http-**server**; Composio's client SDK is a legitimate runtime dependency.

**Reported, not fixed (pre-existing, outside this PR per standing rules):**
- `InMemoryAutomationStore` (merged in PR #21) uses the same `::`-delimited scope key — same theoretical collision. Suggest a small follow-up.
- `apps/demo-bank/src/components/flowlet/run-query.test.ts` fails raw `tsc --noEmit` (2 errors) — file untouched here, vitest suite passes.
- The 4 demo-bank react-hooks lint errors already listed in the PR body.

Full verify after fixes: build/typecheck/test all exit 0; @flowlet/runtime 264 passed | 4 skipped.

https://claude.ai/code/session_01UACA474phNbCadE5iv5res

PR #33, opened 2026-07-03
feat: OSS install DX — flowlet init codemod, @flowlet/next handler, capability-additive keys

yousefh409:
## OSS install DX: `flowlet init` codemod + `@flowlet/next` handler + capability-additive keys

Implements the locked zero-infra BYO-keys install model end-to-end and **proves it on a fresh `create-next-app`**.

### What shipped

**1. `@flowlet/next` (new package)**
- `createFlowletHandler()` — one catch-all route (`runtime='nodejs'`) serving **chat** (streamed agent turn, toolkit+generation-keyed agent cache = the Composio ingestion cache-keying), **action** (sandbox dispatch through the policy), **integrations** (Composio connect/status/disconnect over an in-memory connections store), **capabilities**, **tick** (in-memory automations world). Zod-validated options (`model`, `instructions`, `policy`, `tools`, `components`, `hostTools`, `principal`, `connections`, `cacheKey`, `integrations`, `automations`, `maxSteps`); **zero-config reads `.env` + `.flowlet/` and just works**.
- `@flowlet/next/client` — `<FlowletRoot>`: FlowletProvider (HTTP transport + browser host-tool executor) + theme + shell + `FlowletOverlay`, generic sandbox stage, connect cards, reads-only saved-view query replay (client-executed per topology B).
- **Default policy surface** (please attack): client host tools → manifest annotations; engine tools + automation reads → allow; Composio names → whole-segment verb heuristic (writes win); everything else → approve. Default **local-requests-only** guard; deploys need a `principal` resolver or `FLOWLET_ALLOW_REMOTE=1`.
- **Approval hardening over the demo pattern:** stage actions gate on single-use, TTL'd, payload+user-bound approval tokens instead of a trusted `approved:true` re-POST.

**2. `flowlet init` codemod (CLI)**
Detects Next App Router and writes: `app/api/flowlet/[...path]/route.ts`, `app/flowlet-root.tsx` (imports `.flowlet/` JSON), wraps root layout `{children}` (JSX-position-aware, idempotent, respects existing providers), `.env.example` (three keys documented), `public/flowlet/` sandbox assets (bundled into the CLI at build), adds `@flowlet/next` dep, runs the existing extractor. **Fail-open contract:** anything uncertain is skipped with exact manual instructions — the layout edit only fires when exactly one JSX-position `{children}` exists.

**3. Capability-additive keys**
`ANTHROPIC_API_KEY` = chat + generated UI. `+COMPOSIO_API_KEY` = integrations (absent → tray empty, POST 503, no errors). `+OPENAI_API_KEY` = voice **flag only** (ENG-185 owns voice UX). Client gates on `GET /capabilities`.

**4. demo-bank migrated onto the handler** — chat/action/integrations routes deleted; demo store/world/policy/prompt flow in via options; poll/reset stay custom. Proof the handler covers a real host.

**5. `docs/quickstart.md`** — the story exactly as proven, honest about npm publish being ENG-198.

### Acceptance-bar proof (browser-verified)

Fresh `create-next-app` in a scratch dir → tarball install (`pnpm pack` × 8 + vendored fluidkit, npm-registry snapshot in this env forced pnpm) → `npx flowlet init .` → paste one Anthropic key → `npm run dev` → **chat answered and a generated dashboard (stat cards + line chart + table) rendered in the sandboxed iframe**. User-path wall clock well under 10 min.

fresh app launcher
fresh app generated view

demo-bank after migration — beat 1 ($87 DoorDash time-of-day view) + Gmail connect fast path verified live:

demo-bank beat 1

### Verification evidence
- `pnpm build` 18/18 · `pnpm test --force` 19/19 (55 new @flowlet/next tests, 12 codemod fixture tests, 76 demo-bank) · `pnpm typecheck` 19/19 · dependency-guard 4/4 (runtime untouched)
- `pnpm lint`: demo-bank has **pre-existing** `react-hooks/set-state-in-effect` errors in files byte-identical to main (`src/app/transactions/[id]/page.tsx`, `src/app/flowlet/page.tsx`, `src/components/verify/*`) — not introduced here, reported not fixed.

### Known limitations / flags for Yousef
- **Launcher pill visual is provisional** (built from shell tokens only, `launcher="none"` opt-out) — composed existing approved surfaces otherwise, but this one element is new chrome awaiting your eye.
- Packages not on npm (ENG-198); `@flowlet/shell`'s `fluidkit` relative `file:` dep is a publish blocker to solve there.
- Env-specific: npm registry snapshot in the dev sandbox rejected `@openuidev/react-ui@0.12.1` by date; proof used pnpm.
- Handler instructions are built once per process (component registries are import-time static, so parity holds).

https://claude.ai/code/session_014rPw8E7VuYuEbcQkV78k6T


yousefh409:
## Dual-review triage (fresh Codex + fresh Opus)

Both reviewers attacked the three briefed surfaces: codemod edit safety, key handling, and the default policy/guard surface. **Both independently flagged the local-only default guard as P0** — that convergence drove the biggest fix. All real findings are fixed with regression tests; scope-bounded ones are documented. Commits: `9bb492ff` (Opus), `67f9c399` (Codex).

### Fixed in code

| # | Sev | Finding | Fix |
|---|-----|---------|-----|
| Host spoof (both) | P0 | `Host: localhost` reaches the default principal on a public deploy | **Fail closed when `NODE_ENV=production`** — a deployed app gets no default identity; the spoofable Host check is now dev-only. Prod requires a `principal` resolver or `FLOWLET_ALLOW_REMOTE=1`. (`guard.ts`) |
| Codemod multi-line import | P0 | Import insertion landed inside a Prettier-wrapped `import {…}` block → syntax error | Mask strings/comments, insert at an always-valid position (after any leading directive, else top). (`next-wiring.ts`) |
| Codemod string/comment match | P1 | `{children}` inside a string/comment could be matched & corrupted while reporting success | Locate matches on the masked source → such matches are invisible → fail open. |
| Codemod `use client` + comment | P1 | A leading comment before `"use client"` pushed the import above the directive → server-component demotion | Skip header comments; keep the import below the directive. |
| Codemod idempotence | P2 | Incidental mention of the component in a comment false-skipped wiring | Idempotence checked against masked source for the real import. |
| Integrations status binding | P1 | Any active `connectedAccountId` flipped any toolkit connected | Require `hasActiveConnection(user, toolkit)` before flipping. (`integrations.ts`) |
| Integrations arbitrary id | P1 | `connect` spent the Composio key on arbitrary slugs | Validate id against the catalog before touching Composio. |
| Policy read-verb | P1 | `GMAIL_MARK_AS_READ` (mutating) auto-allowed via trailing `READ` | Read auto-allow now requires the read verb in the **verb position** (segment 1); write-anywhere still gates `FIND_AND_REPLACE`. (`default-policy.ts`) |
| Chat degradation | P1/P2 | Missing `ANTHROPIC_API_KEY` errored mid-stream | Server returns a clean **503**; client **hides** the assistant surface (capability-gated launcher/overlay). |
| Integrations error echo | P2 | Composio SDK error message echoed to caller | Logged server-side, generic message returned. |

### Accepted as documented (scope-bounded, not code-fixed)

- **Approval token "self-mint" (Codex P0)** — the token is a **human-consent tamper/replay binding**, not an authz gate; the gate is `principal` + policy. Remote-anonymous self-approval is closed by the production fail-closed guard above; an already-authorized caller scripting their *own* actions is inherent to topology B (they can hit the host API directly). Clarified in `action.ts`.
- **Single-tenant world/connections/approvals (both, P1)** — the in-memory stores are process-global and not user-partitioned. A `principal` resolver is an access *gate*, not tenant *isolation*. Documented in `world.ts`, `guard.ts`, and **quickstart → Deploying** (run as one long-lived process; multi-tenant installs inject their own stores). Consistent with the locked zero-infra **single-user** OSS model.
- **Serverless in-memory state (Opus P1)** — approval tokens / connections don't survive across cold instances. Documented: run as one long-lived Node process, not spread-serverless.
- **Codemod JSX-attribute `{children}` (Codex P1)** — a `{children}` nested in a JSX *attribute* can't be distinguished from a real child by regex; documented as a known limitation (exotic hand-authored layouts wire manually; the whole edit is one reviewable diff). create-next-app and typical provider layouts are handled correctly.
- **Route-group-only roots (Codex P2)** — `app/(group)/layout.tsx` with no top-level root is skipped with an **accurate** message (was wrongly implying Pages Router).
- **`/capabilities` returns booleans unauthed (Opus P2)** — intentional: the client needs it to decide UI; it discloses only which key *categories* are set, no values.

### Held under attack (both confirmed)
No API key reaches the client bundle or any HTTP response. Approval-token crypto/binding (unforgeable UUID, single-use, TTL, `(action,payload,user)`-bound) is sound. Chat can't be injected with tools via the request body. Verb heuristic uses whole-segment matching with write-precedence.

### Verification
Fresh-app e2e **re-proven with the hardened packages** (bar chart rendered in-sandbox — new screenshot in `verification/install-dx/`). `@flowlet/next` 61 tests, `@flowlet/cli` 70 tests (incl. regressions for every codemod fix), full workspace `build`/`test`/`typecheck` green.

PR #56, opened 2026-07-05
Hardening on top of launch prep: reliability + error hygiene + publish gates

yousefh409:
## What

**Stacked on #60 (merge that first).** The hardening-audit fixes that launch prep doesn't cover — reconciled after #60 landed the overlapping work (CLI bin guard, NodeNext sweep, manifests, Apache-2.0, fluidkit bundling). 10 commits, every behavioral fix TDD'd. Full audit context: bar + 131-finding inventory + decision queue in PR #58.

### Reliability (crash/data-loss blockers)

1. **Scheduler tick per-fire isolation** — one rejecting firing handler was an unhandled rejection (process-fatal on default Node ≥15) and silently starved every other due schedule in the consumed window. Now isolated per firing (`onFireError` hook); one-shot occurrences are consumed regardless of delivery outcome.
2. **Same pattern in the other two fan-outs** (Codex review catch): host-event ingest and the Composio webhook loop; the webhook's `fired` count now reports actual successes.
3. **pg.Pool error listener** — an idle Postgres connection drop crashed the entire host app.

### Provider ladder

4. **Missing optional provider peer degrades instead of 500ing every route** — `OPENAI_API_KEY` without `@ai-sdk/openai` failed state assembly before routing, breaking "missing key hides the surface, nothing errors" on the documented one-key install. Typed `ModelPeerMissingError` now degrades like the keyless state (chat gated, `npm i` hint in server log + chat 503); a misconfigured `VENDO_MODEL` still fails loudly. Quickstart synced.

### Error hygiene (server-side; DOM copy is Yousef-gated, queued separately)

5. **Boot/assembly errors no longer echoed verbatim to callers** (file paths, connection strings) — generic body, full detail in the server log; deliberate `[vendo]`-prefixed developer messages still pass through for local dev requests only.
6. **Route-level error boundary** — a throw inside any route was a framework-level HTML 500 (which the sandbox parsed as JSON → raw SyntaxError); now a generic JSON 500 + logged detail. Next-adapter regression tests pin the contract.
7. **Run failures stream a generic error part** — provider 401s (with key prefixes) and network errors were serialized raw into the UI stream at two points (`createUIMessageStream.onError` and `toUIMessageStream()`'s default serializer); sanitized at the stream boundary so telemetry/policy classification still see real errors.
8. **Unreadable `.vendo` files fail boot loudly** — EACCES/EIO were treated as file-absent, silently serving an empty manifest.

### Gates + residuals

9. `scripts/check-publish-hygiene.mjs` + `scripts/check-node-esm.mjs` — rerunnable release gates. On top of #60 they caught: `vendo-sandbox-shims` missed by the `.js`-extension sweep (only package skipped as private — its dist wasn't Node-loadable), and the CLI's conflicting duplicate typescript pins. Also: the CLI bin guard refactored to an exported `isCliEntrypoint` with regression tests (npm .bin symlink, spaces in paths, missing argv).

## Verification

- `pnpm typecheck` 27/27, `pnpm test` 25/25 on top of #60's tree; both gates green (hygiene: 11 publishable packages OK; ESM: 17/21 + 4 browser-only allowlisted)
- Wave-1 versions of these fixes were verified against real npm-installed tarballs (symlinked `npx vendo --version` prints; `import('@vendoai/server')` resolves 56 exports in plain Node)

No UI changes.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -84,8 +84,10 @@ VENDO_MODEL=claude-sonnet-4-6     # bare id: applied to whichever provider key i
 `VENDO_MODEL` alone names a model, not a credential: without a real
 provider key (or a code-injected `model`), chat stays off. OpenAI and Google
 are optional peers (`@ai-sdk/openai`, `@ai-sdk/google`): resolving to one
-without its package installed fails fast with an actionable `npm i` hint, not
-a silent fallback.
+without its package installed degrades like the keyless state — every route
+stays healthy, `/capabilities` reports `chat: false`, and the server log and
+the chat 503 both carry an actionable `npm i` hint. A misconfigured
+`VENDO_MODEL` (unknown provider) still fails loudly.
 
 ## MCP servers
 
--- a/packages/vendo-cli/package.json
+++ b/packages/vendo-cli/package.json
@@ -2,7 +2,13 @@
   "name": "@vendoai/cli",
   "version": "0.1.0",
   "description": "One-command Vendo install: extracts theme, tools, and components from a host app and wires in the embedded agent.",
-  "keywords": ["ai", "agent", "generative-ui", "cli", "devtool"],
+  "keywords": [
+    "ai",
+    "agent",
+    "generative-ui",
+    "cli",
+    "devtool"
+  ],
   "license": "Apache-2.0",
   "repository": {
     "type": "git",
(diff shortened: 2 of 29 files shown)


yousefh409:
Superseded by #75 (full hardening stack reconciled onto main and merged). This PR's commits are all included there.

PR #69, opened 2026-07-06
Shell-UI wave: glass skeletons, mobile takeover, summary-only consent, MCP consent (Yousef-approved)

yousefh409:
## What

The four UI changes Yousef approved on 2026-07-05, browser-verified on the integration build. Stacked on #66 (client/state). 5 commits, TDD'd. **Screenshots: `~/Desktop/vendo-ui-bash-2026-07-05/final-verification.html`** (5 of 6 fixes verified working in-browser).

1. **Glass skeleton loading states** — the landing page's glass+shimmer recipe, tinted from each host's theme accent, now fills the render_view placeholder, the saved-vendo refresh veil, and library/tray loading. FluidThinking still owns chat thinking (different moment). *Verified.*
2. **Full-screen mobile takeover** — below 768px any open Vendo surface takes over full-screen (Intercom pattern): fixed, composer pinned and visible, close reachable; reverts to the inset panel above the breakpoint. Fixes the old clipped-composer breakage. *Verified.*
3. **Summary-only consent card** — ordinary cards show only the human summary (no raw `Is html`/`User id` params, no "Unverified tool" badge on catalog Composio tools). Per Yousef's amendment, **critical-tier cards keep their material fields, humanized** (Amount, Recipient — not a raw dump). *Ordinary verified; critical unit-tested (no demo exposes a money tool for a live shot).*
4. **MCP approvals through the consent channel** — MCP dynamic-tool approvals dropped the toolCallId and skipped the consent/audit/grant pipeline; they now go through the same ConsentRequest/audit path as host tools (server test proves the audit-row parity).

### Verified alongside this wave (fixes from stacked PRs, re-confirmed on the integration build)
- Decline semantics: declining no longer re-pitches; re-proposes only on new intent.
- Donut center: the $40.18 (100×) bug is gone — center now $4,017.81 = legend sum (PR #65's deterministic fix + descriptor).

### For your call (in the decisions doc, non-blocking)
- **Chart brand palettes** — root-caused: OpenUI 0.12.1 ignores the `*ChartPalette` keys entirely, so brand chart colors never applied in either demo; this batch silenced the false-positive warnings (net improvement). Real brand-chart theming is a follow-up.
- **Critical-card live screenshot** — needs a demo with a genuine money/irreversible tool.

`pnpm typecheck` 27/27 · `pnpm test` 25/25.

---

<a href="https://app.devin.ai/review/runvendo/vendo/pull/69" target="_blank">
  <picture>
[...]

PR #70, opened 2026-07-06
Hardening wave 7: integrations + telemetry honesty

yousefh409:
## What

Stacked on the shell-UI wave (#69). Nine confirmed audit findings — integrations UI telling the truth about the server, and telemetry consent/privacy being honest. 9 commits, each TDD'd; Codex review returned **no findings**.

### Integrations truthfulness (vendo-server, vendo-next)
1. **Status no longer reports "active" without the store write** — a foreign active account returned `active` while the toolkit was never registered; now `active` only inside the store-write branch, anti-spoof case reports `pending` (bounded 120s poll, never latches).
2. **Disconnect no longer silently succeeds on a server reject** — now throws on non-ok and leaves the row untouched (mirrors `connect()`).
3. **Status poll no longer swallows errors** — logs server-side and reports transient `pending`, reserving `failed` for real failures.

### Telemetry consent (vendo-telemetry — anonymous, opt-out, posture unchanged)
4. The one-time notice is actually shown (was routed to a no-op logger then marked seen).
5. Prod guard fails **closed** on unset NODE_ENV (runtime collection only in explicit dev/test).
6. An explicit `{"optedOut":true}` without an anonymousId is honored, not overwritten to opted-in.
7. No tracking UUID minted/persisted under an env opt-out (`DO_NOT_TRACK`/`VENDO_TELEMETRY_DISABLED`/`CI`).
8. `vendo telemetry` can't crash — `loadConfig` degrades to in-memory on a write failure.
9. Allowlisted event values are bounded (coerced to primitives, strings capped, objects dropped).

### Flags (deliberate, non-blocking)
- Findings 1 & 3 use `pending` for not-connected/transient — correct per the client contract, but the spoof case polls to the 120s timeout rather than fast-failing. Your call if you'd prefer fast-fail.
- Fix 5 treats `development` AND `test` as dev envs (so vitest/CI runtime callers aren't surprised); one word to make it strictly `development`.

`pnpm typecheck` 27/27 · `pnpm test` 25/25. No UI changes.

---

<a href="https://app.devin.ai/review/runvendo/vendo/pull/70" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---
## Summary by cubic
Hardens integrations and telemetry so the UI shows real connection state and telemetry consent is enforced. Prevents false “connected” states and avoids minting or persisting tracking IDs when opted out.

- **Bug Fixes**
  - `vendo-server`/`vendo-next`: Status only returns active after the store write; foreign/other-toolkit “active” polls return pending so the client keeps polling instead of latching a false connected state.
  - `vendo-server`: Status poll errors are logged and reported as pending (transient), reserving failed for real failures.
  - `vendo-next`: `disconnect()` throws on non-ok responses and leaves the cache unchanged, mirroring `connect()`.
  - `vendo-server`/`vendo-telemetry`: The one-time telemetry notice is actually shown in dev; runtime collection fails closed unless `NODE_ENV` is explicitly `development` or `test`; a file with `{ optedOut: true }` is honored even without an `anonymousId`; env opt-outs (`DO_NOT_TRACK`, `VENDO_TELEMETRY_DISABLED`, `CI`) avoid minting/persisting an ID; config save failures degrade to in-memory to prevent crashes.
  - `vendo-telemetry`: Allowlisted event values are bounded to primitives; strings are capped at 512 chars and objects/arrays are dropped.

<sup>Written for commit d993e8ad499bcc9b828efb3cd98297e9056bd76a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/70?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/vendo-next/src/client/integrations.test.ts
+++ b/packages/vendo-next/src/client/integrations.test.ts
@@ -1,0 +1,28 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { createServerIntegrations } from "./integrations.js";
+
+afterEach(() => {
+  vi.unstubAllGlobals();
+});
+
+describe("createServerIntegrations.disconnect", () => {
+  it("throws and does not mark the row disconnected when the server rejects", async () => {
+    const fetchMock = vi.fn(async () => new Response("nope", { status: 500 }));
+    vi.stubGlobal("fetch", fetchMock);
+    const integrations = createServerIntegrations("/api/vendo");
+    await expect(integrations.disconnect("gmail")).rejects.toThrow();
+  });
+
+  it("updates the cache and returns disconnected on an ok response", async () => {
+    const fetchMock = vi.fn(
+      async () =>
+        new Response(JSON.stringify({ integrations: [{ id: "gmail", name: "Gmail", connected: false }] }), {
+          status: 200,
+        }),
+    );
+    vi.stubGlobal("fetch", fetchMock);
+    const integrations = createServerIntegrations("/api/vendo");
+    const result = await integrations.disconnect("gmail");
+    expect(result).toEqual({ id: "gmail", name: "Gmail", connected: false });
+  });
+});
--- a/packages/vendo-next/src/client/integrations.ts
+++ b/packages/vendo-next/src/client/integrations.ts
@@ -51,10 +51,11 @@ export function createServerIntegrations(basePath: string): VendoIntegrations {
         body: JSON.stringify({ id, action: "disconnect" }),
         cache: "no-store",
       });
-      if (res.ok) {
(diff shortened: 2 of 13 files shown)


yousefh409:
Superseded by #75 (full hardening stack reconciled onto main and merged). This PR's commits are all included there.

PR #111, opened 2026-07-12
wave 3: @vendoai/core built fresh to the frozen contract + quarry + dependency guard + conformance kits

yousefh409:
Wave 3 of the v0 campaign — the serial gate the six block builds fan out from. Contracts are FROZEN on main (PR #109); this branch builds `@vendoai/core` fresh to `docs/contracts/01-core.md` and does the wave-3 prep the ground rules assign.

## What's here

**1. The quarry** — all 11 pre-v0 packages and both demo apps moved to `legacy/` (read-only, out of the pnpm workspace and turbo). `@vendoai/telemetry` stays active (00-overview: orthogonal to the campaign). Demos + corpus are red **by design** until wave 7; corpus-nightly is stubbed to a loud notice (original in git history at `3c9ff4c8`). `docs/contracts/seams.md` deleted (decision 12).

**2. The dependency guard** — `scripts/dependency-guard.mjs`, runs in `pnpm lint`: enforces the frozen layering (core→apps→automations; store/agent/actions/guard/ui → core only; umbrella → everything), forbids quarry imports (by path, by relative escape out of a package dir, and by the 8 retired npm names that would silently resolve to 0.2.x publishes), and requires every `@vendoai/*` import to be declared. Unknown packages fail loudly. Residual gaps (computed dynamic imports, tsconfig aliases) documented in the header.

**3. `@vendoai/core`** (`packages/core`, zod-only, Node≥20/edge/Bun, NodeNext, smoke-tested on Node ESM + Bun):
- Every contract shape as TS type + `<name>Schema` zod pair (passthrough — tolerant reader, lossless round-trip); seams type-only. Full §1–§16 export inventory verified mechanically (53 types + 53 runtime exports).
- `validateTree` — pinned `vendo-genui/v1` wire semantics ported from the legacy validator (all caps, reserved primitives, dangling-children-allowed, generated-needs-definition, error codes) + the additive `fn:` scheme.
- `validateAppDocument` — `vendo/app@1` with the cross-field rules: components lifted to doc level at rest, `fn:` requires a machine (query/prop-action/step positions), reserved `state` collection, `host.` refs, provider-prefixed server ref, `sha256:` pin bases; unknown payload format tags accepted opaque (contained-failure doctrine).
[...]

PR #114, opened 2026-07-12
Wave 3: @vendoai/store — StoreAdapter on Postgres (PGlite default), typed helpers, reserved-collection seam

yousefh409:
## Wave 3 — @vendoai/store: persistence under everything

Implements `docs/contracts/02-store.md` (FROZEN) as a brand-new package against the real `@vendoai/core` (built on the wave-3 core skeleton; core commits ride along until #111 merges — this PR should land after it). Nothing here imports the quarry; `scripts/dependency-guard.mjs` enforces it.

### What's in the box

- **`createStore`** — one adapter, one schema: PGlite zero-config default at `.vendo/data` (fsync-on-write durability), real Postgres via `url`. Lazy driver init, loud failures (unwritable dataDir, serverless-without-url), pool error recovery, `ensureSchema()` advisory-locked on one dedicated session (the legacy pooled-lock bug stays dead), `close()`, `raw()`.
- **The public `vendo_` table map** exactly as contracted (11 tables), all JSON `jsonb`, GIN (`jsonb_path_ops`) on `vendo_records.refs` so the contract's containment join is index-usable.
- **Typed per-table helpers** (`appStore`, `stateStore`, `threadStore`, `grantStore`, `approvalStore`, `auditStore`, `runStore`) speaking core shapes 1:1, zod-validated at every write boundary, ISO-8601-UTC at every API boundary.
- **Secrets**: `envSecrets(prefix?)`, `storeSecrets(store)` over AES-256-GCM-encrypted `vendo_secrets` (32-byte base64 key, fresh 12-byte IV, tag-verified, no material in errors), plus blocks-only `secretStore` write surface.
- **Ephemeral principals never touch disk**: per-store in-memory overlay across all typed helpers, routed collections, AND `app:<appId>:<name>`-convention records/blobs (ownership resolved through the app row); `registerEphemeralSubject` for subject-only write paths.

### ⚑ Cross-block seam decision (needs your eyes, Yousef)

Blocks depend on core only and receive a plain `StoreAdapter`, so guard (and agent/automations/apps) cannot import this package's typed helpers — a gap in the frozen set (02 §3 vs the dependency rule vs 05 §1). Resolved with the guard lane, both PRs flag it:

- `records(<public table name>)` **routes** six reserved names (`vendo_grants`, `vendo_approvals`, `vendo_audit`, `vendo_threads`, `vendo_runs`, `vendo_apps`) onto the typed tables: data validated against core schemas, refs synthesized from columns on read (ignored on write), same rows the helpers produce (one codepath). The public table map stays true; hosts see exactly the contracted tables.
- This carves an exception into 02 §3's "collection names are opaque" (bare reserved names can't collide with the `app:<appId>:<name>` convention).
- `RecordStore.list()` is newest-first everywhere (core's memory double aligned in 193cc1bc).
- Additive columns: `vendo_approvals.session_id`, `consumed_at` (guard's session-grant + single-use machinery).
- `vendo_state` is NOT routed (composite pk, no id column) — flagged to the apps lane.

### Testing (e2e-first doctrine)

The ENTIRE suite runs twice — PGlite always, real Postgres gated on `POSTGRES_URL` (CI stays deterministic and keyless; a visible skip marks the gate). **Both legs verified here against a throwaway Postgres 18.4** — the postgres leg caught two real driver-parity bugs (node-pg JSON-serializes plain `Uint8Array` bytea; collation-dependent `ORDER BY`), both fixed with reads normalized so the two drivers are indistinguishable.

- Host-queryability proven with the contract's exact SQL join (`refs @> jsonb_build_object(...)`) plus table/column/index assertions through `information_schema`.
- **Kill-the-server drill**: a spawned child process writes through the built package, is SIGKILLed after awaited writes, and the reopened store proves every row survived (PGlite leg; Postgres owns its own durability, skip documented).
- Core conformance kits green on both backends: `storeAdapterConformance` + `secretsProviderConformance` (env + encrypted store).
- Ephemeral proofs are real SQL `COUNT(*) = 0` assertions, not API echoes.

### Dual review

Codex adversarial pass (`.lanes/R-REVIEW.md` process): 6 findings → 5 fixed (ephemeral app-data overlay for convention collections, zod at helper write boundaries, visible postgres-gate skip, `raw()` loud-before-open, strict canonical cursor decoding), 1 acknowledged-by-design (the ⚑ routing above). My own pass covered SQL parameterization, GCM usage, advisory-lock session pinning, cursor tie-breaking, LIKE escaping, overlay-shadow merge semantics.

### Live e2e status

The block's env-gated live e2e is the real-Postgres leg: the ENTIRE suite ran against a real Postgres 18.4 server (`POSTGRES_URL`) in addition to the always-on PGlite leg — 130 passed / 1 skipped (the skip is the PGlite-only SIGKILL drill; Postgres owns its own durability). **No key-gated e2es are pending for this block.**

### External AI review

Round-1 triage posted as a PR comment: Devin 7 findings — 2 real (both fixed in aeef00ff), 3 intended-and-documented/acknowledged, 2 core-lane files forwarded to #111. Greptile punted on diff size (carries the core merge until #111 lands).

### Gates

`pnpm build && pnpm test && pnpm typecheck && pnpm lint` green at root (dependency-guard: OK). No UI surface in this block — no browser evidence applicable.

https://claude.ai/code/session_016TDhkMQH4ZZSGJNKTqSkFT


yousefh409:
## External review triage — round 1 (Devin 7, Greptile punted on size)

Greptile: `Too many files changed` (the PR carries the core-branch commits until #111 lands — same situation as #111 itself). Devin reviewed the full diff; two findings target core-lane files and were forwarded to the core worktree for disposition on #111.

| # | Reviewer finding | Verdict | Disposition |
| --- | --- | --- | --- |
| D1 | 🟡 Ephemeral overlay stores/returns live references (`records.ts`) — mutations silently corrupt session data; SQL path returns copies | **Real — fixed** | `snapshot()` (structuredClone) now applied at every overlay boundary — generic records/blobs, all 7 typed helpers, routed collections. Regression tests mutate returned objects and prove the stored copy is untouched. (aeef00ff) |
| D2 | `jcs.ts` UTF-16 sort-order comment suggestion | Core-lane file | Forwarded to #111's worktree — informational; implementation + test already correct per RFC 8785 §3.2.3. |
| D3 | dependency-guard: document devDependencies exclusion | Core-lane file | Forwarded to #111's worktree — behavior is intentional (layering applies to test imports too). |
| D4 | `stateStore.get` registers ephemeral subjects on read | Intended | The caller's `principal.ephemeral` flag is the authority; registration caches it so later subject-only writes (e.g. routed grants) stay off disk. Now documented in the README's ephemeral section. (aeef00ff) |
| D5 | Routed list merge over-fetches/sorts when overlay is large | Acknowledged, no change | Correctness confirmed in the finding itself; overlays are session-scoped and small. Revisit only if a real workload shows it. |
| D6 | Failed lazy PGlite open caches the rejected promise forever | **Real — fixed** | `opening` clears on rejection so the store can retry once the cause (perms, path) is fixed; regression test asserts a second attempt re-runs instead of replaying a zombie promise. (aeef00ff) |
| D7 | `audit.export()` omits ephemeral overlay events | Intended | `export` is the durable compliance log; session-scoped ephemeral events surface via `query()` only. Now documented in the README. (aeef00ff) |

Gates re-run after fixes: package suite green on both backends (PGlite + real Postgres 18 via `POSTGRES_URL`), root `build/test/typecheck/lint` green.

### Live e2e status (per the wave-3 DONE bar)

Store's env-gated live leg is the **real-Postgres run of the entire suite**: executed locally against a real Postgres 18.4 server (`POSTGRES_URL` set) — **130 passed / 1 skipped** (the one skip is the documented PGlite-only SIGKILL durability drill; Postgres owns its own durability). No keys are pending for this block. If a hosted-Postgres (Supabase/Neon) run is also wanted, the same suite takes any `POSTGRES_URL` verbatim.

PR #123, opened 2026-07-12
security(redteam): adversarial wave — harness, regressions, 2 guard fixes

yousefh409:
## Red-team (security) wave

Adversarial testing of Vendo's deliberately-unusual security model — **apps never hold authority; the guard asks the running user in context at call time; grants are per-user + app-bound; artifacts carry zero authority (import mints a fresh AppId); secrets are handles substituted at the egress boundary; the app backend always runs in the sandbox.** Both independent clean-room designs chose declared manifest permissions instead — ours is the novel bet, so it was attacked hard. Every attack is now a permanent regression test.

**130 new tests**: block-local suites in every owning package + a `fixtures/redteam` mini-umbrella (real store + guard + actions + apps + automations against a live fixture host app) + a **real-e2b** live egress leg + a **live prompt-injection** leg run against a real model.

### Two real fixes landed in `@vendoai/guard` (each with a revert-to-fail regression)

- **P1 — away-park bypass via the policy code hatch.** `normalizeCodeDecision` preserved a code-supplied `decidedBy`, and the away-downgrade gate exempts `decidedBy === "grant"` — so `policy.code` returning `{action:"run", decidedBy:"grant"}` ran an **away** call with no real app-bound automation grant, subverting the model's central invariant. Fix: code-stage provenance is now forced to `"rule"` (mirroring the code-error path) and any code-supplied `grantId` is dropped, so a code-sourced run is away-downgraded to a park and honestly audited. Operator-trust-boundary → ranked P1, but it's the one place the away invariant could be subverted in code, so it's fixed. *(Reverting the fix fails `away-code-hatch.test.ts`.)*
- **P2 — empty `constrained` grant = silent tool-wide wildcard.** `constraints.every(...)` returns `true` over an empty array, so an empty-constraints grant matched any args while the preview implied it was bounded. Fix: rejected at mint (`VendoError("validation")`) + fails closed in `scopeMatches`. *(Reverting fails `empty-constrained.test.ts`.)*

No frozen contract shape changed; the fixes enforce the contract's away invariant, which is the higher law.

### Key architectural finding (fail-safe — flagged to apps/composition owners)

`substituteSecretHandles` is exported and unit-tested but **not wired into the OSS e2b/modal egress path** — `machine.ts` injects the handle as the env value, egress is enforced by the provider-native network allowlist, and `SecretsProvider.get()` isn't even called at machine boot. So the real secret value **never enters the sandbox** — exfil is impossible *by construction*, strictly stronger than the contract's allowlist-gated substitution. Confirmed on real e2b (handle-only env; non-allowlisted/raw-IP/redirect-to-non-allowlisted egress all blocked). Functional handle-resolution is effectively Cloud-only — a functionality gap to reconcile, not a security defect.

### Proven fail-closed (now regressions)

Shared/imported dormant privileged calls (critical always asks with real inputs; default posture auto-runs-but-audits — the model's teeth are the critical tier + configured policy); artifact-borne authority (fresh AppId, no grants/appId/pins/egress/forkedFrom transfer, tampered `.vendoapp` bytes inert); away-run authority (chat/batch grants never authorize away; revoked/drifted grants park; approved critical call single-use); prompt injection (a fully-compromised judge can't unlock critical/away/breaker; a live agent holding a standing app-bound grant still parked a destructive critical call); run-token abuse; egress/secret exfil; webhook forgery/replay/skew/oversize; descriptorHash/JCS/AES-GCM/SQL-isolation.

### Delegated boundaries (outside this wave's blocks — flagged to composition/wave-5)

`AppsRuntime.history(appId)` ownership at the umbrella wire route; egress cross-hop (redirect/DNS-rebind) in the provider network adapter (probed live, holds); unbounded tree `data`/`props` size (upstream body limit); reserved-collection writes trust their caller (structural isolation — app code never holds a `StoreAdapter`).

Full write-up: [`fixtures/redteam/SECURITY-FINDINGS.md`](fixtures/redteam/SECURITY-FINDINGS.md).

### Gates
`build` 10/10 · `typecheck` 19/19 · `lint` (dependency-guard passes) · `test` 21/21 tasks. Wave e2e green incl. live e2b egress + live injection. No npm publish.

https://claude.ai/code/session_015xrmEJHPGu8Y211P7ACbrE

---

<a href="https://app.devin.ai/review/runvendo/vendo/pull/123" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---
## Summary by cubic
Adds a red‑team security harness and 130 regression tests to stress our “apps hold no authority” model, plus two fixes in `@vendoai/guard`. The live e2b suite now hard‑asserts allowlisted egress reachability and confirms secrets never enter the sandbox.

- **Bug Fixes**
  - Force code-stage decisions to `decidedBy: "rule"` and drop any code-supplied `grantId` to prevent forged `grant` provenance from running away calls in `@vendoai/guard`.
  - Reject empty `constrained` scopes at mint and fail closed at match-time in `@vendoai/guard` so an empty array cannot act as a tool-wide wildcard.

- **New Features**
  - Added `fixtures/redteam` harness that composes the real store, `@vendoai/guard`, actions, apps, and automations against a live fixture host app.
  - 130 regression tests covering artifact authority, away authority, approvals/replay, grant drift, proxy escalation, secrets crypto/isolation, webhook verification, prompt-injection, and DoS bounds.
  - Live e2b egress suite now hard‑asserts allowlisted control reachability and proves provider allowlist enforcement; also confirms secret handles are never resolved inside the OSS sandbox.

<sup>Written for commit 9711731500762ceb4dd0ee14fcaff553326b19a2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/123?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/fixtures/host-app/tsconfig.json
+++ b/fixtures/host-app/tsconfig.json
@@ -39,7 +39,9 @@
     ".next-automations-e2e/dev/types/**/*.ts",
     ".next-automations-e2e/dev/dev/types/**/*.ts",
     ".next/automations-e2e/dev/types/**/*.ts",
-    ".next/automations-e2e/dev/dev/types/**/*.ts"
+    ".next/automations-e2e/dev/dev/types/**/*.ts",
+    ".next/redteam-e2e/dev/types/**/*.ts",
+    ".next/redteam-e2e/dev/dev/types/**/*.ts"
   ],
   "exclude": [
     "node_modules"
--- a/fixtures/redteam/SECURITY-FINDINGS.md
+++ b/fixtures/redteam/SECURITY-FINDINGS.md
@@ -1,0 +1,139 @@
+# SECURITY-FINDINGS — v0 red-team wave
+
+Adversarial testing of Vendo's deliberately-unusual security model: **apps never hold authority;
+the guard asks the running user in context at call time; grants are per-user + app-bound; artifacts
+carry zero authority (import mints a fresh AppId); secrets are handles substituted at the egress
+boundary; the app backend always runs in the sandbox.** Both independent clean-room designs chose
+declared manifest permissions instead — ours is the novel bet, so it was attacked hard.
+
+Every attack below is now a permanent regression test (130 new tests across block-local suites and a
+`fixtures/redteam` mini-umbrella that composes the real store + guard + actions + apps + automations
+against a live fixture host app, plus a real-e2b live leg). **Two real fixes landed in
+`@vendoai/guard`; everything else was already fail-closed and is now pinned.** No remotely-exploitable
+P0 was found.
+
+Root gates green: `build` 10/10, `typecheck` 19/19, `lint` (dependency-guard passes), `test` 21/21
+tasks. The wave's own e2e is green, including the live e2b egress leg and a live prompt-injection leg
+run against a real model.
+
+## Fixes landed (in the owning block, each with a revert-to-fail regression)
+
+### P1 — away-park bypass via forged `decidedBy` from the policy code hatch  (guard)
+`normalizeCodeDecision` preserved whatever `decidedBy` a `PolicyConfig.code` function returned,
+including `"grant"`. The away-downgrade gate exempts exactly `decidedBy === "grant"`, so a code hatch
+returning `{action:"run", decidedBy:"grant"}` executed an **away** call with no real app-bound
+automation grant behind it — subverting the model's central invariant ("away runs hold only
+automation grants bound to the running app; everything else parks"). The trust boundary is the
+operator-authored policy module (not a remote attacker), so this is ranked P1, not a
+remotely-exploitable P0 — but it is the one place the away invariant could be subverted in code, so
+it is fixed.
+**Fix:** `normalizeCodeDecision` now forces the policy-code stage's provenance to `decidedBy:"rule"`
+for all outcomes (mirroring the already-correct handling of code *errors*) and drops any
+code-supplied `grantId`. A code-sourced run is now treated like a rule-sourced run: away-downgraded
+to a park, and honestly attributed in audit. No frozen contract shape changes; this enforces the
+contract's away invariant, which is the higher law.
+Regression: `packages/guard/test/security/away-code-hatch.test.ts` (verified: reverting the fix lets
+the forged grant-run execute the away call).
+
+### P2 — empty `constrained` grant = silent tool-wide wildcard  (guard)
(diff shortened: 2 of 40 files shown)


yousefh409:
## Reviewer triage

**Codex review**: ran the full guard suite (137 passed / 3 skipped) against the diff — **no exploitable issue**. Adversarial self-pass: after the fix, `decidedBy:"grant"` can only originate from a real stored-grant match or an approved single-use replay (never the code hatch), so the away-gate exemption is sound; the dropped `grantId` also closes a secondary forged-grantId-fetch vector.

**Greptile — P1 (live-egress control soft-skips):** ✅ **fixed** in `97117315`. The allowlisted control (`example.com` reachable) was a soft-skip, so on a deny-all/dead network the deny-side probes would pass vacuously while the control merely warned. Converted to a **hard assertion** (retries kept for transient flakiness) so the suite proves *both* allow and deny. Re-verified live on real e2b: control reached + all deny-side/secret-handle assertions hold (15/15). Deny-side assertions untouched.

**Devin — 4 Info items, all acknowledged, no change needed:**
1. *normalizeCodeDecision normalizes all 3 actions, not just `run`* — intentional and *more* correct. `decidedBy` is a fixed enum, so a code hatch can't emit a custom label (the hypothetical is type-impossible); forcing `"rule"` on `ask`/`block` too also prevents a code hatch from claiming `"critical"` on an `ask`. Audit stays honest (code *is* the policy/rule stage).
2. *empty-constrained dual-layer relies on guard↔automations interaction* — correct: automations' `mintGrant` only ever uses `scope:{kind:"tool"}`, so it can't produce an empty-constrained grant today; the `scopeMatches` layer is exactly the defense-in-depth for any future change. Working as designed.
3. *enableAndApprove batching* — Devin concludes "subtle but correct." Agreed.
4. *global-setup port TOCTOU* — copied verbatim from the already-merged `automations-e2e` harness; the window is milliseconds and changing it here would diverge from the established pattern. Accepted as by-pattern.

Full write-up: `fixtures/redteam/SECURITY-FINDINGS.md`.

PR #126, opened 2026-07-12
E2E hardening wave: every block through public contracts against real neighbors (+3 product bugs fixed, +11 review fixes)

yousefh409:
# E2E hardening wave — every block, through the public contract surface, against real neighbors

Per the hardening execution model (orchestrator = Fable; all implementation on Codex/Opus lanes): one lane per block audited its FROZEN contract line-by-line against existing e2e, wrote the missing coverage, ran it, and fixed the product bugs it exposed. E2E here means: through the public contract surface, against real neighbors (real PGlite store, real `guard.bind`, real fixture host app, real Chromium), asserting real side effects (raw SQL on the public `vendo_*` tables, real HTTP, real browser).

**Net: ~120 new e2e tests, 3 product bugs found + fixed, 1 fire drill proving the format-evolution seam opens.**

## Bug ledger (what the wave exists for)

| # | Bug | Root cause | Fix |
|---|---|---|---|
| 1 | **`vendo_state` routing** (known wave-4 follow-up, now proven): apps' state singleton (`records("vendo_state")`, id `${appId}:${subject}`) landed in the generic `vendo_records` table — the contracted host-queryable `vendo_state` table stayed empty, `appStore.delete`'s cascade missed the rows (state leaked after app deletion), and `stateStore` read a table the runtime never wrote | `vendo_state` missing from the store's reserved-collection routing | `packages/store/src/routing.ts`: routed config (first-colon id split, `(app_id,subject)` upsert, shared ephemeral overlay with `stateStore`); idempotent `ensureSchema` backfill relocates legacy rows; 8 new tests incl. cascade + ephemeral zero-disk proofs |
| 2 | **Agent thread persistence never worked against the real store** (showstopper): threads silently failed to persist through `@vendoai/store` — composite `subject:threadId` record ids rejected by the store's `thr_` id validation; `Thread` read from `record.data` while the reserved routing projects data to `{subject, messages}` with id/timestamps on the envelope; ai-SDK `UIMessage` tool parts carry explicit `undefined` props the store's `Json` validator rejects | agent written/tested only against the verbatim in-memory conformance adapter | `packages/agent/src/threads.ts`: bare-id keying, envelope reconstruction, subject check on read + read-before-delete, plain-JSON serialization |
| 3 | **Cross-subject thread-id takeover** (introduced by fix #2's bare-id keying, caught by orchestrator review): streaming to a foreign subject's threadId reused the id, and the store's upsert (`ON CONFLICT (id) DO UPDATE SET subject=…`) would overwrite the victim's thread | `resolve()` couldn't distinguish "doesn't exist" from "foreign-owned" (subject-filtered read) | closed **atomically at the store layer** (see review round below): the `vendo_threads` upsert refuses cross-subject flips (`WHERE subject = EXCLUDED.subject` + empty-RETURNING → conflict), mirrored on overlay paths; agent `resolve()` keeps a single ownership-blind pre-check as the friendly fast path. Victim-row-intact proven by SQL at unit + cross-block level, including a persist-time (mid-turn) takeover attempt |

Positive drill finding: the **format-evolution seam genuinely opens** — `uiPayloadSchema` passthrough, `validateTree`/`validateAppDocument` dispatch by tag, ui's extensible `registerTreeRenderer` registry + contained notice for unregistered tags, apps' byte-identical payload passthrough. Proven with a throwaway `vendo/tree@2-drill` format (tests only; grep-verified absent from all product src + dist).

## Coverage (before → after, by block)

| Block | New e2e | Headline behaviors newly exercised |
|---|---|---|
| core | +52 (129 total) | descriptorHash vs independent JCS+SHA-256 oracle; zod completeness sweep (every contracted type has its schema — zero missing); id-prefix schemas; `org` principal rejected; GrantConstraint op matrix; error-taxonomy forward-compat; format-drill seam proofs |
| store | +18 (88 total) | GCM nonce non-determinism; app-data-stays-plaintext-under-encryption; retention `DELETE` path; jsonb column assertions; EXPLAIN-verified GIN containment join; ephemeral runs/approvals zero-disk; exact/constrained grant round-trip; `vendo_state` routing suite |
| guard | +4 (123 total) | **descriptorHash drift → lapse → re-approval full cycle (both grant rows SQL-asserted)**; per-principal `maxCallsPerMinute` + per-run `maxWritesPerRun` through `bind` with `decidedBy:"breaker"` audit; critical unsuppressible vs grant+rule+judge simultaneously |
| agent | +3 (24 total) | stream Response consumed by the ai-SDK's own client reducer (both Vendo data parts survive); thread-id conflict refusal |
| actions | +3 (44 total) | fail-closed extraction proven through to the runtime surface (disabled = unlisted + unexecutable); away attaches only actAs headers (present-material leak-check via real fixture echo route); descriptorHash post-merge at the runtime boundary |
| apps | +13 (116 total) | ladder walk 1→2→3 with `open()`-answers-from-last-state at each rung; failing escalation keeps old rung serving; rung-4 resuming→http cover; doctored artifact (doc + zip) cannot take over a victim app id; two-principal state isolation; away-presence tool-proxy authority; format-drill passthrough |
| automations | +10 (37 e2e in fixture) | kill switch mid-run (held fn, owner-scoped, terminal-conflict); cron backlog collapses to one firing; `at` one-shot; revoke-parks-next-run; `fn:` step against a real machine incl. error-envelope hard-fail; agentic dryRun plans full surface, executes nothing (host-state proof) |
| ui | +25 vitest / +12 browser (76 + 49) | containment quartet in real Chromium (erroring node, dangling→skeleton→swap, unknown formatVersion direct + in-thread, pin fallback); keyboard-complete on all chrome surfaces incl. automations kill + approval deny; XHR/WebSocket exfil probes; computed-style theme proof; SSR renderToString of every surface; collision-safe ephemeral-port harness |
| cross-block (new `fixtures/chat-e2e`) | +9 | real agent + real guard + real store: full approval round-trip (ask → part → SQL pending → decide+remember → resume → grant → second call runs grant-authorized); deny; blocked-told-to-model; grant drift through the full chat stream; ephemeral zero-disk full turn; away runner park + 05§6 authority boundary; thread scoping + takeover refusal |

Full per-behavior coverage tables (contract § → before/after test) are in the lane reports; available on request.

## Escalations / notes on the record

- **01 §8 fn:-without-machine**: `validateTree` structurally cannot see `server`; the rule is enforced at `validateAppDocument` (which owns it). Tested from both sides; recommend the contract doc attribute the rule to the app-document layer (doc-only, no code change).
- **Layering vs e2e**: the dependency-guard (correctly) forbids block tests importing sibling blocks — cross-block e2e lives in exempt `fixtures/*` packages (`chat-e2e` new, `automations-e2e` extended). The umbrella wave inherits the same pattern.
- **OBS-1**: a blocked chat call audits 2× `policy-decision` (agent consults `check` via `needsApproval`, then bound `execute` re-runs the pipeline). Spans the agent↔guard seam; recorded, not fixed here.
- **03 §3 catalog/theme**: v0 agent injects no catalog/theme summary (umbrella folds it into `system.instructions`); no venue-conditional behavior exists to test in-block.
- `@vendoai/apps` `./testing` subpath is not exported; `fixtures/automations-e2e` carries a local `SandboxAdapter`. Adding the export is a clean follow-up.

## Dual-review round (8-angle adversarial self-pass + independent Codex review)

The review round on the wave commit found and fixed 11 further defects before this PR opened (second commit):

**Blockers**
- **Thread takeover was still open as a TOCTOU** — the resolve-time check left the whole streaming turn as a race window, and `threadStore.put`/`records("vendo_threads").put` were unprotected. Moved the refusal into the store's upsert itself (atomic, covers every caller).
- **Ephemeral state disk leak**: `stateStore.put` never registered the ephemeral subject (only `get` did), so a subsequent seam write INSERTed ephemeral data durably — split-brain between overlay and DB. Now registers on put/delete.
- **Backfill data destruction**: the `vendo_state` relocation ran unconditioned on every boot and its DELETE swept rows the INSERT never relocated (colon-less ids; mixed-version writes lost to `DO NOTHING`). Now version-keyed (`SCHEMA_VERSION` 2, runs once), DELETE scoped to relocated rows, conflicts resolved by `updated_at` (newer legacy data wins).

**Majors**
- `threads.list()` discarded the store cursor — silent truncation at 100 threads, and page-by-`created_at` vs sort-by-`updatedAt` could drop the most recently active thread. Now follows the cursor to exhaustion.
- One malformed thread row (writable via the `threadStore` helper) crashed the whole `list()`; now validated/skipped defensively.
- `vendo_state` id mechanism redesigned: additive `STORED` generated `id` column + `created_at` column + index (02 §2 allows additive columns) — point lookups use the index instead of seq-scanning an expression, `createdAt` is stable across puts (the seam guarantee core's conformance kit enforces everywhere else), pagination cursors on immutable `created_at`, and the upsert SQL lives in one shared `putStateRow` instead of two drifting copies.
- State ids made uniquely decodable (Codex): the appId segment must be colon-free (`^app_[^:]*$`), rejecting doctored ids that could alias another app/subject's row; same predicate in the backfill.
- An apps e2e test asserted an **ungranted away read runs** — contradicting guard 05 §6 (away without an app-bound automation grant parks). The test now seeds a legitimate app-bound grant. (The apps guard-fixture intentionally doesn't model away-park; noted below.)

**Accepted / recorded, not fixed**
- `records(...).delete(id)` carries no subject (frozen core §12 signature), so a narrow double-delete interleaving (A passes its ownership read, B legitimately claims the freed id, A's stale delete lands) cannot be made atomic without a contract change. Same acceptance class as the guard's single-process `AsyncLock` caveat.
- Store-vs-memory divergence pinned by test: in memory/ephemeral mode two subjects can privately use the same thread id (per-subject maps); with a store the id namespace is global and the second subject gets `conflict`.
- OBS-1: a blocked chat call audits `policy-decision` twice (agent consults `check` via `needsApproval`, bound `execute` re-runs the pipeline) — spans the agent↔guard seam, deferred.
- Overlay-merge pagination has a collation-tie edge (localeCompare vs code-unit vs DB collation) — pre-existing, affects timestamp ties only.
- Test-infra dedup follow-ups: export the scripted model via an `@vendoai/agent` `./testing` subpath and a shared grant-seeding helper (three near-copies exist across guard tests and the two fixture packages); `@vendoai/apps` `./testing` subpath export.

## Verification

- Root gates green: `pnpm build && pnpm test && pnpm typecheck && pnpm lint` (includes dependency-guard).
- UI playwright suite green in real Chromium (49 passed; 1 skip = OPENAI_API_KEY-gated live voice).
- All live-provider tests remain env-gated and untouched (Composio, e2b/Modal, live judge/agentic/voice).
- Dual-reviewed per the wave's done-definition (Codex review + adversarial orchestrator pass; findings addressed pre-merge).

https://claude.ai/code/session_01GPBYsXHVTynv7bnja7BC2m

---

<a href="https://app.devin.ai/review/runvendo/vendo/pull/126" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---
## Summary by cubic
Hardens E2E across all blocks via public contracts against real neighbors; adds ~120 tests, fixes three product bugs, and proves the UI format‑evolution seam. Store schema advances to v2 with a safe, idempotent backfill run by `ensureSchema`.

- **Bug Fixes**
  - `vendo_state` routing: state rows landed in `vendo_records`, cascades missed, reads hit the wrong table → routed into the dedicated `vendo_state` table, shared ephemeral overlay, and a version‑keyed backfill relocates legacy rows.
  - Agent thread persistence: composite ids and JSON shape mismatched the real store → bare‑id keying with envelope reconstruction and plain‑JSON serialization; reads and deletes scoped by subject.
  - Cross‑subject thread takeover: bare‑id reuse could overwrite another subject’s row → atomic store upsert refuses subject flips; overlay mirrors the check; victim rows stay intact.

- **New Features**
  - Cross‑block `fixtures/chat-e2e`: real `@vendoai/agent` + `@vendoai/guard` + `@vendoai/store` (PGlite) with SQL assertions; covers approval round‑trip, deny/blocked, grant drift, away runner, ephemeral, and thread scoping.
  - Automations E2E: kill switch, `fn:` step against an in‑process machine sandbox, `at` one‑shot + cron backlog collapse, and agentic dry‑run planning.
  - Format evolution drill across core/apps/ui using a test‑only `vendo/tree@2-drill`: unknown tags contain to a notice, registered tags render, and stored v0 trees render identically.

<sup>Written for commit 70f6e646151a816a58b631f22d4baf6a1d68daad. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/126?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/fixtures/automations-e2e/src/emit-and-revoke.e2e.test.ts
+++ b/fixtures/automations-e2e/src/emit-and-revoke.e2e.test.ts
@@ -1,0 +1,98 @@
+/** 07 §2 host-event scoping and 07 §3 revocation:
+ *  - emit fires only the EMITTING principal's automations, even when a second
+ *    principal has an enabled automation on the identical event name.
+ *  - revoking a captured grant disarms nothing; the next away run simply parks
+ *    pending-approval (the guard binding, not a cached decision, gates the run).
+ */
+import { beforeEach, describe, expect, it } from "vitest";
+import { automationDoc, createStack, ownerCtx, resetFixture } from "./harness.js";
+import { ADA, BOB, enableAndApprove, fixtureInvoices } from "./support.js";
+
+const listTrigger = {
+  on: { kind: "host-event" as const, event: "shared.event" },
+  run: { kind: "steps" as const, steps: [{ id: "list", tool: "host_invoices_list" }] },
+};
+
+describe("host-event scoping and grant revocation", () => {
+  beforeEach(resetFixture);
+
+  it("fires only the emitting principal's automation for a shared event name", async () => {
+    const stack = await createStack();
+    try {
+      const adaApp = "app_shared_ada";
+      const bobApp = "app_shared_bob";
+      await stack.putApp(ADA.subject, automationDoc({ id: adaApp, trigger: listTrigger }));
+      await stack.putApp(BOB.subject, automationDoc({ id: bobApp, trigger: listTrigger }));
+      await enableAndApprove(stack, adaApp, ownerCtx(ADA.subject, adaApp));
+      await enableAndApprove(stack, bobApp, ownerCtx(BOB.subject, bobApp));
+
+      const adaRuns = await stack.automations.emit("shared.event", {}, ADA);
+      expect(adaRuns).toHaveLength(1);
+
+      const byApp = await stack.sql<{ app_id: string; count: unknown }>(
+        "SELECT app_id, COUNT(*)::int AS count FROM vendo_runs GROUP BY app_id ORDER BY app_id",
+      );
+      expect(byApp.map(({ app_id, count }) => ({ app_id, count: Number(count) })))
+        .toEqual([{ app_id: adaApp, count: 1 }]);
+
+      // Bob's identically-named automation only fires for Bob's own emit.
+      const bobRuns = await stack.automations.emit("shared.event", {}, BOB);
+      expect(bobRuns).toHaveLength(1);
+      expect(Number((await stack.sql<{ count: unknown }>(
+        "SELECT COUNT(*)::int AS count FROM vendo_runs WHERE app_id = $1",
+        [bobApp],
+      ))[0]?.count)).toBe(1);
+    } finally {
+      await stack.close();
+    }
+  });
+
+  it("parks the next run after its standing grant is revoked", async () => {
+    const stack = await createStack();
+    try {
+      const appId = "app_revoke_park";
+      const ctx = ownerCtx(ADA.subject, appId);
+      await stack.putApp(ADA.subject, automationDoc({
+        id: appId,
+        trigger: {
+          on: { kind: "host-event", event: "invoice.autosend" },
+          run: { kind: "steps", steps: [{ id: "send", tool: "host_invoices_send", args: { id: "event.id" } }] },
+        },
+      }));
+      await enableAndApprove(stack, appId, ctx);
+
+      // First run: the captured grant authorizes the away send.
+      const first = await stack.automations.emit("invoice.autosend", { id: "inv_0003" }, ADA);
+      expect((await stack.automations.runs.get(first[0] ?? "", ctx))?.status).toBe("ok");
+      expect((await fixtureInvoices()).find(({ id }) => id === "inv_0003")?.status).toBe("open");
+
+      // Revoke the standing automation grant.
+      const grants = await stack.guard.grants.list(ADA);
+      const grant = grants.find((entry) => entry.appId === appId && entry.tool === "host_invoices_send");
+      if (!grant) throw new Error("automation grant not found");
+      await stack.guard.grants.revoke(grant.id, ADA);
+      expect((await stack.sql<{ revoked_at: unknown }>(
+        "SELECT revoked_at FROM vendo_grants WHERE id = $1",
+        [grant.id],
+      ))[0]?.revoked_at).toBeTruthy();
+
+      // Next run parks — revocation disarmed nothing, the run just asks again.
+      // A parked run never executed the tool: the send outcome is the
+      // pending-approval the guard returned in place of running it.
+      const second = await stack.automations.emit("invoice.autosend", { id: "inv_0003" }, ADA);
+      const secondId = second[0];
+      if (!secondId) throw new Error("second emit did not return a run id");
+      const secondRun = await stack.automations.runs.get(secondId, ctx);
+      expect(secondRun?.status).toBe("pending-approval");
+      expect(secondRun?.steps.at(-1)).toMatchObject({ tool: "host_invoices_send", outcome: "pending-approval" });
+      const away = (await stack.guard.approvals.pending(ADA)).find((request) =>
+        request.call.tool === "host_invoices_send"
+        && request.ctx.presence === "away"
+        && request.ctx.appId === appId
+      );
+      expect(away).toBeDefined();
+    } finally {
+      await stack.close();
+    }
+  });
+});
--- a/fixtures/automations-e2e/src/fn-machine.e2e.test.ts
+++ b/fixtures/automations-e2e/src/fn-machine.e2e.test.ts
@@ -1,0 +1,123 @@
+/** 07 §4 fn: steps through the REAL apps runtime against an in-process machine
+ * (fn-sandbox.ts). Proves the deterministic v0 rule end to end without a live
+ * key: a steps pipeline whose fn: step reaches a machine's POST /fn/<name>,
+ * with the event in its args, the { result } consumed by a later host tool via
+ * JSONata, and a machine { error } envelope hard-failing the run.
+ */
+import { beforeEach, describe, expect, it } from "vitest";
+import type { Json } from "@vendoai/core";
+import { automationDoc, createStack, ownerCtx, resetFixture } from "./harness.js";
+import { ADA, enableAndApprove, fixtureInvoices } from "./support.js";
+import { fnSandbox, type FnResponse } from "./fn-sandbox.js";
+
+function machine(name: string, args: Json): FnResponse {
+  const record = (args ?? {}) as Record<string, Json>;
+  if (name === "main") {
+    // 06 §4.1 success envelope: exactly one of { result } | { ui }.
+    return { status: 200, body: { result: { echo: record.note ?? null, from: "machine" } } as Json };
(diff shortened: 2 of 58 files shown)


yousefh409:
## External review triage (round 1: Devin + Greptile + cubic)

| Finding | Source | Disposition |
|---|---|---|
| Thread delete ownership race (stale bare-id delete can remove a re-claimed id) | Devin 🟡 + Greptile P1 | **Accepted, recorded** — already in the PR body's accepted-races section: core §12's `delete(id)` carries no subject, so the delete cannot be made atomic without a contract change (contracts are frozen). Same acceptance class as the guard's single-process AsyncLock caveat. The takeover-by-put path IS closed atomically at the store upsert. |
| `splitStateId` accepts empty subject (`app_demo:`) → principal-less state row | Greptile P2 | **Fixed** in 7d85f1e9 — empty subject and empty app segment both reject with `validation`; backfill predicate aligned so such legacy rows stay in `vendo_records`. Tests added. |
| Released playwright port can race before Vite binds | Greptile P2 | **Accepted** — the reservation closes the cross-lane collision that actually bit (hardcoded 4173); the residual close-to-bind gap is the standard ephemeral-port pattern and single-machine only. Noted in the PR body. |
| `vendo_state.created_at` nullable, relies on write paths | Devin ℹ️ | **Hardened** in 7d85f1e9 — `DEFAULT now()` so direct host INSERTs can't produce a NULL pagination cursor. Test added. |
| Id scheme change / list pagination perf / backfill semantics / ephemeral ordering / splitStateId analysis notes | Devin ℹ️ ×5 | Informational — behavior is intentional and covered by tests; list() exhausting the cursor is the contract-correct trade (03 §5 list is complete), with a title-projection follow-up noted in the PR body. |

CI + both reviewers green on the previous round; this comment covers the new commit's disposition.

https://claude.ai/code/session_01GPBYsXHVTynv7bnja7BC2m

PR #136, opened 2026-07-13
Permanent cross-block integration suite: real composed umbrella, whole-product journeys, required CI gate

yousefh409:
## Why

Wave-4 and the composition wave each proved the same lesson: blocks pass their own tests but break when composed (5 latents in wave-4, 8 in composition). This wave generalizes that into a **standing CI gate**: a fixtures suite that boots the REAL composed umbrella — `createVendo` from `@vendoai/vendo` — and drives whole-product user journeys end-to-end through the public wire, asserting real side effects. Every existing fixture suite composes blocks by hand; nothing on main tested the actual composition unstubbed until now.

## What

**`fixtures/integration`** — the node suite (10 journeys). The harness passes only what a real host passes (model, principal resolver, store, actAs, policy); host tools load through the real `.vendo/tools.json` + `.vendo/policy.json` contract from cwd (a load path no composed test exercised); the wire is served on a loopback `node:http` origin and driven with fetch/SSE; a scripted deterministic `LanguageModel` feeds both the agent loop and the apps generation engine; asserts are raw SQL on the public `vendo_*` tables, host-app HTTP state, and wire GETs.

- **J1** chat turn generates an app (vendo_apps row, list/open, cross-user invisible)
- **J2** destructive approval round-trip (park → decide+remember → real host DELETE → grant authorizes the next call; cross-user decide rejected)
- **J3** app edit + history undo through the wire
- **J4** automation lifecycle: enters via `POST /apps/import` (.vendoapp, id re-minted) → enable captures approvals → standing app-bound automation grants → fires via BOTH public trigger paths (`POST /tick` bearer + `vendo.emit`) → away run hits the host app for real through actAs → runs/dry-run/disable
- **J5** away park/resume/revoke: capture-deny parks the run, wire decide resumes with the deferred side effect, `DELETE /grants/:id` parks the next run, chat-source grants never authorize away runs — all at the composed level
- **J6** MCP door round-trip with a real MCP SDK client, door composed from the umbrella's own parts (the `createVendo({mcp:true})` hookup is unlanded — see Known gap): discovery → OAuth → tools/list exact-set → in-band destructive park → **one approvals plane** (wire decision authorizes the MCP retry) → `venue='mcp'` + door-auth audit + door state tables via SQL

**`fixtures/integration-browser`** — the real-browser leg (J7): Playwright Chromium drives the shipped React surface (`VendoRoot` + `VendoThread` chrome + `useApps`) against the same real umbrella + host app: chat streams, a destructive call parks on policy, the in-thread ApprovalCard approve resumes and executes the real host DELETE, the generated app lists. Deterministic; no live keys.

**CI** — a dedicated `integration` job in ci.yml (scoped turbo build → node suite → chromium install → browser suite → failure-trace artifact). To be added to branch-protection required checks after merge.

## Known gap the suite surfaces (for the MCP hookup wave, not fixed here)

J6 documents a real composition gap as `it.fails` (executable spec): **present MCP host-route tool calls have no way to authenticate to the host API under the composed umbrella.** The door correctly mints a session-less `venue:"mcp"` context (10-mcp §3 forbids forwarding the MCP bearer), but the frozen contracts define no present-session host-auth seam — `actAs` is reserved for grant-authorized away execution (04 §4). An earlier commit in this PR tried to bridge it via `actAs`; **dual + external review (Codex live-probe, adversarial self-pass, Devin) showed that was a net-negative** — it skipped the origin-trust check (credential exfiltration to an untrusted origin) and violated the frozen `ActAs` type — so it was reverted. Designing the seam is a contract change owned by the unlanded `createVendo({mcp:true})` hookup (docs/contracts/10-mcp-umbrella-hookup.md); the J6 `it.fails` cases hand that wave an exact, executable target. Everything else in J6 (discovery, OAuth, tools/list, in-band destructive park, one approvals plane, apps ride-along, SQL) passes.

MCP-Apps `_meta.ui` ride-along is also shadowed in composition (the umbrella already registers `vendo_apps_open` as an agent tool; the door only adds ride-along tools for absent names) — expected per the door's design, worth revisiting with the hookup.

## Verification

- Both suites ≥5 consecutive green full runs locally + a `--sequence.shuffle` run proving no order-dependence (flake gate; no test retries anywhere)
- Root gates green: build / test / typecheck / lint
- Neighbor suites green: mcp, mcp-e2e, chat-e2e, automations-e2e, redteam-e2e
- Frozen contracts untouched (`git diff c7039240..HEAD -- docs/contracts` is empty); `packages/actions` byte-clean vs pre-wave after the revert
- Dual review (Codex + adversarial self-pass) + external (Devin/Greptile) triaged in a comment below

https://claude.ai/code/session_01LvTv8VGGYDaUHvCxvW8RoH

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,3 +22,30 @@ jobs:
       - run: pnpm test
       - run: pnpm typecheck
       - run: pnpm lint
+
+  integration:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        with:
+          node-version: 22
+          cache: pnpm
+      - run: pnpm install --frozen-lockfile
+      # Build only the workspace deps the two integration suites need (the `...`
+      # suffix pulls each fixture package's transitive workspace deps).
+      - run: pnpm turbo run build --filter=@vendoai-fixtures/integration... --filter=@vendoai-fixtures/integration-browser...
+      # The node suite boots the real composed umbrella over HTTP against the host app.
+      - run: pnpm --filter @vendoai-fixtures/integration test
+      # The browser leg drives the shipped hooks/chrome in Chromium against the same wire.
+      - run: pnpm --filter @vendoai-fixtures/integration-browser exec playwright install --with-deps chromium
+      - run: pnpm --filter @vendoai-fixtures/integration-browser test:browser
+      - if: failure()
+        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4
+        with:
+          name: integration-browser-playwright-trace
+          path: fixtures/integration-browser/e2e/test-results
+          if-no-files-found: ignore
+          retention-days: 7
--- a/docs/superpowers/plans/2026-07-12-cross-block-integration-suite.md
+++ b/docs/superpowers/plans/2026-07-12-cross-block-integration-suite.md
@@ -1,0 +1,123 @@
+# Cross-Block Integration Suite (fixtures/integration) — Wave Plan
+
+**Goal:** A permanent, CI-required integration suite that boots the REAL composed
+umbrella (`createVendo` from `@vendoai/vendo`) against the fixture host app and
+drives whole-product user journeys end-to-end through the public wire and hooks,
+asserting real side effects. Wave-4 and composition each found latent bugs that
+block-local tests missed (5 and 8 respectively); this suite makes that class of
+bug a standing CI catch instead of a per-wave discovery.
+
+**Why this is different from every existing fixture suite:** chat-e2e,
+automations-e2e, mcp-e2e, and redteam all compose blocks BY HAND ("the way the
+umbrella will compose them"). Nothing on main tests the actual composition —
+`createVendo` + its wire — with the blocks unstubbed. `packages/vendo/server.test.ts`
+stubs block methods (routing coverage) and runs exactly one real chat turn.
+
+**Scope discipline:** additive only — `fixtures/integration` + a CI job. Frozen
+contracts (docs/contracts) are law and are NOT modified. Real cross-block bugs
+found by the suite are fixed in the owning block, each with a permanent
+regression test here.
+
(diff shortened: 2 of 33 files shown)


yousefh409:
## Reviewer triage (dual internal + external)

Four reviewers ran on this branch. Their findings converged on one thing — the `venue:"mcp"` `actAs` branch was wrong — which is now **reverted**; the remaining real findings are **fixed**.

| # | Source | Severity | Finding | Disposition |
|---|--------|----------|---------|-------------|
| 1 | Codex (live-probe) | P0 | The MCP `actAs` branch attached host credentials **without the `mayForwardPresentHeaders` origin check** → host cookie/bearer sent to an external OpenAPI origin | **Fixed by revert** — `packages/actions` restored byte-clean to pre-wave (`git diff c7039240 -- packages/actions` empty) |
| 2 | Codex + self-pass + **Devin** | P0/P1 | The branch passed `undefined as PermissionGrant`, violating the FROZEN `ActAs` seam (04 §4 reserves `actAs` for grant-authorized away) — crashes any host that reads the grant | **Fixed by revert** |
| 3 | self-pass | P0 | **J5 never called `resetFixture`** and read seed invoices J2 deletes → order-dependent flake | **Fixed** — `resetFixture()` at the top of every `it` in J5 + J4; proven under `--sequence.shuffle` (J5-before-J2 green) |
| 4 | Codex | P1 | J6's retried destructive call could pass vacuously (tolerated host errors) | **Fixed** — now asserts `isError` + the `→ 401` host-auth signature (the current known-gap failure mode), becomes a real-deletion assert when the hookup lands |
| 5 | Codex | P2 | J6 "verbatim" tools/list was a subset check | **Fixed** — asserts the exact set = bound registry ∪ the door's three ride-along tools (rejects any unguarded extra surface) |
| 6 | Codex | P2 | Browser smoke didn't prove the wire was reached | **Fixed** — drives the `useApps` wire GET and asserts a real response |
| 7 | Codex | P2 | Ephemeral-port `freePort`-then-close race | **Won't-fix (out of scope)** — this is the established pattern across all sibling fixtures (automations-e2e, mcp-e2e, packages/ui) and is not wave-introduced |
| 8 | Greptile | — | Approved 5/5 (T-Rex replayed the browser journey) | Noted — it did **not** catch #1/#2; the stricter internal reviews did, which is why both ran |

The root disposition for #1/#2: present MCP host-route auth has **no seam in the frozen contracts**. That's a real gap, but its correct fix is a new present-session host-auth seam — a contract change owned by the unlanded `createVendo({mcp:true})` hookup. The suite now hands that wave an executable spec (J6 `it.fails`) rather than shipping a wrong fix. Post-revert green across both suites 5x + shuffle, neighbors, and root gates.

PR #171, opened 2026-07-14
fix(apps): isolate sandbox forks and durable resumes

yousefh409:
## What was broken

- E2B `snapshot()` paused a live sandbox and returned its sandbox id, while `resume()` reconnected to that same id. `AppsRuntime.fork` therefore produced two documents backed by one live machine, so files/processes could bleed between source and fork. The same behavior let `machines.withFork` pause the machine still serving the live cache.
- Modal kept env, egress, and port metadata only in adapter-local Maps. Resuming an image ref in a fresh process silently fell back to empty env and port 8080, losing secret handles and dropping the egress allowlist fail-open.
- The shared adapter conformance suite only covered the fake, whose adapter-local snapshot state masked the fresh-process failure class.

The regression tests were written first and failed against the old behavior: E2B returned/reconnected to the source sandbox id, Modal lost configured state in a fresh adapter instance, and the fake could not resume a provider ref across adapter instances.

Security review added a second red test for Modal: it models the root app winning a rewrite race after the adapter's last state write but before filesystem capture. The old manifest-based implementation resumed without outbound restrictions when the forged state omitted `egress`.

## Fix

This remains entirely inside the frozen `SandboxAdapter` seam; neither `packages/apps/src/sandbox.ts` nor `docs/contracts/06-apps.md` changed.

- E2B creates a provider point-in-time snapshot and encodes its snapshot id plus adapter-owned egress/port metadata in an opaque versioned `e2b:v1:` ref. Each resume creates a distinct sandbox from that checkpoint, so source and fork remain independent and the source is never paused.
- Modal now mirrors that design: `snapshot()` returns an adapter-generated `modal:v1:` ref containing the image id, env, egress, and port. Secret env entries are opaque handles, never real secret values. Fresh adapters restore directly from the ref.
- The app-writable Modal state manifest and the probe sandbox were deleted. App filesystem tampering can no longer remove `egress` or weaken the resumed sandbox's provider-native network policy.
- The conformance suite covers fork independence/source liveness and fresh-adapter env/egress/port durability. Real E2B and Modal suites run it behind credential gates.
- The fake models provider-persistent snapshot refs across adapter instances and preserves egress, so it no longer masks either original bug class.

## Compatibility / migration

Accepted compatibility break: pre-existing `e2b:<sandbox-id>` and `modal:im_<image-id>` refs are intentionally not migrated. Resume rejects them with validation errors so they fail loudly as unresumable instead of reconnecting or restoring without the required policy metadata. New refs use `e2b:v1:` and `modal:v1:`. Legacy `modal:sb_` refs are also rejected because they do not contain durable runtime policy.

## Validation

Scoped validation after the security rework:

- `@vendoai/apps`: 25 files passed, 152 tests passed; 9 credential-gated tests skipped in the ordinary package run
- `pnpm --filter @vendoai/apps typecheck`: passed
- focused Modal tampering suite: 6/6 passed
- `git diff --check`: clean
- no full root-gate rerun, per review instruction; the prior root-gate result and caveat below stand

The earlier repository gates remain green:

- `pnpm build`: 18/18 tasks successful
- `pnpm typecheck`: 34/34 tasks successful
- `pnpm lint`: dependency guard OK; 2/2 lint tasks successful

Real E2B conformance rerun:

```text
✓ creates, executes, snapshots, resumes, serves, and stops through the frozen seam 26109ms
✓ forks into an independent machine without disrupting the live source 9175ms
✓ restores env, egress, and port through a fresh adapter instance 10373ms
✓ creates, serves, snapshots, forks, serves again, and kills 3115ms
Test Files 1 passed (1)
Tests 4 passed (4)
Duration 76.74s
```

Real Modal conformance rerun:

```text
✓ creates, executes, snapshots, resumes, serves, and stops through the frozen seam 13924ms
✓ forks into an independent machine without disrupting the live source 11410ms
✓ restores env, egress, and port through a fresh adapter instance 12967ms
✓ creates, serves, snapshots disk, restores a new machine, serves again, and terminates 7854ms
Test Files 1 passed (1)
Tests 4 passed (4)
Duration 58.98s
```

## Root-test fleet caveat

The final root test command was run exactly once with bounded resources:

```text
nice -n 19 env TURBO_CONCURRENCY=2 VITEST_MAX_THREADS=2 VITEST_MIN_THREADS=1 VITEST_MAX_FORKS=2 VITEST_MIN_FORKS=1 pnpm test

Tasks: 32 successful, 35 total
Failed: @vendoai/redteam-e2e#test
Time: 14m8.503s
```

The host's one-minute load was approximately 329 from external fleets. The failures were confined to the untouched red-team HTTP fixture boundary: `loginCookie()`/run polling reported `ECONNRESET`, `EPIPE`, `fetch failed`, and timeouts. An isolated run of untouched `shared-app-dormant.e2e.test.ts` produced 2 passes and 2 failures, both at `loginCookie()` with `read ECONNRESET`; that harness does not import or use the apps fake sandbox. An earlier unbounded root run was killed with exit 137 under the same fleet contention.

Per orchestrator direction, this PR keeps that root-gate caveat rather than retrying into the external load storm. The changed apps package, fake conformance, and both real-adapter conformance runs are green.


yousefh409:
Orchestrator review (verified against code, not the PR text):

**Verdict: merge after one must-fix.**

- **MUST-FIX (security)** — `modal/index.ts`: the runtime-state manifest at `/app/.vendo-modal-state.json` is app-writable and the app runs as root. The app can overwrite it between `persistState()` and `snapshotFilesystem()`; `deserializeState` is shape-only and an omitted `egress` key decodes to `undefined` = unrestricted outbound. An egress-restricted app can therefore mint itself open internet on the next fork/resume. Fix: encode `{imageId, env(handles), egress, port}` in an opaque adapter-generated `modal:v1:` ref, mirroring the E2B design; drop the FS manifest + probe sandbox.
- **SHOULD-FIX** — add a tampering regression test (app forges/deletes state before snapshot → fresh-adapter resume still denies egress).
- **SHOULD-FIX (compat)** — legacy `e2b:<id>` / pre-manifest `modal:im_` refs now fail loudly as unresumable: fail-loud is right, but document the accepted break in the PR body.
- **Follow-up (pre-existing, out of scope)** — non-empty Modal domain allowlist leaves `outboundCidrAllowlist` unset; if Modal permits raw-IP egress in that state the domain filter is bypassable. Tracking as a Linear issue.
- **Verified OK**: E2B ref design, exec error mapping (non-swallowing), probe always-terminated (before its removal), import path strips hostile `server` refs, secret handles only (no real values persisted), fake-sandbox alignment breaks no existing tests, frozen seam untouched.

E2B half and both new conformance axes are genuinely regression-proof for the original bugs. Root-gate caveat (redteam-e2e ECONNRESET under external load ~330, untouched fixture boundary) accepted per fleet conditions.

PR #173, opened 2026-07-14
fix(apps): vendo_apps_edit blank-iframe + tree-ops fast-path hardening (ENG-245)

yousefh409:
## Summary

- make the tree-ops edit dialect strict and repairable instead of accepting malformed operations that can produce invisible/orphaned content
- let the runtime's outer repair loop actually run after the engine exhausts its inner attempts, while carrying cumulative, actionable repair issues across all attempts
- return an additive structured `edit-rejected` failure that tells the agent to retry `vendo_apps_edit` on the same app rather than rebuild with `vendo_apps_create`
- reject newly empty rooted layouts before persistence and render a contained tree-level notice for an already-empty layout

## Root cause

ENG-245 had two independent causes in the observed edit path.

First, model-generated tree ops were not validated strictly enough. Real malformed classes included a nested `{ component: { name, source } }` payload, duplicate-root `add-node`, unsupported `position` aliases, placement fields nested inside `node`, and out-of-range indexes. In particular, nesting `parentId` inside `node` let `add-node` persist an unattached orphan: the document validated, but the rooted view never displayed the added component.

Second, retries were nested but the outer loop was effectively dead. `engine.editTree` used its two attempts and returned `kind: "failure"`; `runtime.edit` then returned immediately on the first outer attempt instead of using its second attempt with the engine's repair feedback. Both layers also replaced repair issues with only the latest failure, so the model could lose earlier actionable context.

This change validates exact op shapes with indexed errors, requires every added node to attach to the rooted view, rejects gap indexes and newly empty rooted layouts, preserves component sources, accumulates distinct repair issues, and continues through the runtime repair loop. The frozen edit response shape is unchanged; `failure` is additive.

## Browser evidence

Playwright drove the real `demo-bank` UI against the provisioned local environment. It opened existing app `app_4a5dccd2-773d-4114-a11b-ea8e42c4bf67`, sent “Add a filter dropdown above the visual spending dashboard and chart … Edit this existing app in place; do not rebuild it,” approved once, and asserted:

- app count stayed `1 -> 1`
- app id stayed unchanged
- document hash changed `3334052be30c2067 -> 50019afeb5f1cbec`
- the only mutation tool card added was `vendo_apps_edit` (no `vendo_apps_create`)
- added node `filter` rendered at `1023 x 150`; generated `PeriodFilter` iframe rendered at `1023 x 150`
- run passed in 84.6 seconds with one approval

Before edit:

Before the in-place edit

After the same app was edited and reopened:

After the in-place edit

The final local evidence manifest is `artifacts/eng-245/evidence/2026-07-14T22-05-00-206Z-summary.json`. The two frames are hosted from the dedicated `evidence/eng-245-edit-reliability` branch so browser artifacts remain outside this PR's code commits.

## Residual gap: jail null-render - handed to gen-fidelity

`packages/ui/src/tree/jail/runtime-entry.tsx:126-140` loads the component and calls `root.render(...)`, but the async function resolves without checking whether React committed any DOM. The message handler at `packages/ui/src/tree/jail/runtime-entry.tsx:163-166` consequently posts `{ kind: "ready" }` from `.then(...)` even when the component rendered nothing.

Minimal repro:

```tsx
export default function Blank() {
  return null;
}
```

Send that source in a jail `render` message. The component throws no error, the mount stays empty, and the jail reports `ready`. This PR does not touch `packages/ui/src/tree/jail/*`; the null-render acceptance remains handed to gen-fidelity.

## Verification

Passed:

- `pnpm build` — 18/18 tasks
- changed apps suites with one Vitest worker — 3 files, 30/30 tests
- `@vendoai/ui` suite in the serialized repo run — 13 files, 77/77 tests
- Express corpus e2e with one Vitest worker — 6/6 tests
- `pnpm run typecheck --concurrency=1` — 34/34 tasks
- `pnpm run lint --concurrency=1` — dependency guard OK; 2/2 lint tasks
- Playwright real-browser edit flow described above
- GitHub CI — exact `pnpm build`, `pnpm test`, `pnpm typecheck`, and `pnpm lint` sequence passed; integration, perf, audit, and changeset-status also passed

Local host note: the unconstrained local `pnpm test` run timed out five Express e2e files and one unrelated UI thread test while many other worktrees were simultaneously running Vitest/Next builds. A serialized local retry cleared the UI failure (77/77) but four unrelated `packages/vendo/src/server.test.ts` cases crossed their 30-second timeout; that file took 606 seconds under load. The Express suite subsequently passed 6/6 in isolation, all changed-package suites passed locally, and the clean GitHub runner then passed the full root gate including `pnpm test`.

### DIFF
--- a/packages/apps/src/agent-tools.test.ts
+++ b/packages/apps/src/agent-tools.test.ts
@@ -70,6 +70,7 @@ describe("apps agent tools", () => {
     expect(descriptors.map((descriptor) => descriptor.risk)).toEqual([
       "read", "write", "read", "read", "write", "write",
     ]);
+    expect(descriptors.find(({ name }) => name === "vendo_apps_edit")?.description).toMatch(/retry.*same app/i);
   });
 
   it("classifies only provable tree edits as read-class", async () => {
@@ -153,6 +154,39 @@ describe("apps agent tools", () => {
       tool: "vendo_apps_edit",
       args: { appId: "app_foreign", instruction: "Make the heading blue" },
     }, ctx)).resolves.toBe("write");
+  });
+
+  it("surfaces a structured retryable edit failure instead of implying the app changed", async () => {
+    const broken = JSON.stringify({
+      ops: [{ op: "set-prop", nodeId: "missing", prop: "value", value: 1 }],
+    });
+    const runtime = createApps({
+      store: memoryStore(),
+      guard: guardFixture(),
+      tools: hostTools,
+      catalog: [],
+      model: scriptedLanguageModel(generated, broken),
+    });
+    const created = await runtime.create({ prompt: "Build a dashboard" }, ctx);
+
+    const outcome = await runtime.agentTools().execute({
+      id: "call_edit_failure",
+      tool: "vendo_apps_edit",
+      args: { appId: created.id, instruction: "Change a missing card" },
+    }, ctx);
+
+    expect(outcome).toMatchObject({
+      status: "ok",
+      output: {
+        app: created,
+        failure: {
+          code: "edit-rejected",
+          retryable: true,
+          message: expect.stringMatching(/same app/i),
+        },
+        issues: expect.arrayContaining([expect.stringContaining("missing")]),
+      },
+    });
   });
 
   it("creates and opens an app through the guard-bound fixture", async () => {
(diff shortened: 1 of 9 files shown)


yousefh409:
Cross-link: the residual gap documented below (jail null-render accepted as ready) is now covered by gen-fidelity's #177 — explicit empty-mount signal from runtime-entry (flushSync commit + mount.hasChildNodes classification) with a regression test using the exact Blank-returns-null repro from this PR. Layering: this PR's tree-level check catches empty trees before render; #177 catches non-empty trees whose jailed component renders null.

PR #283, opened 2026-07-15
demos/vendo: remixable wiring in Maple+Cadence, init remix offering, remix + in-client promotion demos (ENG-288 M6)

yousefh409:
## What

ENG-288 **M6 — the final apps-block remix milestone**: remixable wiring in both demo hosts, a genuine `vendo init` remix offering, and the two mandatory captured demos (slate items 2 and 3), on the real demo hosts with the real model.

**Demo wiring**
- **Maple**: new presentational `NetWorthView` (self-contained — inline styles, hand-rolled monotone SVG area chart, own count-up + range control) used for real by the home net-worth card, registered `remixable + exportable` with seed-matching `sampleProps` (slot `MapleNetWorthCard`), plus a `MapleNetWorthCard` catalog entry, a `/vendo/apps` host page (apps open OUTSIDE the thread with an edit box and a ship-review panel) and an Apps nav item.
- **Cadence**: the dashboard's `MissingDocsHero` extracted into a self-contained remixable component (slot `CadenceMissingDocsHero`) with seed-matching `sampleProps` + catalog entry.
- Both slots capture statically with **zero `unresolvedPins`** (`vendo sync` exit 0); baselines committed.

**Init codemod remix offering** (audit close-out — counters were hardcoded 0)
- New `scanRemixRegistrations` in `@vendoai/actions`: every statically capturable `{ name, component }` registration site, wrapped or not (router tables with a `path` field excluded).
- `vendo init` now proposes permission-gated `remixable: true` wraps as ordinary confirmed code changes, re-syncs immediately so approved wraps capture right away, and reports real `remixOffered` / `remixWrapped` / `remixSkipped` telemetry. Tests for offer-approve-capture, decline-and-skip, and idempotent re-runs.

**Product bugs the demo wiring exposed — fixed here (small, flagged)**
1. `actions` sync: `@/` imports ignored the host's tsconfig `paths` (demo hosts map `@/*` → `src/*`, not the root). Now tsconfig-first with the old convention as fallback.
2. `actions` sync: on semicolon-free (Prettier `semi: false`) hosts, `export interface X {}` swallowed every later export in the fallback module lexer, and the `import type` eraser was greedy across newlines and deleted real imports (including the root-CSS import). Both line-scoped now, with tests.
3. `ui`: `InClientMount` didn't mirror the jail's `sampleProps` rehearsal fallback — a hash-pinned, prop-less fork rendered in rehearsal but **crashed after promotion**. Venue parity restored + unit test.
4. `vendo` wire: unexpected errors were swallowed into a bare 501 "Internal Vendo error" with no server-side trace; now logged for the host operator (response unchanged).

## Why

Locked decisions 3 and 8 of the apps-block design (docs/superpowers/specs/2026-07-14-apps-block-design.md): the remix accuracy bar is visual parity vs the host original on the real demo hosts, and all demos are mandatory, real-browser, captured. This closes the last unchecked ENG-288 items.

## Demo capture 2 — Maple remix journey + visual parity

Real Maple host under `next dev`, real Auth.js login, **real Anthropic model** emitting the `fork-pin`. The forked pin renders from the genuinely sync-captured source inside the CSP jail.

Left: host original (Maple home). Right: the remixed render (captured source in the jail):

parity side by side

remix journey

**Parity verdict (honest):** widths equalized to ≤2px; ffmpeg **SSIM 0.913**. Same layout, colors, copy, chart geometry, working range control. Residuals: the jail cannot load the host's next/font **Inter** (no network; `font-src data:`) so type falls back to the system stack, and the jail content box lands ~6px shorter, which shifts the rows SSIM compares.

## Demo capture 3 — in-client promotion end-to-end

Fork → customize (a generated `FetchProbe` button proves authority live) → ship-diff → approval injected via the documented dev seam → hash-pinned host-page mount → new version drops back, loudly:

in-client promotion

- Before approval: jail render, probe → `fetch: FAILURE (CSP)`.
- Ship-diff panel shows the pinned fork vs its baseline hash + the net-new generated component.
- After `POST /api/vendo/dev/inclient-approval`: `[data-vendo-inclient-mount]`, **zero iframes** in the surface, probe → `fetch: SUCCESS (host authority)`.
- After one more model edit (rename): drop-back to the jail with the "In-client approval invalidated" notice, probe → `fetch: FAILURE (CSP)` again.

Key frames + rerunnable scripts + README in `docs/verification/eng-288-m6/`.

## Host-visual notes (UI-affecting, screenshots)

- Maple home net-worth card is now `NetWorthView`: visually equivalent (see the parity left pane, captured from the live home page); deliberate deltas: the meaningless recharts x-axis index labels ("0","1","2"…) are gone, and the chart tooltip is a simplified hover dot + value.
- Cadence dashboard hero unchanged in place: cadence

## Parked / reported (not expanded here)

- **Named-export remixables fail at render** ("must have a React default export") even though capture succeeds — demos carry default exports + documented in docs/host-components.md; suggest a Linear issue (record the imported name in the baseline or fail loudly at sync).
- **Edit router keyword heuristic**: `instructionRequiresServer` routes any instruction containing "api"/"function"/etc. to the machine-code dialect — a plain UI ask mentioning `/api/...` fails slowly. Suggest a Linear issue.
- **PGlite scratch store is unrecoverable after an uncleanly killed dev server** (`Aborted()` at `PGlite.create`); the capture scripts park `.vendo/data` before boot. Suggest a Linear issue.
- **Init catalog picker counters** (`componentsOffered`/`componentCount`) remain 0 — that's the (unbuilt) catalog picker feature, distinct from the remix offering.

## Verification

- [x] `pnpm build && pnpm test && pnpm typecheck && pnpm lint` green locally (sole exception: the documented pre-existing `packaging.e2e.test.ts` failure under paths with spaces — reproduced byte-identical on clean `origin/main` in a spaced dir; CI is authoritative)
- [x] Screenshots attached (if UI-affecting)
- [x] Both capture scripts run end-to-end and exit 0 (`docs/verification/eng-288-m6/README.md` has the invocation)

---

<a href="https://app.devin.ai/review/runvendo/vendo/pull/283" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---
## Summary by cubic
Wires remixable components in Maple and Cadence, adds a real `vendo init` remix offering, and ships the ENG-288 M6 demos (remix parity and in‑client promotion) on the real demo hosts. Improves sync reliability and in‑client behavior.

- New Features
  - Maple: presentational `NetWorthView` registered as remixable/exportable `MapleNetWorthCard` with seed `sampleProps`; new `/vendo/apps` workspace with ship‑review; catalog updated.
  - Cadence: extracted `MissingDocsHero` as remixable `CadenceMissingDocsHero` with seed `sampleProps`; catalog updated.
  - Baselines captured with zero `unresolvedPins`; parity and in‑client promotion demos added under `docs/verification/eng-288-m6/` (real hosts, real model).
  - `vendo init`: uses `scanRemixRegistrations` from `@vendoai/actions` to offer permission‑gated `remixable: true` wraps, re‑syncs approved wraps immediately, and reports real `remixOffered`/`remixWrapped`/`remixSkipped` telemetry. Connects to ENG‑288 requirements for remix wiring and in‑client venue.

- Bug Fixes
  - `@/` imports now resolve via the host’s tsconfig `paths` during `actions` sync; legacy fallback kept.
  - Fallback lexer fixes for semicolon‑free hosts: prevents `export interface` from swallowing later exports and limits `import type` erasure to a single line.
  - In‑client mount mirrors the jail’s `sampleProps` fallback to avoid crashes on prop‑less approved nodes.
  - Wire handler logs unexpected server errors before returning the generic 501 response.

<sup>Written for commit 35eb0c858fff766514d801e3e1e51431323d49bb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/283?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Triage of Devin's 5 info-level findings (all verified non-blocking, as its own analysis concluded):

1. **Line-scoped type erasure leaves multi-line `import type` body fragments** — correct, and safe as Devin verified: the fragments match none of the specifier-finding regexes. The alternative (newline-spanning erasure) is the exact bug this fixes.
2. **`@/` now prefers tsconfig paths over the root convention** — intentional; the convention was wrong for every host that maps `@/*` to `src/*` (both demo hosts). The convention remains as fallback when no alias maps it, so no-tsconfig hosts are unchanged.
3. **Statement-boundary heuristic could over-split on a continuation line starting with `import`/`export`** — acknowledged as theoretical; it only fires at bracket-depth 0 outside strings/comments, where a bare `export `/`import ` continuation is not valid JS/TS expression syntax in practice. Over-splitting also only truncates the *previous* statement, whose exports were already recorded.
4. **Remix-wrap offset race guarded, `return []` abandons all files** — deliberate: a stale offset means the scan snapshot is untrustworthy as a whole; offering nothing (init can be re-run) beats corrupting one host file.
5. **Standalone hero badge won't track the shared `Badge`** — intentional and documented in the component comment: remixable demo components must be self-contained for faithful jail capture; the shared Badge keeps serving every other call site.

PR #428, opened 2026-07-20
execution v2 wave 7: product hardening (fn paths, iframe keepalive, warm template, flake)

yousefh409:
Wave 7 Lane H2: four product-hardening items from the execution-v2 close-out appendix, Yousef-authorized.

## 1. fn-binding envelope paths (the PR #418 em-dash class)

Graduation's rebind could keep a host tool's `/data/` response envelope in an `fn:` binding path; fn responses unwrap `{result}`, so the board rendered em-dashes until a follow-up edit. Root fix, three parts:

- **Sample the real shape**: after the rebind lands, each query-bound fn is sampled once (`POST /fn/<name>` with empty args — the exact call the query makes at `open()`; action fns wired as actions are never sampled). The derived shape card is keyed `fn:<name>`, exactly like host-tool cards.
- **Compile check extended to fn results**: sampled fn shapes feed the edit compiler's binding shape check on retries, and the landed tree is post-checked with `checkBindingShapes` (newly root-exported from core; the patch compiler path already treats `fn:`-keyed cards like host tools — pinned by a new `patch.test.ts` case). A miss becomes a per-binding repair issue; a failed sample stays defensive (never blocks a rebind).
- **Emission fix**: the rebind directive now states the `{result}` unwrap rule and forbids carrying the old tool's envelope segments.

Red-first regression (`graduation.test.ts`): envelope-path rebind must repair to `/digest/summary`, plus a defensive-sampling guard test.

## 2. Layer-3 iframe keepalive

An embedded served app died under the user when the machine idled out (Wave-4 backlog). Now:

- `runtime.machine.ping(appId)`: one cheap HEAD through the idle-tracked machine wrapper — re-arms the idle timer and slides the e2b TTL (rides the adapter's extend-on-activity). A sleeping machine wakes and reports `{state:"woke"}`; wake is shared with the follow-up `open()`.
- Wire `POST /apps/:id/machine/ping` (owner-scoped), client `apps.pingMachine`.
- `AppFrame` keepalive loop (opt-in prop, wired in `VendoSlot`/`VendoPage`): throttled ping on user activity (default one per 60s tick; iframe focus counts — in-frame activity is invisible to the host page). A `woke` ping swaps in the **existing** resuming cover and re-opens once for the fresh machine URL. No reconnect daemon.

TDD across all three layers (`served-apps.test.ts`, `served-apps-wire.test.ts`, `frames-and-jail.test.tsx`, `client.test.ts`).

**Real-browser evidence** (`docs/verification/exec-v2-wave7-h2/browser/`): real `AppFrame` embedding the real scaffold server —

| embedded + awake ping | woke → resuming cover | reloaded on fresh URL |
| --- | --- | --- |
| awake | resuming | reloaded |

## 3. Box build speed (Yousef's 4.5-minute layer-3 kanban)

Measured live (6 runs, real e2b + real Claude, harness task logs now ISO-stamped; full transcripts in `docs/verification/exec-v2-wave7-h2/`):

| config | runs | mean |
| --- | --- | --- |
| baseline (harness only) | 195.2s, 153.9s | 174.6s |
| shipped (scaffold + prompts) | 133.5s, 174.5s | 154.0s |

**Where the time actually goes**: zero `npm install` in any run (zero-dep doctrine holds, so the "+ node_modules" half of the item is empty by construction), boot ≈ 2s — the build is model round trips. The biggest single sink was writing ~14.5KB of skin-contract plumbing and sometimes fully rewriting it on a syntax bug (~97s in the baseline run). Shipped fix: a pre-baked served-app scaffold in the base template (`packages/apps/box/scaffold/`, conformance-tested in `box-scaffold.test.ts`), a copy-and-edit instruction in the served contract, and a batched-verification working style. Every warm run wrote exactly 2 app files and never touched plumbing.

**Honest numbers**: ~12% mean cut (n=2/config, high model-latency variance), 32% best-vs-worst — short of the >40% example target. The remainder is app-specific UI generation (~40s of tokens for a ~9KB page) + self-verification turns, both model-latency-bound; cutting verification depth would trade away the self-verification floor. The next real lever is a faster in-box model (`VENDO_INFERENCE_MODEL`) — a behavior decision left for Yousef, not template warming.

Rebuild templates with `node build-template.mjs` and set `VENDO_BOX_TEMPLATE` (this PR's evidence template: `02ar6qdzqils9e3hgek5`). All sandboxes destroyed; account swept to zero after the campaign.

## 4. box-harness Procfile flake (load-average 40)

Root cause, not a bigger timeout: the supervisor spawned the app entry with `bash -lc`, so every spawn sourced the host machine's login profiles — arbitrarily slow under load (the flake) and an env leak into the app. A Procfile entry is one shell line with the boundary env; now `bash -c`. Red-first test proves profile isolation (a slow/env-mutating `~/.bash_profile` never runs, nothing leaks). In the box the template's profiles were trivial, so production behavior is unchanged.

## Gate

- `pnpm build && pnpm test && pnpm typecheck && pnpm lint` green at root.
- TDD red-first on items 1/2/4; item 3 gated on live measured evidence.
- Docs synced: `docs/machine-model.md` (fn-shape sampling, scaffold), `docs/served-apps.md` (keepalive).
- Scope discipline: no adapter/TTL/store internals touched (sibling H1 owns those); the ping rides the request-path knobs that exist today.

---

<a href="https://app.devin.ai/review/runvendo/vendo/pull/428" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---
## Summary by cubic
Hardens execution v2: fixes fn-binding path issues, adds a reliable served-app keepalive with a single re-open, warms the box scaffold for faster builds, and removes a Procfile spawn flake. Improves reliability, prevents idle dropouts, and trims layer-3 build time (~12%).

- New Features
  - Fn shape sampling and checks: sample each query-bound `fn:` using the query’s landed input via `POST /fn/<name>`, re-check with `checkBindingShapes`, and state the `{result}` unwrap in rebinds. Bad envelope paths now route to per-binding repair.
  - Served-app keepalive: adds `runtime.machine.ping`, `POST /apps/:id/machine/ping`, and `client.apps.pingMachine`. `AppFrame` pings on user activity (throttled) and performs one absorbed re-open on ping failure or `{state:"woke"}`; shows the resuming cover and refreshes once. Wired in `VendoSlot` and `VendoPage`.
  - Warm box template: pre-baked scaffold at `/opt/vendo-box/scaffold` (`server.js`, `fns.js`, `index.html`, `vendo.json`, `package.json` with `"type":"module"`). Hardened scaffold: own-entry `/fn` dispatch, bounded body (413 >1MB), `HEAD /` for keepalive, and safe `vendoTheme`. Conformance test included; harness logs are ISO-stamped.

- Bug Fixes
  - Procfile flake: supervisor now spawns with `bash -c` (no login-shell profiles) to avoid host profile sourcing and env leaks; test verifies isolation.

<sup>Written for commit 335f1d686cbb0d9e0db12ebeda5cad796bd70cc8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/428?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Review triage (31f2eae1):

**Fixed**
- cubic P2 `/fn/toString` inherited-property dispatch → own function entries only (test added).
- cubic P2 unbounded `/fn` body buffering → streaming 1MB cap, 413 (test added).
- cubic P2 + Greptile P2 malformed `vendoTheme` blanks the page → try/catch to null in the scaffold page.
- cubic P2 double re-open on a rejected wake-path refresh → ping failure and woke ping now share exactly one absorbed re-open.
- cubic P2 fn sampling with `{}` instead of the landed input → samples with `query.input`, the exact call `open()` makes.
- Devin scaffold ESM syntax-detection reliance → `package.json` `type:module` added and baked into the template.
- Devin HEAD-probe 404 note → scaffold now answers `HEAD /` 200 (any request re-arms idle either way, so this is cosmetic cleanliness).
- cubic P2 evidence nits → README now documents the profile-bucket attribution caveat (BUILD wall-clock totals are the load-bearing numbers) and the `pnpm --filter @vendoai/ui build` prerequisite for the harness bundle.

**Accepted as designed (with rationale in code)**
- cubic P1 / Devin: sampling "executes arbitrary fns" — a query-bound fn fires WITHOUT user action the moment the graduated tree is opened or emitted (the progressive resolver calls it with the same input seconds later). Sampling adds no new authority, at most one extra invocation of something already about to run; an fn too dangerous to sample was already too dangerous for the model to wire as a query. Action-bound fns are never sampled.
- cubic P2 iframe-focus keeps the machine alive while focused — in-frame activity is invisible to the host page, so focus is the only observable proxy; ticks are already gated on page visibility (hidden tab = no pings), so the exposure is a visible, focused, untouched iframe. Noted as a candidate for a postMessage activity bridge if it matters in practice.
- Devin transient-ping-failure reload: a ping is a host-proxied call to the host's own wire; if it fails, fn calls from the embed are failing too, so treating it as the stale-frame symptom (one covered re-open, absorbed failure) errs toward a working surface. Requiring consecutive failures would leave a dead embed up for another full interval.
- cubic P3 freePort race — standard ephemeral-port test pattern; window is microseconds and the suite binds immediately after.

PR #433, opened 2026-07-20
v3 W4b: islands — ambient scope (React + Kit + fmt, no imports) + ambient tools with compiler-stamped, host-enforced manifests

yousefh409:
v3 Wave 4b — islands: ambient scope + ambient `tools` (spec §format Islands).

## What ships

**1. Ambient scope — islands have no imports.**
React + hooks, the ENTIRE Kit (31 components incl. recharts-backed charts), and `fmt` value helpers are injected into the jail evaluation scope (react-live pattern; shadow-safe per-module param filtering). The name list is `ISLAND_AMBIENT_NAMES` in core, pinned to the ui Kit registry by test. Known specifiers (`react*`, kit-ish names) are **silently stripped** at compile (pretraining habit); unknown specifiers stay compile errors → repair. The Kit is bundled ONCE into the jail runtime via the existing `build-jail-runtime.mjs` pipeline (bundle 348KB → 906KB; it also resolves habit imports for streaming partials — same objects, never a second copy).

**2. Ambient `tools` — literal member access only, compiler-stamped manifest, host-enforced.**
`await tools.<tool_name>(args)` inside islands. The compiler scans literal member chains (string/comment/JSX-text safe), rejects computed access (`tools[expr]`) and aliasing, resolves every chain against the live registry (unknown → repair), and stamps the per-island manifest into the app document (`componentTools`, validated by `validateAppDocument`, served with payloads, restamped on every edit). The runtime side: the jail posts `tool-call` over the existing postMessage seam; **JailedComponent (host side) exposes ONLY the manifest** — never trusting the iframe — and routes allowed calls through the EXACT same guarded pipe as tree actions. Unstamped/streaming payloads fall back to a host-derived source scan. The legacy `action` channel is now also least-privilege (prop-embedded `$action` names ∪ manifest ∪ literal `vendo.action` names). Empty manifest = island can call nothing.

**3. Fear rules retired.**
"LAST RESORT / never data in islands" is gone; the posture is the spec's: Kit when it covers the need, island for custom visuals/logic/interaction. Kept: byte caps, TSX + default-export gates, no-network CSP, never-the-whole-app as streaming-economics guidance.

**Live-run hardening (found by the browser runs, fixed + unit-tested):**
- HOST TOOLS prompt lines now carry an `(input: {…})` sketch (one nesting level). Without it the live P3 island called a body-nested tool with flat args → the host route read an empty JSON body → **the approved mutation ran on defaults**. Reproduced, fixed, re-proven.
- `fetch`/XHR/WebSocket in island source is a compile error → repair (live P3 variant wrote a habit `fetch("/api/…")`; the CSP blocked it silently).
- Host catalog / prewired component tags inside island JSX are a compile error → repair (live P4 rendered `<MapleSpendingDonut/>` → jail ReferenceError).

## Adversarial coverage
- `tools[expr]` and `tools["name"]` computed access → compile error → repair (engine test).
- Aliasing (`const t = tools`, passing `tools` around) → compile error → repair.
- Runtime: a forged `tool-call` naming a tool outside the stamped manifest is **blocked host-side** before the pipe; malformed paths dropped; stamped-era island with no manifest entry = zero tools; forged legacy `action` names blocked (ui tests).
- Import gate: unknown modules still unresolvable in the jail (`node:fs` test), blob/network walls unchanged.

## Live end-to-end proof (prod boot, fresh prompts — never the frozen 30)
`docs/verification/w4-islands/` — harness `verify-live.mjs` + screenshots, demo-bank `next start` on :3000, live sonnet:

- **P1 derivation island** (`p1-derivation-island.png`): what-if savings calculator — no imports, Kit + fmt ambient, typed $500/mo → correct per-goal months (ceil(1880/500)=4mo) off live goal data.
- **P2 search-as-you-type read island** (`p2-search-island.png`): transaction search calling `tools.host_listTransactions({search})` per keystroke — "tartine" filters to live Tartine rows, formatted amounts. Stamped manifest: `["host_listTransactions"]`.
- **P3 mutating island → approval → approve → EFFECT LANDS** (`p3-approval-pending.png`, `p3-approved-effect.png`): island Pay button → ambient tools mutation with **body-nested args** (`{"body":{"amountCents":4137,…}}` visible on the approval card, "asked in an app") → destructive tool parks at the approval gate → Approve → W0 approve→resume re-dispatches the byte-exact call → **transaction POSTED: merchant "Jordan Avery", amount −4137 cents (= the typed $41.37), memo token intact**, verified via the host API.
- **P4 interactive Kit-chart island** (`p4-chart-island.png`): category comparison with two Kit Selects — changing a picker re-computes bars + delta from live spending data.

Persisted-document proof (PGlite dump): every generated island source has **zero import statements**; `componentTools` stamped per island — `QuickPay → ["host_createOrder"]`, search → `["host_listTransactions"]`, calculator → `[]` (least privilege).

## Honest findings (not fixed here)
- Catalog-collision repair loop: a "spending mix donut + toggles" ask conflicts with the catalog's "MapleSpendingDonut is THE spending viz" steering; the model repeatedly re-emits the host component inside the island and the repair loop is slow (>10min) and can exhaust. The new compile gate + prompt line reduce it; the pipeline/W5 lanes should measure it.
- Generated islands sometimes render optimistic "sent" states for pending-approval outcomes, and one search island showed `+$` for debits — model output quality, contained, not infra.
- Jail runtime bundle grew to ~906KB (recharts) inlined per island iframe srcdoc; fine for now, a follow-up could lazy-split charts.

Gates: `pnpm build && pnpm test && pnpm typecheck && pnpm lint` green (test at `--concurrency=4`; `box-harness` Procfile test is flaky under full parallel load — passes standalone, pre-existing). Rebased onto main past W3 #425 (kept W3's `assembleTree`, extended it with `componentTools`).

---

<a href="https://app.devin.ai/review/runvendo/vendo/pull/433" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---
## Summary by cubic
Ships ambient islands: React + full Kit + `fmt` are injected (no imports), and `tools` calls are allowed only via a compiler-stamped, host-enforced manifest. Hardened scanning (called-use only with `?.`/`?.[` parity) and strict host checks make islands safer and simpler.

- New Features
  - Single-source ambient contract in `@vendoai/core/island-ambient`: tuple-pinned ambient names, known-import stripping (non-code safe), literal `tools` scan/resolve, `vendo.action` name scan, and network gate. UI test pins the Kit registry to the ambient list to prevent drift.
  - Payload wiring: `@vendoai/apps` stamps `componentTools` on create/edit; `@vendoai/ui` passes it with payloads and `JailedComponent` exposes only stamped tools. Legacy/streaming falls back to a host-derived manifest; manifests survive fork-only edits.
  - Jail runtime: injects React + hooks, full Kit (incl. charts), and `fmt` into scope; allows local component shadowing; blocks malformed `tool-call`s host-side. Added live verification harness and screenshots in `docs/verification/w4-islands`.

- Migration
  - Write islands with no imports; use Kit/`fmt` directly; call `tools.<name>` via literal member access only (supports optional chaining; no aliasing or `tools[expr]`).
  - Network calls and host components inside island JSX are compile errors.
  - Island names must not conflict with Kit/catalog/prewired components; built-ins win.
  - Existing apps continue; if `componentTools` is absent, the host derives a manifest; an empty manifest means the island cannot call tools.

<sup>Written for commit 27124078ac53d75d46dd875ae519093011bddfe0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/433?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Review triage (2bc9daba):

**Fixed:**
- `ISLAND_AMBIENT_NAMES` was annotated `readonly string[]`, which collapsed the jail-scope `Record` pin to `Record<string,…>` — now a literal `as const` tuple + exported `IslandAmbientName`, so name-list drift is a compile error again (cubic P3 ×2, Devin).
- Import strip and the legacy `vendo.action` literal scan now locate matches on a non-code-masked view (offset-preserving) — import-like text or action names inside strings/comments can no longer corrupt source or widen the action gate (cubic P1/P2, Devin).
- `tools` scan now collects only CALLED chains — JSX prose like "great tools.Buy now" no longer produces spurious unknown-tool repairs; un-called chains in expression position stay aliasing violations (cubic P3, Devin).
- `tools?.["x"]` optional-chained computed access now rejected like `tools[expr]` (cubic P2).
- Malformed `tool-call` messages now get a `blocked` tool-result instead of leaving the island promise hanging (Devin).
- A locally-declared component that shadows a host catalog/prewired name is the island's own — no longer rejected (cubic P3).
- Fork-only edits keep `componentTools` DEFINED whenever components exist, so pinned islands stay on the stamped-era zero-tools default, never the source-scan fallback (cubic P1).
- Harness: per-run memo in the P3 effect predicate, unknown scenario names exit 2, P4 picks an option different from the current one (cubic).

**Accepted / by design:**
- Legacy `action` channel becoming least-privilege is intentional (W4b §2); the guard+approval pipe still gated everything before, this narrows the reachable name set. Dynamic action names must be prop-embedded — which is how trees wire them.
- Regex literals aren't masked by `blankNonCode` (needs a real lexer); worst case is a spurious repair round or an extra manifest entry that the guard still gates.
- `stripIslandImports` flagging non-ambient locals (e.g. `React2`) routes to repair on purpose — silently keeping the import would break in the jail anyway.

PR #510, opened 2026-07-22
cli: Wave 2 install-funnel fixes — visible surface, principal alignment, doctor honesty, login write-preflight (0.4.x E2E cert)

yousefh409:
Wave 2 of the 0.4.x E2E certification follow-ups: the install-funnel fixes for B3 (invisible surface), B4 (principal mismatch / infinite skeleton), M3 (doctor render honesty + E-WIRE-004), M4 (login write-preflight), and D8 (`--url` copy). Full repros live in the campaign reports (`~/vendo-test-repos/.e2e-harness/report-{quickstart,invoify,mastra}.md`, master matrix `E2E-CERT-0.4.x.md`).

## 1. Visible surface — init auto-wires `<VendoRoot>` + `<VendoOverlay />` (B3, launch blocker)

Two certified stacks ended a by-the-book install doctor-green with **nothing on screen**: `<VendoRoot>` is a context provider, and neither init's Last steps nor the playbook ever mentioned `VendoOverlay`. Worse, the printed paste taught the registry-in-server-layout pattern that RSC-crashes every page once the registry holds real components (invoify defect 4).

- Init now generates **`vendo/vendo-root.tsx`** next to the registry: a `"use client"` wrapper that owns the registry + theme imports (the RSC-safe home for them) and mounts `<VendoOverlay />` inside the provider. Generated only while absent, never clobbered.
- Init makes **one bounded, idempotent edit** to `app/layout.tsx`: one import line + wrapping the single JSX `{children}` in `<VendoRoot>` (the wrapper's export keeps the name, so doctor's grep and the docs stay coherent). Skipped when any Vendo mount already exists; a layout without exactly one unambiguous JSX `{children}` is left untouched and the exact paste lines are printed instead (`wireNextLayout` returns null — fallback preserved).
- A layout that already mounts `<VendoRoot>` but no surface (the quickstart D6 case) gets one overlay paste line; BYO hosts with their own surface (embeds/hooks/VendoProvider) are left entirely alone.
- `VendoOverlay`/`VendoOverlayProps` are re-exported from `@vendoai/vendo/react` so the scaffold resolves under pnpm strict linking (same TS2307 story as the registry's `ComponentRegistry` import).
- Last steps, agent tail, `--agent` plan (layout edit rides `codeChanges` as a diff), Express client hints, and the docs (quickstart, vendo-init, cli reference, verify) all updated accordingly.

## 2. Principal alignment + `open?pending=1` honesty (B4, blocker)

The scaffolded wire route used `principal: () => null` while both existing-agents quickstarts document a `demo-user` chat principal — every app/approval created through the host's agent loop was invisible to the embeds, and the embed's `open?pending=1` poll masked the owner-scoped not-found as `{kind:"pending"}` forever (the infinite skeleton).

- The anonymous scaffold now resolves `{ kind: "user", subject: "demo-user" }` with a comment stating the invariant: the wire route must resolve the SAME subject the agent loop uses; replace both sides with a real session lookup. Docs' shared-module snippets (ai-sdk + mastra) gained the matching principal line, and `--auth none` docs updated.
- The wire's `?pending=1` arm now distinguishes the two not-founds: a record that **exists under another principal** answers terminal `{kind:"failed", retryable:false}` with the principal-mismatch diagnosis (embeds resolve promptly with an actionable reason); a genuinely absent record (the build window) keeps `{kind:"pending"}`. Unflagged callers keep the contracted 404. The existence probe is an unscoped id-only check (`appStore(deps.store).get`) that leaks no document content; hosted wire-door stores that can't answer degrade to today's pending window.

## 3. Doctor honesty (M3 + minors)

- **New E-WIRE-006 (fails)**: no visible agent surface mounted anywhere — `<VendoOverlay>`, `<VendoThread>`, the BYO embeds, or a surface hook. Green now means a user can actually reach the agent.
- **New E-LIVE-006 render gate**: after `/status` succeeds, one GET of the app origin root; a 5xx fails doctor (the certified invoify install had every page 500ing while doctor exited 0).
- **New E-DEP-002 version-skew check** (coordinator scope addendum from the live 0.4.2 re-run): doctor compares its own version against the wire `/status` version and fails on mismatch with the exact fix (`npm install @vendoai/vendo@<cli-version>`, or drop the direct dep, then restart the dev server). Catches the split-brain install where a direct `@vendoai/vendo@^0.4.1` pin beats the `vendoai` umbrella and `/status` silently serves the old runtime — and stale dev servers after any upgrade.
- **E-WIRE-004 broadened**: accepts a `<VendoRoot>` (or `<VendoProvider>`) mount in ANY `layout.{tsx,jsx,js}` under `app/`/`src/app/` — nested i18n/route-group layouts (invoify mounted in `app/[locale]/layout.tsx`) no longer fight the check — and the message points at the wrapper pattern instead of the RSC-crashing paste. verify.mdx fix sections rewritten (the old E-WIRE-004 fix snippet taught the crashing pattern).
- **Exit codes (3b)**: `runDoctor` already returned 1 on any broken check at HEAD, and I could not reproduce broken+exit-0 from the shipped 0.4.2 code — the mastra observation is most plausibly `$?` read through a pipe (I reproduced exactly that artifact with `doctor | tail`). Hardened anyway: regression tests pin broken-live-turn → exit 1, and the probe dev-server's `stop()` now destroys the piped stdout/stderr and unrefs the child, so surviving `npm run dev` grandchildren can no longer hold doctor's event loop open past its verdict.
- **`--url` copy (D8)**: the unreachable-`/status` failure now says doctor expects the WIRE BASE (`http://localhost:3000/api/vendo`), not a bare site origin.

## 4. Login write-preflight (M4)

`vendo login` redeemed the single-use claim before ever proving `.env.local` was writable — under headless Claude Code env-file protection the key minted server-side and was lost.

- A real append-open probe of `.env.local` now runs **before any network call** (permission checks can't see sandbox policies that deny at open time; a probe-created empty file is removed). Failure exits 1 with copy that names the file, says the ceremony was NOT started and no key was minted, and explicitly distinguishes itself from a timeout.
- A redemption-time write failure (the race preflight can't fully close) now reads as a **write** failure — key minted but NOT saved, revoke in the console Keys page, re-run after fixing access — never the generic polling/timeout copy. The consumed claim's pending file is cleared so the next login opens a fresh claim instead of resuming into `invalid_grant`.

## Tests

- `wireNextLayout` unit tests (single-wrap, multi-line imports, directive handling, whitespace preservation, ambiguity/idempotence nulls) + init integration tests for the wired path, the uneditable-layout fallback, overlay-missing, and BYO leave-alone.
- Wire test: foreign-owned record → `{kind:"failed"}` under `?pending=1`, 404 unflagged; absent record stays pending.
- Doctor tests: E-WIRE-006 fail + BYO-embed pass, nested-layout E-WIRE-004, E-LIVE-006 render gate, broken-turn → exit 1, `--url` copy.
- Device-login tests: preflight blocks pre-network with zero requests, no probe artifact, redemption write-failure copy + pending-claim cleanup + key never printed.

Changeset: patch on `@vendoai/vendo`.

## Known limits / notes for the re-run wave

- A truly nonexistent appId under `?pending=1` still answers pending until the embed's 5-minute deadline (the server cannot distinguish "never existed" from "build in flight" — the record lands at build completion). The B4 skeleton case (exists-under-other-principal) resolves promptly now.
- The `demo-user` default principal means a no-auth production install shares one subject across visitors — the scaffold comment and host-auth docs say to replace it; auth presets replace it wholesale. This trade was the approved W2 direction (align with what the docs' chat routes teach) — flagging for visibility.
- E-WIRE-006's marker list is deliberately generous (chrome components, embeds, hooks); a host with a fully bespoke surface built on `useVendoThread`/`useVendoOverlay` passes.

---

<a href="https://app.devin.ai/review/runvendo/vendo/pull/510" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---
## Summary by cubic
Auto-wires a visible agent surface during `vendo init`, aligns the default principal, strengthens `vendo doctor`, and adds a safe write-preflight to `vendo login`. Fixes invisible UI after install, endless pending states, and lost API keys during device login.

- **New Features**
  - `vendo init`: generates `vendo/vendo-root.tsx` ("use client") that mounts `<VendoOverlay />` and imports the registry/theme, and makes one idempotent edit to `app/layout.tsx` to wrap `{children}` with `<VendoRoot>`; falls back to printed paste lines if the layout is ambiguous.
  - `@vendoai/vendo/react`: re-exports `VendoOverlay` and its types so scaffolds resolve under pnpm strict linking.
  - `vendo doctor`: adds checks for a visible surface (fails if none), a root-page render gate (5xx fails), and a CLI-vs-wire version skew check; accepts nested layouts.

- **Bug Fixes**
  - Wire principal now defaults to `{ kind: "user", subject: "demo-user" }` to match quickstarts; `open?pending=1` returns `{ kind: "failed", retryable: false }` when a record exists under another principal (no more infinite skeleton).
  - `vendo login`: preflights `.env.local` writability before any network call; redemption write failures report clearly (revoke and retry) and clear the pending claim.
  - Doctor: ignores the generated `vendo/vendo-root.tsx` in surface detection so an unmounted wrapper no longer passes E-WIRE-006; hardened exit code and no-hang shutdown; clearer `--url` guidance (expects the wire base, e.g., `http://localhost:3000/api/vendo`).

<sup>Written for commit 1e2babba8c6d5431bded7114e022563ccfacc1b0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/510?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
**End-to-end smoke of the B3 fix (real CLI, real Next app):** copied the campaign's pristine `quickstart-next` (create-next-app, Next 16.2.10) into a scratch dir, overlaid this branch's built `@vendoai/vendo` dist into its `node_modules`, restored the stock `app/layout.tsx`, and ran `vendo init --yes --byo`:

- init wired `+ vendo/vendo-root.tsx`, `+ app/api/vendo/[...vendo]/route.ts`, `~ app/layout.tsx` (import + `<body><VendoRoot>{children}</VendoRoot></body>`), printed the "Mounted <VendoRoot> + <VendoOverlay />" line, and the agent tail's `surface:` fact — no manual paste left.
- `next build` (webpack) exits 0 and **statically prerenders `/` through the wrapper** — the exact spot the old registry-in-server-layout paste crashed with the RSC serialization error.
- The prerendered `.next/server/app/index.html` contains the **`AI agent` launcher pill** markup alongside the page content — the agent is on screen after a by-the-book install, which is the whole point of B3.

(Turbopack builds in the scratch setup fail only on its symlinked-node_modules project-root inference — a harness artifact, not this change; the campaign repos with real installs build turbopack fine.)

Re-run of the flaky gate pieces: `@vendoai-corpus/express-host` and `demo-bank` test timeouts from the parallel turbo run both pass in isolation; full `@vendoai/vendo` suite green (4 shards), `pnpm typecheck` + `pnpm lint` green, all previously turbo-cancelled packages re-run green.

yousefh409:
Review pass done (coordinator). Devin's wrapper finding was real and is fixed in the follow-up commit: doctor now mirrors init's layout decision — the generated wrapper is excluded from the scan, and a user-code `<VendoRoot>` mount next to an overlay-bearing wrapper satisfies the surface; two regression tests added (unmounted wrapper → E-WIRE-006, auto-wired shape → pass). Devin's other notes triaged: the pending-probe build-window nuance and E-DEP-002's strict equality are intended behavior; the render gate's swallow-on-unreachable-origin is acceptable as belt-and-braces atop the static surface gate (the static gate is the load-bearing check).

Issue #521, opened 2026-07-22
Persona layer: a per-user model so the agent can resolve habit-driven requests


anshulforyou:
**Problem**

Vendo reshapes per prompt, but it has no model of who the user is across sessions. Every turn starts cold. I went looking for where user identity feeds the agent and found that it does not: `ctx.principal.subject` threads through every table and tool call, but only for isolation and authorization. `assembleSystemPrompt` is identical for every user. UI generation is deliberately user-blind. There is no persona, profile, or preferences abstraction anywhere, and the v0 contracts deferred `memory` on purpose.

The cost shows up on under-specified requests. When a user says "pull up my usual view" or "do my routine thing", the agent has nothing to resolve it against, so it guesses. A finance clerk who lives in invoices and a treasurer who lives in transfers send the same words and should get different tools, and today they cannot.

**Proposed solution**

Add a per-user persona: a small, accumulated model of how a person works, fed back into the agent. I built it to be purely additive, living entirely in the umbrella, touching no frozen block:

- A per-subject persona record (how the user works, the tools they reach for, the formats they prefer, durable facts they state) stored as opaque data in a `vendo_records` collection keyed by subject. No schema change. It rides the host's own database like everything else Vendo stores.
- Two guard-bound, subject-scoped agent tools, `vendo_persona_load` and `vendo_persona_remember`, folded into the same registry the app tools use. They inherit the guard binding, so persona reads and writes are policed and audited like every other tool, with no separate path.
- `distillPersona`, which builds a persona from a subject's own threads and audit trail, deterministically by default.

Generation stays user-blind: the persona reaches the agent prompt only, never a generated document, and subject isolation is never crossed.

I proved it with offline replay: hold out a user's decisions, and measure whether a persona-conditioned agent reproduces them versus the stock agent, one model held constant across both arms. On requests the prompt already determines, persona neither helps nor hurts (both 87.5 percent), which is the control I cared about most. On persona-dependent requests (the elliptical ones above), it went from 25 percent to 100 percent. The fixtures are synthetic and built to isolate the persona-dependent case, so this shows the mechanism, not a real-traffic frequency. The harness takes an injected model turn, so the same replay reruns against real threads on a deployment to get a representative number.

[...]

PR #530, opened 2026-07-23
feat(vendo): per-user persona layer for the agent

anshulforyou:
Draft, tied to the design discussion in #521. Opening it so the code is visible and easy to react to. Happy to iterate on the shape or hold it entirely per your steer.

## What

A per-user persona layer for the agent, additive and living entirely in the umbrella so no frozen block changes. A per-subject persona record (how the user works, the tools they reach for, the formats they prefer, durable facts they state) in a `vendo_records` collection keyed by subject, with no schema change. Two guard-bound, subject-scoped tools, `vendo_persona_load` and `vendo_persona_remember`, folded into the same registry the app tools use, so persona reads and writes are policed and audited like every other tool. `distillPersona` builds a persona from a subject's own threads and audit trail, deterministically by default. An offline replay harness plus fixtures measure whether a persona-conditioned agent reproduces a user's held-out decisions.

## Why

Vendo has no model of who the user is across sessions, so the agent starts every turn cold and cannot resolve habit-driven requests: "pull up my usual view" should mean different things to a user who lives in invoices and one who lives in transfers, and today it cannot. This adds that model in the smallest additive way. The persona reaches the agent prompt only, generation stays user-blind, and subject isolation is never crossed. The always-on `assembleSystemPrompt` variant is deliberately left out as a follow-up, since it would touch the frozen agent contract. Full rationale and alternatives are in #521.

## Verification

- [x] `pnpm build`, `pnpm typecheck`, `pnpm lint` green, dependency-guard passes.
- [x] New persona unit and eval tests pass (18 of 18). The full `@vendoai/vendo` suite shows only failures that pre-date this change and are unrelated to it, confirmed by running the suite on a clean baseline with this diff removed. They trace to the local Node version, not this code, and should clear on CI.
- No UI-affecting changes, so no screenshots.

---
## Summary by cubic
Add a per-user persona layer so the agent remembers how each user works across sessions. Persona writes are now concurrency-safe with atomic CAS to prevent lost updates.

- **New Features**
  - Per-subject persona stored in `vendo_records` (collection `persona`, id = subject); bounded facts with tolerant reads.
  - Guard-bound tools added to the action registry: `vendo_persona_load` (risk: read) and `vendo_persona_remember` (risk: write), audited like app tools.
  - `distillPersona` builds/refines a persona from the user’s own `vendo_threads` and `vendo_audit`; deterministic by default with an optional summarizer hook.
  - Offline replay eval (`runPersonaReplay`) plus grounded fixtures to measure how persona-conditioned turns reproduce held-out tool choices.
  - Server wires in `createPersonaTools`, keeping subject isolation; persona only influences the system prompt, never generated documents.

- **Bug Fixes**
  - Persona writes now use the store’s atomic compare-and-swap with bounded retries; fails closed when the adapter omits `atomic`.

<sup>Written for commit 9052f9bc81be57f449ab32ddb7c86c3e6db1eb81. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/530?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/persona-layer.md
+++ b/.changeset/persona-layer.md
@@ -1,0 +1,12 @@
+---
+"@vendoai/vendo": minor
+---
+
+Add a per-user persona layer to the agent.
+
+- A per-subject persona record (how the user works, the tools they reach for, the formats they prefer, durable facts they state) stored as opaque data in a vendo_records collection keyed by subject, so it rides the host's own database with no schema change.
+- Two guard-bound, subject-scoped agent tools, vendo_persona_load and vendo_persona_remember, folded into the same registry the app tools use, so persona reads and writes are policed and audited like every other tool.
+- distillPersona, which builds or refreshes a subject's persona from their own threads and audit trail, deterministically by default.
+- An offline replay harness that measures whether a persona-conditioned turn reproduces a user's held-out decisions.
+
+Generation stays user-blind: the persona reaches the agent prompt only, never a generated document, and no subject boundary is crossed.
--- a/packages/vendo/src/persona-eval.test.ts
+++ b/packages/vendo/src/persona-eval.test.ts
@@ -1,0 +1,90 @@
+import { memoryStoreAdapter } from "@vendoai/core/conformance";
+import { describe, expect, it } from "vitest";
+import { distillPersona } from "./persona/index.js";
+import { runPersonaReplay, seedSubject, type RunTurn, type SubjectFixture } from "./persona/eval.js";
+import { GROUNDED_FIXTURES } from "./persona/eval-fixtures.js";
+
+const repeat = (value: string, times: number): string[] => Array.from({ length: times }, () => value);
+
+// Synthetic labeled subjects: each consistently reaches for one tool for a kind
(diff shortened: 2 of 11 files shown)


yousefh409:
Thanks for this! Will take a look tomorrow and get back to you ASAP

yousefh409:
Am going deep into the code right now; but love the idea and this has been on our roadmap. Will keep you updated and great work!

anshulforyou:
> Am going deep into the code right now; but love the idea and this has been on our roadmap. Will keep you updated and great work!

Awesome, glad it lands. I will keep the persona side moving on my end, so shout if you want me to shape it to whatever you have in flight. Excited to see where you take it.

PR #546, opened 2026-07-25
feat(store): knowledge record collections + SCHEMA_VERSION 4->5 (ENG-356)

YSharara1:
Closes ENG-356 (Stage 0 — Foundations, @vendoai/knowledge). Builds on Amr's contract freeze (#543) and wire protocol (#545).

## What

Adds the store-side home for the knowledge corpus, in `@vendoai/store`:

- **Two dedicated record collections** in `schema.ts` `DDL` — `vendo_knowledge_docs` (one row per document-level corpus entry) and `vendo_knowledge_chunks` (one row per engine-minted chunk of a synced doc; the local engine's index — the cloud engine keeps its corpus server-side and never populates these). Same `id / data / refs / created_at / updated_at` layout as the MCP door tables, each with a `refs` GIN index (`jsonb_path_ops`) and a newest-first keyset index in `ADDITIVE_DDL` (F2's 1000-row page bound).
- **`SCHEMA_VERSION` 4 → 5.** This is load-bearing, not cosmetic — the fix the ticket calls F1: the `DDL` loop runs only while `version < SCHEMA_VERSION`, so appending the tables **without** the bump would leave every existing v4 database on 4 forever and the new tables would never be created.
- **Erase cascade.** Both tables join `ERASE_TABLES` and the `bySubject` / `byApp` cascades via `refs @> selector`, exactly like the door tables (the knowledge engine owns what it refs). The existing `erase.conformance.test.ts` invariant — `ERASE_TABLES` must equal the tables `DDL` creates (+ `vendo_meta`) — keeps the two in lockstep, so a missed cascade entry fails the suite.

Docs/tests updated in step: `schema.test.ts` (contract columns, jsonb columns, GIN-index loop, version + 14→16 table count), `state-routing.test.ts` version assertion, `erase.conformance.test.ts` report + comment, `packages/store/README.md` (14→16 tables).

## Verification

**CI is authoritative for this PR — I could not run the gates locally.** The dev box's git transport resets on any large pack, so the full workspace tree / `node_modules` could not be materialized (I worked from a partial + sparse checkout of `packages/store`). This is the same "CI is authoritative" posture as #543/#545 (there it was the broken Windows turbo binary; here it's the transport). The change is deliberately narrow and self-checking:

- The DDL mirrors the `vendo_mcp_clients` / `vendo_mcp_grants` precedent statement-for-statement.
- `schema.test.ts` asserts the tables, key columns, jsonb storage, and GIN indexes exist; the version-migration tests (v1→current, v2→current) assert the 4→5 landing.
- The `ERASE_TABLES` ⇔ `DDL` invariant test guarantees cascade completeness.

Please confirm the store package's test job (both pglite + postgres backends) is green before merge.

Changeset: minor bump of `@vendoai/store` (fixed lockstep group).

https://claude.ai/code/session_01Qx6i5Q7h2qpvGpRcPXp91n

---

<a href="https://app.devin.ai/review/runvendo/vendo/pull/546" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/.changeset/knowledge-store-ddl.md
+++ b/.changeset/knowledge-store-ddl.md
@@ -1,0 +1,5 @@
+---
+"@vendoai/store": minor
+---
+
+Add the dedicated knowledge record collections `vendo_knowledge_docs` / `vendo_knowledge_chunks` (MCP-door table layout: id/data/refs/created_at/updated_at, GIN index on refs, newest-first keyset index) and bump `SCHEMA_VERSION` 4→5 so existing databases actually create them (the DDL loop only runs while `version < SCHEMA_VERSION` — review fix F1). Both tables join the erase-by-subject/app cascade.
--- a/packages/store/README.md
+++ b/packages/store/README.md
@@ -78,4 +78,4 @@ Because the registry is a table in the same database, sessions survive restarts 
 
 ## Retention and erasure
 
-`eraseStore(store)` is the store-level erase API — `bySubject(subject)` for full erasure and `byApp(appId)` — cascading the matching rows across all 14 tables (ephemeral subjects included — their rows are ordinary disk rows) and returning per-table deleted counts. It is the only sanctioned deletion path for `vendo_audit` rows. It is also re-exported from `@vendoai/vendo/server`. Host SQL remains available for everything else.
+`eraseStore(store)` is the store-level erase API — `bySubject(subject)` for full erasure and `byApp(appId)` — cascading the matching rows across all 16 tables (ephemeral subjects included — their rows are ordinary disk rows) and returning per-table deleted counts. It is the only sanctioned deletion path for `vendo_audit` rows. It is also re-exported from `@vendoai/vendo/server`. Host SQL remains available for everything else.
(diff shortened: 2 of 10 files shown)


YSharara1:
Thanks — all three review findings were legitimate and are fixed in `2122c71a`:

- **Greptile P2 (routing):** ✅ Fixed. Registered `vendo_knowledge_docs` / `vendo_knowledge_chunks` in `DEDICATED_RECORD_COLLECTIONS` (and the `DedicatedRecordTable` type) so `store.records(...)` routes them to their dedicated tables instead of the generic `vendo_records`. This was the right catch — "add dedicated record collections" in ENG-356 means routing, not just DDL; I'd only done the tables.
- **Greptile P2 (erase unexercised):** ✅ Fixed. The conformance test now seeds a subject-ref'd doc+chunk (bySubject) and an app-ref'd doc (byApp) and asserts they're erased (count 1) with the sibling spared — so a regression in the new cascade DELETEs now fails the suite.
- **Devin 🟡 (README count vs listing):** ✅ Fixed. Added both tables to the store README table map and the dedicated-collections section, and corrected the "two door-owned collections" phrasing.

Also extended the routing round-trip/refs/claim test loop to cover the knowledge tables. No false positives to reject.

PR #564, opened 2026-07-25
fix(apps): smoke-render gate — an unloadable module is an environment skip, not an island crash

yousefh409:
Found by the **2026-07-25 nightly demo QA** run against `eee10d2e`. The flagship
"ask Maple where my money went → get a live spending view" demo was broken: the
user got a red **"Couldn't build the app — app build failed: generation
failed"**.

## What was happening

The new smoke-render gate (#498) renders each island headless in a worker, with
module paths resolved on the main thread. Under Turbopack,
`require.resolve("jsdom")` **does not throw** — it *succeeds*, handing back a
synthetic specifier:

```
[externals]/jsdom [external] (jsdom, cjs, [project]/node_modules/.pnpm/jsdom@25.0.1/node_modules/jsdom)
```

Two things then go wrong:

1. **The escape hatch never fires.** The file's own contract promises
   *"Environment failures (react/jsdom unresolvable, worker 'ready' never
   arrives) skip the gate silently … so a bundler that cannot reach the modules
   degrades to today's behavior instead of failing creates."* That path is only
   reached when `.resolve()` **throws**.
2. **The failure is blamed on the island.** The worker's
   `require(workerData.paths.jsdom)` fails in the prelude; the IIFE's catch-all
   posts it as a per-island *result*, so the user is told:

   > island "LateNightTransactions" crashed when rendered against stubbed tool
   > results: Cannot find module '[externals]/jsdom [external] (…)' — the
   > component must render without throwing … guard undefined/empty results
   > before .map/.reduce

   …and it is routed to repair, which cannot fix a module that will not load.

**Why it looked intermittent.** Three generations on the unpatched server: 2
failed, 1 shipped. The one that shipped had **0 islands** — nothing for the gate
to render. Both failures carried islands, and *every* island failed with the
identical jsdom error (one hit both `RecentTxList` and `SummaryBar`). The rule
is not "sometimes": **any app with an island cannot ship.**

**Blast radius.** Published `0.4.8` is *not* affected — it has no smoke-render
gate and no `jsdom` dependency. But `packages/apps/package.json` now declares
`jsdom` in `dependencies`, so on the next release every Next+Turbopack host
installs it and takes this same path.

## The fix

- **Main thread** — a resolver only wins when every resolved path is a real file
  (`existsSync`). A synthetic specifier falls through to the next resolver and,
  failing that, skips the gate as designed.
- **Worker** — track whether `ready` was posted and report a pre-`ready` failure
  as `{type:"environment"}`; the main thread resolves that to *no issues*, the
  same skip an unresolvable module already produced. The `error`/`exit` handlers
  already drew this before/after-`ready` line — the IIFE catch-all did not.

## Scope, stated plainly

This restores the **documented degradation** (skip) under Turbopack. It does
**not** make the gate execute there — I verified from inside the running dev
server that Turbopack's `createRequire` returns non-existent synthetic paths for
`react` and throws for `jsdom` on both resolver rungs, so the gate correctly
skips rather than runs. Making it actually run under Turbopack is a separate
design call for whoever owns the gate; this PR stops it from failing every
island-bearing create.

## Verification

- `pnpm build` — **21/21 green**
- `pnpm typecheck` — **37/37 green**
- `pnpm lint` — **4/4 green**
- `pnpm --filter @vendoai/apps test` — **603 passed / 18 skipped** (was 602;
  +1 is the new regression test)
- The new test **fails without the worker-side change** (verified by reverting
  just that hunk: `1 failed | 602 passed`), so it is not vacuous.
- **In the real demo**, patched `demo-bank` on Turbopack: 5 generations, **zero**
  jsdom misattributions and zero island-blamed failures (the single failure was
  an unrelated genuine issue, *"tree root renders an empty layout"*).

**Honest gap:** I could not get the generator to emit an island on the patched
server across 5 attempts (all tree-only, 0 island nodes), so the island path is
proven by the regression test and the in-server resolver probe rather than by an
end-to-end browser capture of an island-bearing app.

### Pre-existing, not from this PR

Full `pnpm test` is **already red on `main`**: `@vendoai/ui` fails 5 files / 28
tests (`chrome/discoverability-*`, `lane-thread-picks`). Identical failures on a
clean checkout of `eee10d2e`. Untouched here, but worth a look given the repo's
"must be green before any PR" rule.

---
## Summary by cubic
Fixes the smoke-render gate to treat unloadable modules as environment failures and skip, instead of reporting island crashes. This prevents Turbopack builds from failing when `require.resolve("jsdom")` returns a synthetic external path.

- **Bug Fixes**
  - Main thread now accepts a resolver only if all resolved paths are real files (`existsSync`); synthetic externals fall through and, if unresolved, skip the gate.
  - Worker tracks a pre-`ready` state and posts `{ type: "environment" }` for failures before readiness; the main thread treats this as a skip, not an island error.
  - Exported the worker source and added a regression test to verify that an unloadable `jsdom` path reports an environment failure instead of an island crash.
  - Restores the documented skip behavior under Turbopack; the gate still does not execute there.

<sup>Written for commit 93a97f7bc5a0efe7d721ec82aeb265951ba24989. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/564?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/smoke-render-environment-skip.md
+++ b/.changeset/smoke-render-environment-skip.md
@@ -1,0 +1,5 @@
+---
+"@vendoai/apps": patch
+---
+
+Smoke-render gate: treat a module the worker cannot load as an environment failure (skip) instead of an island crash. Under Turbopack `require.resolve("jsdom")` does not throw — it succeeds with a synthetic `[externals]/jsdom [external] (…)` specifier — so the gate's documented "unresolvable → skip silently" path never fired, and the worker's prelude `require` failure was reported as a per-island render crash and routed to repair, which cannot fix a module that will not load. Every generated app carrying an island failed to build. Resolved paths are now checked against the filesystem, and any failure before the worker posts `ready` comes back as an environment failure.
--- a/packages/apps/src/generation/validation/smoke-render.ts
+++ b/packages/apps/src/generation/validation/smoke-render.ts
@@ -24,10 +24,14 @@
  * The worker gives two guarantees an in-process render cannot:
  *   - a hard per-island timeout (worker.terminate preempts infinite loops),
  *   - no globals leaked into the host server process.
- * Environment failures (react/jsdom unresolvable, worker "ready" never
- * arrives) skip the gate silently — the esbuild lazy-load precedent — so a
- * bundler that cannot reach the modules degrades to today's behavior instead
- * of failing creates.
+ * Environment failures (react/jsdom unresolvable or unloadable, worker "ready"
+ * never arrives) skip the gate silently — the esbuild lazy-load precedent — so
+ * a bundler that cannot reach the modules degrades to today's behavior instead
+ * of failing creates. "Unresolvable" includes a resolver that SUCCEEDS with a
+ * synthetic specifier (Turbopack hands back `[externals]/jsdom [external] …`):
+ * the paths are checked against the filesystem, and anything the worker still
+ * cannot load before it posts "ready" comes back as an environment failure
+ * rather than an island render error.
  */
 import {
   ISLAND_AMBIENT_KIT_NAMES,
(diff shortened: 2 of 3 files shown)


yousefh409:
smoke-render and the island validation architecture were deleted in #1303; superseded.

PR #585, opened 2026-07-26
fix(sync): load dotenv model credentials

### LINKED ISSUE #567: vendo sync AI enrichment does not read model key from project .env file
Surfaced by Devin on #554. The enrichment pass resolves the model credential only from process.env (resolveDevCredential), so a dev whose ANTHROPIC_API_KEY / VENDO_API_KEY lives in a project .env / .env.local (not exported to the shell) gets structural-only sync with no enrichment and no clear signal why. Safe (keyless fallback is intended) but a DX gap — Next.js auto-loads .env, users expect the vendo CLI to as well. Fix direction: load the project's .env(.local) into the enrichment credential resolution (or print a hint when a key is present in .env but not process.env). Deferred from the cloud-supported-everything wave (#554).

CoderSufiyan:
Closes #567.

Loads AI enrichment credentials from the target project's `.env` and `.env.local` files, with nonblank process environment values taking precedence. The dotenv bridge allowlists model credentials and `VENDO_API_KEY` so unrelated application secrets and process controls are not passed to enrichment subprocesses.

Tests:
- `pnpm --filter @vendoai/vendo exec vitest run src/cli/sync.test.ts src/cli/doctor.test.ts`
- `pnpm build`
- `pnpm lint`
- `pnpm typecheck`

### DIFF
--- a/packages/vendo/src/cli/doctor.ts
+++ b/packages/vendo/src/cli/doctor.ts
@@ -22,7 +22,19 @@ import { detectFramework, detectVendoWiring } from "./framework.js";
 import { CONFIG_SURFACES, OVERRIDES_ENABLEMENT_CAVEAT } from "../config-surface.js";
 import { walk } from "./theme/walk.js";
 import { remoteUrls, sameUrl, validateRegistryServer } from "./mcp/registry.js";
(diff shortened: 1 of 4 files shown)


yousefh409:
Thanks — #1385 replaced these helpers with a unified readEnvFiles() on a different design, so this gap no longer exists in this shape. Closing as superseded.

PR #605, opened 2026-07-26
actions: feature detect modifier and satisfies APIs on old host TypeScript (#551)

### LINKED ISSUE #551: extract: taxonomy corpus repo crashes init — ts.getModifiers is not a function (host TS version incompat)
Surfaced by a corpus control-run during the cloud-supported-everything wave (reproduces on main). Static extraction calls ts.getModifiers against the host's own TypeScript, which predates that API — taxonomy's init crashes. Extraction should pin/bundle its TS or feature-detect. Also note: corpus baselines (2026-07-06) show theme.background drift (#f1f5f9 vs #ffffff) on both branches — baselines need a refresh pass, tracked for the eval lane.

devanonyme42:
## What

Static extraction now probes the TS 4.8 modifier APIs (`canHaveModifiers` and `getModifiers`) and the TS 4.9 `ts.isSatisfiesExpression` before use, falling back to the legacy node properties. The probes live in a small `ts-compat` module that mirrors the existing `decoratorsOf` pattern in `graphql.ts`. All twelve unguarded call sites across the sync extractors and the theme font scan now go through it.

## Why

Extraction loads the host repo's own TypeScript (`loadTypescript` resolves the host first), so `vendo init` crashed with `TypeError: ts.getModifiers is not a function` on hosts whose TypeScript predates 4.8, as reported for the taxonomy corpus run in #551. Feature detection keeps the compiler out of our bundles, in line with the portability gate, and matches the direction suggested in the issue.

Note: the TS 5.0 only `resolveModuleNameLiterals` compiler host hook silently does nothing on older hosts. It does not crash, so it is left untouched here.

## Verification

New tests in `ts-compat.test.ts` cover both helper paths (modern API and legacy fallback) plus an end to end case: a fabricated host with a simulated TypeScript 4.7 in `node_modules` extracts its server actions with zero warnings. Reverting the call site changes makes that test reproduce the exact #551 TypeError, so it guards the real path. Package coverage 91.85 percent with the helper at 100.

---

<a href="https://app.devin.ai/review/runvendo/vendo/pull/605" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---
## Summary by cubic
Feature-detect TypeScript 4.8 modifier APIs and 4.9 satisfies checks in static extraction, falling back to legacy node properties. Prevents `vendo init` crashes on hosts with TypeScript <4.8 (fixes #551).

- **Bug Fixes**
  - Added `ts-compat` helpers: `modifiersOf` and `isSatisfiesExpressionNode` with safe fallbacks.
  - Switched all extractor paths and the CLI theme font scan to use these helpers; exposed via `@vendoai/actions/sync`.
  - Added tests for modern and legacy paths, plus an end-to-end run against a real TS 4.7 compiler via a test-only `typescript47` alias.

<sup>Written for commit bc8564bb9d1304cad21636c89e1dd3198fd4645d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/605?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/old-host-typescript-compat.md
+++ b/.changeset/old-host-typescript-compat.md
@@ -1,0 +1,6 @@
+---
+"@vendoai/actions": patch
+"@vendoai/vendo": patch
+---
+
+Feature detect the TS 4.8 modifier APIs and TS 4.9 `isSatisfiesExpression` in static extraction, falling back to the legacy node properties, so `vendo init` no longer crashes on hosts whose own TypeScript predates those APIs (#551).
--- a/packages/actions/package.json
+++ b/packages/actions/package.json
@@ -80,6 +80,7 @@
     "jose": "^6.2.3",
     "next": "16.2.11",
     "typescript": "^5.6.0",
+    "typescript47": "npm:typescript@4.7.4",
     "vitest": "^2.1.0",
     "zod-to-json-schema": "^3.25.2"
   },
--- a/packages/actions/src/sync/catalog-scan.ts
+++ b/packages/actions/src/sync/catalog-scan.ts
@@ -5,6 +5,7 @@ import type tsTypes from "typescript";
 import type { CatalogEntry } from "../formats.js";
 import { walk } from "./common.js";
 import { parseModule, resolveIdentifier, zodFromExpression, type FileModule, type StaticExtraction } from "./static-ts.js";
+import { isSatisfiesExpressionNode } from "./ts-compat.js";
 
 let ts: typeof tsTypes;
 
@@ -76,7 +77,7 @@ function unwrapExpression(expression: tsTypes.Expression): tsTypes.Expression {
     ts.isAsExpression(current)
     || ts.isTypeAssertionExpression(current)
     || ts.isParenthesizedExpression(current)
-    || ts.isSatisfiesExpression(current)
+    || isSatisfiesExpressionNode(ts, current)
     || ts.isNonNullExpression(current)
   ) current = current.expression;
   return current;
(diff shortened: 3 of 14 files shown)


devanonyme42:
Hi @yousefh409,

With #604 merged, the crash from #551 is gone: the capability gate skips extraction on hosts below the 4.9 floor. This PR sits one layer deeper. With the `modifiersOf` and satisfies fallbacks, hosts on TypeScript 4.7 and 4.8 could keep full extraction instead of the skip. Concretely that means removing `canHaveModifiers`, `getModifiers` and `isSatisfiesExpression` from `REQUIRED_COMPILER_API` and letting the fallbacks cover them, which the tests here verify against a real typescript 4.7.4 install.

Is that a direction you want for the project? If so I will rebase this on top of the gate and adapt its tests accordingly. If you prefer to keep the 4.9 floor simple, happy to close this one.

yousefh409:
Hi @devanonyme42, thank you for this, it is careful work and the reverting-reproduces-the-TypeError test is exactly the right kind of guard.

Closing it because main went a different direction on the same problem in #604 before this could land: instead of per-call-site fallbacks, extraction now gates the whole host-resolved compiler behind a feature-detected API floor (TS 4.9, `packages/actions/src/sync/compiler-gate.ts`) and degrades to no-analysis with a named warning when the floor is not met. The two designs solve the same crash but conflict: the gate deliberately refuses to half-run on an old compiler, while this PR makes it half-work, and we chose the loud floor over silent partial extraction. #551 is closed on that basis.

Sorry the timing burned your effort here, and I would be glad to see you pick up another open issue. Thanks again!

PR #624, opened 2026-07-27
perf: recalibrate gen-scripted:create ceiling 3ms → 8ms (kills false-red mains)

yousefh409:
## Impact

Stops the recurring false-red Perf failures on main — 5 this week from one miscalibrated ceiling, each a failure email with no code change behind it.

## What

`gen-scripted:create` ceiling 3ms → 8ms in `bench/budgets.json`, with a CI-RECALIBRATION entry appended to the rationale per the file's own documented procedure.

**Evidence the ceiling was noise-bound:** main run 30234486368 measured p95 3.17ms against the 3.00ms ceiling with every neighboring metric green and no gen-path change in the blamed commits; dev-machine p95 is ~2.3ms. 8ms ≈ 2.5× the observed CI p95 — same margin philosophy as the existing render-100/render-1000 recalibrations — while an order-of-magnitude engine-overhead regression (the failure this gate exists for) still lands far past it.

## Local gate

Full gate run at this tip: 49/50 turbo tasks green (build, all package suites, typecheck, lint). The one red, `demo-accounting#test` (Cadence Supabase login e2e), is a **local-environment artifact proven unrelated**: an active lane's Supabase stack (`vendo-console-k10`) occupies the default port, so the login probe hits a foreign GoTrue and 401s; the identical failure reproduces with this diff stashed (clean tree). CI provisions its own stack in `cadence-supabase-auth` and is unaffected.

Follow-up worth filing: the login e2e's skip-probe should verify the stack's project identity, not just port liveness — shared dev machines running multiple Supabase projects hit this collision.

---
## Summary by cubic
Increase the `gen-scripted:create` performance ceiling from 3ms to 8ms to stop flaky false-red Perf failures on main. Adds a CI-RECALIBRATION note in `bench/budgets.json` based on CI p95 ~3.17ms (dev ~2.3ms), providing safe headroom while still catching real regressions.

<sup>Written for commit 82074fe33a8ebad6d249d82ab1f3da89b2176d2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/624?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Superseded by #619, which merged while this was in CI and root-caused the same flap: the p95 shift came from #608's per-attempt PipelineEvent (justified observability cost), not pure runner noise — so the 4ms recalibration with the cause named is the better fix than this 8ms noise-margin bump. If the flap recurs at 4ms we can revisit with new CI samples.

PR #638, opened 2026-07-27
[HOLD — numbers are a SIMULATION pending the live run] knowledge: the sufficiency gate (K14)

yousefh409:
> [!WARNING]
> **HOLD — do not merge, and do not quote the numbers below.**
> Every figure in this PR came from a SIMULATION: the passages were
> reconstructed by a different retriever (agreement with the live engine 42/94
> overall, **4/34 on the refusal set**) and the tool path was computed rather
> than driven. The live end-to-end run against a real Agentset namespace,
> through the real `vendo_knowledge_search`, is in flight; its numbers replace
> these, worse or better. The spec's zero-false-answer bar cannot be closed by
> a simulation.
>
> The design also changed after this body was written: the signed spec
> (b5a768442) chose a **sufficiency gate on every hits-returning search** over
> the band-gated shape described below. The band survives only as evidence.

## What

The verifier pass (spec §Agent experience and trust, wave 2 — pulled forward on
production evidence), shipping **opt-in** behind `VENDO_KNOWLEDGE_VERIFY=on`.
On the Cloud engine, in the score region where the evidence bar provably cannot
decide, a cheap model reads the passages the search actually returned and
answers one question — *can this question be answered from these?* — before the
tool reports success. Not answerable → the
tool's existing `insufficient-evidence` outcome, so the agent says it does not
know instead of writing a confident answer over adjacent-topic evidence.

No new tool outcome, no new stream part, no new UI. Same envelope, same
outcome set. **Unset is today's behavior, byte for byte** — the switch exists
because this changes what live users see (some questions that got an answer now
get a refusal, which is the point) and spends a model call on about two thirds
of knowledge turns. That is a host's decision to make on a day it chose, not
one that arrives in a patch release. A value that is neither `on` nor `off`
throws at composition rather than leaving a host believing a trust feature is
running when it is not.

**The band is the design.** Rather than verify every answer, the tool keeps the
threshold where it is right and spends a model call only in the overlap region
of the two calibration populations — `[0.6735, 0.7835]`, derived by
`deriveVerifyBand` from the committed cloud measurements. Below it only
unanswerable questions were ever observed; above it only answerable ones.
Outside the band the routing agrees with the shipped bar 0.7211 on **every**
calibration question, so this is a strict refinement of today's decision, not a
replacement.

## Why

Production calibration (vendo-web `docs/eval/knowledge-cloud/`) found that no
score threshold separates answerable from unanswerable questions on Agentset:
answerable questions score 0.674–0.866, unanswerable ones 0.597–0.783. At the
best possible bar, **16 of 34 unanswerable questions get a confident answer
(47%)** while 7 of 60 answerable ones are refused (12%). Raising the bar above
the highest unanswerable score puts it above the answerable *median*. It is a
property of embedding similarity, which cannot tell "how to install in a
framework" from "how to install in **your** framework" — so no bar fixes it,
and the honest fix is to stop asking the score.

## The measured result

| | shipped bar alone | + verifier in the band |
|---|---|---|
| **False answers** (unanswerable question answered) | **16/34 — 47%** | **1–2/34 — 3–6%** |
| **False refusals** (answerable question refused) | **7/60 — 12%** | **4–6/60 — 7–10%** |
| Turns paying a model call | 0 | 61/94 — 65% |
| Latency added to those turns | — | p50 1.6s · p95 3.6s · cap 5s |

15 of the 16 false answers are fixed and none is introduced. False refusals
improve as well, because inside the band the verifier also rescues answerable
questions the bar was refusing — the honesty is not bought with silence.
Ranges are the spread across repeated passes; a model verdict is not
deterministic and every pass is in the committed artifact
(`docs/eval/knowledge/bands/agentset-verifier-replay.json`).

**Read this before quoting the table: it is a REPLAY, not a live run.** There
is no Agentset key in this environment and the client lives in vendo-web, so
the real half was kept real and only the passages are reconstructed:

- **Real** — every question's band placement is the live run's measured
  deep-intent top score, over corpus and question files that are
  **byte-identical (sha1-verified)** to the ones Agentset was measured against.
- **Real** — the verifier under test is the shipped one, on the shipped
  judge-slot model.
- **Reconstructed** — the passage sets, by dense retrieval over the same corpus
  with the house chunker, because the committed measurements record scores and
  the top document but not the returned text. It is not an easier corpus: it
  gets top-1 right 44/60 against the live run's 46/60, recall@5 57/60. The
  artifact also reports the same before/after on the subset where both
  retrievers picked the same top document.

To unpark into a live run: extend vendo-web's calibration script to dump the
passages it retrieved, and re-run
`pnpm --filter @vendoai/corpus-harness knowledge-verifier-replay`.

## Safety properties (all tested)

Knowledge must never become *less available* because of a component whose job
is to withhold answers:

- **No model configured → no verifier**, today's behavior exactly.
- **No verdict** (timeout past 5s, provider error, off-schema output) → falls
  open to the shipped threshold. Never an error, never a forced refusal.
- **Time-capped**, raced as well as aborted, so a provider that ignores the
  abort signal cannot hold a turn open. The 5s cap is set *from measurement* —
  183 verifications: median 1.6s, p95 4.0s, thin tail to 10s — not from taste.
- **Cloud engine only.** Scores are engine-relative; a number calibrated on one
  engine means nothing on another, so local lexical, BYO and self-hosted
  engines are untouched, switch or no switch.
- **Opt-in, and loud about it:** `VENDO_KNOWLEDGE_VERIFY=on` enables, `off` and
  unset are identical, a typo throws.

## Known limitation (deliberate)

The spec sentence is "checks answer-is-entailed-by-citations before send", but
the tool seam runs *before* the agent drafts anything. So the check is the
strictly-earlier half: "can any correct answer come from these passages?" That
catches the failure this lane exists for — an unanswerable question whose
adjacent-topic passages clear the bar — but not an answer that drifts beyond
passages which genuinely did support the question. The verifier interface takes
an optional drafted answer so a post-draft caller can narrow the check without
a contract change.

## Verification

- [x] `pnpm build && pnpm test && pnpm typecheck && pnpm lint` green — run
      **twice** under the gate lock
- [x] All 100 `@vendoai/knowledge` tests green, including every pre-existing
      one (they are contract criteria): 11 new band/policy tests, 6 verifier
      unit tests (garbage output, hang, no-passages, no-model), 8 composition
      rows in `knowledge-resolution.test.ts` (including default-off, explicit
      off, and the loud typo)
- [x] Frozen contract files untouched (`packages/core/src/knowledge.ts`,
      `knowledge-wire.ts`)
- [ ] Screenshots — n/a, no UI change by construction

## Tuning discipline

This is a model-costed leg, so the frozen-set posture in `docs/eval/GOLDEN.md`
applies. The verifier's standard was written from the calibration's failure
mode before any question was run and was not edited afterwards. The one
parameter set from the run is the 5s cap, taken from the latency distribution
rather than from which questions passed. Per-question verdicts stay out of the
repo.


yousefh409:
**HOLD — do not merge.** Conductor: the headline in this PR ("false answers 47% → 3%") comes from the replay the checker found was **not a real-engine run** — it reconstructed passages with a different retriever, called the verifier directly rather than through the tool path, and hand-computed outcomes; top-document agreement with the real engine was 42/94, and 4/34 on the unanswerable set. The PR body does not disclose that.

Two things before this can merge: (1) the numbers are re-produced by a genuine end-to-end run — real Agentset search on the labelled corpus, through `vendo_knowledge_search` with the verifier composed as a host gets it, a real model call, all 94 questions, raw per-question records committed so every aggregate recomputes; three passes with per-pass figures and the true worst case. (2) the fail-open regression is fixed — enabling verification currently also injects a 0.7211 threshold where main's default is 0, so a 0.61 score becomes a refusal with no verifier call, and a verifier timeout turns an answer into a refusal. That is the opposite of fail-open.

Until then this PR publishes a simulation as a live result.

yousefh409:
Holding this pending lane K15's audit fixes. Four real violations of the spec section this lane implements (`docs/superpowers/specs/2026-07-24-knowledge-block-design.md` §The verifier pass, commit b5a768442):

1. **Enabling verification also raises `weakScoreThreshold` from 0 to 0.7211** (server.ts `CLOUD_KNOWLEDGE_POLICY`). Turning on a verifier must not silently move a refusal threshold — that couples two independent decisions and changes behaviour nobody asked it to change.
2. **No unverified marking.** The spec requires fail-open *but marked*; today a verifier that gave no verdict is indistinguishable from one that passed. This wave learned three times over that a check which could not run must report unknown, never OK — that rule applies to the verifier itself.
3. **The verifier's rationale is dropped** by `adjudicate()`, so nothing downstream can record the gap it found.
4. **It rides the `judge` slot** rather than its own `knowledgeVerifier` slot, coupling two unrelated model roles.

On the fifth audit point — band-gating vs the spec's "not threshold-gated" — **the implementation wins and the spec is being amended.** I wrote that rule to stop scalar thresholds reintroducing the bug; this lane band-gated anyway and *measured* that the band contains all 16 false answers, so the gate costs no correctness and saves ~35% of calls. A prior loses to a number.

Also landing with the fixes: the Cloud default flips from opt-in to **on** (Yousef confirmed). Cloud-only scoping stays correct — the band is Agentset-derived and scores are engine-relative.

The measured result this lane produced is excellent and not in question: false answers 47% → 3–6%, false refusals 12% → 7–10%, one cheap call on ~65% of knowledge turns.

yousefh409:
Superseded by #641: the verifier implementation is inherited from this branch, but every measurement here was a simulation (reconstructed passages, substitute retriever, 42/94 top-document agreement with the real engine) — #641 carries the real three-pass live measurement plus four adversarial checker rounds. Closing.

PR #625, opened 2026-07-27
genui-bench: generation iteration playground (+ Claude 5 engine fix)

yousefh409:
> **Stacks on #613** (de-versioning sweep). The first commit here is #613's; merge that first and this diff shrinks to the playground + engine fix.

## Why

There was no fast way to iterate on the micro-app format, the generation pipeline, or the guardrails: change a prompt or a gate, and the only feedback was a test suite or a full corpus run. `genui-bench` is the inner loop — type a prompt, watch four products answer it side by side, keep the history.

Spec: `docs/superpowers/specs/2026-07-26-genui-bench-playground-design.md` · Plan: `docs/superpowers/plans/2026-07-26-genui-bench-playground.md`

## What it is

A private workspace app (`apps/genui-bench`, never published, never deployed):

- **Four lanes on the same prompt + same host tools:** Vendo, Thesys C1, CopilotKit, Tambo. Each pane carries a permanent footnote stating its paradigm, so an asymmetric comparison is never misread.
- **The Vendo pane is the real runtime, not a preview.** It frames `/embed/<host>?run=<id>`, which boots the production stack (`VendoProvider` → `AppFrame` → Kit + the host's own component registry) in a document whose only stylesheet is the demo host's real `globals.css`. Tool calls execute against canned fixture executors through the real transport; actions, `$state`, and jailed islands all work. Clicking a generated button really calls `host_transferMoney`.
- **Live working tree.** Every run re-imports the engine from source in a subprocess, so an edit to `contracts/` or `validation/` is live on the next run — no build step in the loop.
- **Model controls.** Per-run picker across the Claude line plus temperature / thinking, stamped into each RunRecord; split-compare doubles as a model A/B. Absent a choice, the run is byte-identical to production.
- **RunRecords on disk** with history, pinning, split-compare, and an internals drawer showing the pipeline timeline (outline → compile → guardrail verdicts → repair rounds → smoke render), plus raw wire, final document, and each competitor's raw payload.
- **Same runner headless:** `pnpm --filter genui-bench bench run --host maple --pack smoke --lanes vendo` prints a JSON summary for agents; agent runs appear in the cockpit history automatically.

## Production fix included

[...]

PR #613, opened 2026-07-27
refactor: de-version code names — one live implementation, no V2 suffixes

yousefh409:
## Why

Locked ruling: code never carries version suffixes. There is exactly one live implementation of the format — git is the history. The `-v2` files and `…V2` symbols left over from the v1→v2 rewrite imply a parallel v1 that no longer exists, taxing every reader (and every dispatched agent) and breeding new suffixed names by pattern-matching.

Version tags **inside persisted data** are the deliberate exception and are untouched: `"vendo/app@1"` and `"vendo-genui/v2"` remain byte-identical (external customers have current-format apps at rest in the hosted store, and BYO-Postgres installs are unreachable by definition).

## What

- **File renames (core):** `genui/tree-v2.ts` → `genui/tree.ts` (canonical format + `validateTree`); prior `genui/tree.ts` → `genui/tree-node.ts` (node shape/bindings); `genui/wire-v2/` → `genui/wire/`; internal compat barrels (`src/tree.ts`, `src/tree-v2.ts`, `src/tree-limits.ts`, `src/wire-v2/`) deleted — exports map never exposed deep paths, and zero importers remain.
- **File renames (ui/apps):** `renderer-v2.tsx` → `convert-payload.tsx`; `fake-sandbox-v2.ts` → `fake-sandbox-stateful.ts` (named for its trait, since a distinct `fake-sandbox.ts` already exists).
- **Symbol renames repo-wide:** `compileWireV2`→`compileWire`, `printWireV2`→`printWire`, `compileWirePatchV2`→`compileWirePatch`, `validateTreeV2`→`validateTree`, `VENDO_TREE_FORMAT_V2`→`VENDO_TREE_FORMAT`, `TreeV2`→`Tree`, `TreeQueryV2`→`TreeQuery`, `convertV2Payload`→`convertPayload`, harness/test doubles, etc.
- **npm compat:** `@vendoai/core` is published (0.4.x), so every previously-public V2 name lives on as a `@deprecated` alias in one marked block at the bottom of `index.ts` — remove next minor. `@vendoai/ui`/`@vendoai/apps` exported no V2 names.
- **Deliberately still versioned:** persisted-data refs (`e2b:v2:` snapshot prefixes, store crypto v2 envelope, fixture tool ids), external names (AI SDK `LanguageModelV2`), and spec-document citations ("v2 spec §N") — all data or document versioning, not code identity.

## Proof

[...]

PR #639, opened 2026-07-27

yousefh409:
Ships **D5** of the [generation pipeline redesign spec](docs/superpowers/specs/2026-07-27-generation-pipeline-redesign-design.md) — signed to go first, on its own PR, ahead of the rest of the wave. It is a safety fix, not a quality tradeoff.

Stacks on #613 and #625 (base is `yousefh409/genui-bench`; the two spec commits ride along).

## The defect

```js
  amount: 1,
});
```

| case | verdict |
| --- | --- |

## Measurement

| --- | --- | --- |

## Gates

`pnpm build && pnpm test && pnpm typecheck && pnpm lint` all green, plus `pnpm --filter @vendoai/apps test` (731 passed) and `pnpm --filter genui-bench test` (122 passed).

---
## Summary by cubic
Prevents capability substitution by rejecting mutating tool calls with fabricated targets or amounts. When a host lacks a needed capability, the code now routes to the disclaimer instead of improvising with tools like `host_transferMoney`.

- **Bug Fixes**
  - Add a gate in `@vendoai/apps` that blocks mutating tools when target/amount operands are bare literals not sourced from user input, tool data, row context, or a declared enum/flag.
  - Covers both surfaces: declarative `{action, payload}` and island `tools.*(...)`, on create and edit.
  - Exemptions: user-named values (numbers via exported `requestNumberValues` in `@vendoai/core`, strings matching the request), enums/flags/consts, variables/spreads; content keys like `memo` are ignored.
  - Violations trigger repair with `DISCLAIMER_TEXT` (moved to its own module). Tests include the live Slack-memo transfer (FAIL) and the correct client-message flow (PASS).

- **New Features**
  - Add `genui-bench` measurement CLI and sweep script to track island-escape metrics and compare arms.
  - Results (58 apps, both hosts): failure rate 7% → 2%; p50 16.1s → 16.9s; p90 103.3s → 81.7s; this gate rejected 0 apps. Offline replay over 187 apps flags exactly the live defect.

<sup>Written for commit 0c2d7d864917bf4fc342467f9ca87d61a2c79150. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/639?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Superseded by **#648** — closing.

This PR sat three-deep on a stack whose base was 21 PRs stale. The replacement chain is #644 (fresh de-versioning sweep, now also carrying the squash-merged playground from #645) → #648 (this gate). The three generation-redesign spec commits that rode along here landed separately in #643, already merged.

#648 also fixes the two Greptile P1s that were raised here and never triaged: the numeric-spelling bypass (`amount: 0x1 / 1e0 / 1_00 / +1 / (1)` all classified as references, so the mutating call shipped) and the follow-up spaced-unary-sign bypass (`amount: + 1`). The literal classifier now peels wrappers structurally — parens and unary signs, any order, any spacing — instead of enumerating spellings, so the class is closed rather than the case. `capability-substitution.test.ts` is 20 → 44 tests, with `count + 1`, `-count`, `(count)` and `(a) + (1)` as controls.

PR #646, opened 2026-07-27

yousefh409:
> **Re-land of #639 against current main.** The original sat three-deep on a stack whose base was 21 PRs stale. It is replaced by #644 (fresh de-versioning sweep) → #645 (playground + Claude 5 fix) → this. The three generation-redesign spec commits that used to ride along here are **not** in this branch — they land independently in #643. #639 is superseded and closed.
>
> **Base is `yousefh409/genui-bench-2` (#645).** Merge order: #644 → #645 → this. Cherry-picked clean; no conflicts.

Ships **D5** of the generation pipeline redesign spec (`docs/superpowers/specs/2026-07-27-generation-pipeline-redesign-design.md`, landing via #643) — signed to go first, on its own PR, ahead of the rest of the wave. It is a safety fix, not a quality tradeoff.

## The defect

```js
  amount: 1,
});
```

| case | verdict |
| --- | --- |

## Measurement

| --- | --- | --- |

## Gates

Re-verified on the re-landed branch: `pnpm build` 23/23 · `pnpm test` 53/53 · `pnpm typecheck` 41/41 · `pnpm lint` 6/6, plus `pnpm --filter @vendoai/apps test` (740 passed / 18 skipped).

---
## Summary by cubic
Adds a gate that rejects mutating tool calls with fabricated targets or amounts when the host lacks the needed capability, preventing write APIs from being used as message channels and routing to the disclaimer instead. Now also catches all hand‑typed number spellings in amount operands.

- **Bug Fixes**
  - Blocks mutating calls with hand-typed target/amount unless values come from user text, queries, form state, row context, or declared enums/flags/consts.
  - Applies to `{action, payload}` and `tools.*(...)` in islands; edit path threads the instruction so recent user values are exempt.
  - Violations point to `DISCLAIMER_TEXT` as the fallback.
  - Numeric literal detection hardened to include signed, parenthesized, separator, exponent, and hex/octal/binary forms; arithmetic expressions (e.g., `count + 1`) still pass.
  - Exported `requestNumberValues` from `@vendoai/core` to share the user-number carve‑out; tests and bench utilities updated, with sweep improving failure rate 7% → 2% and replay flagging only the live defect.

<sup>Written for commit baade1c4e8db5454ec4eb3523e248d2e050a760c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/646?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Addressed Greptile's P1 (`baade1c4e`) — the same finding was raised on #639 and never triaged, so it is fixed here rather than carried forward.

`sourceArgValue` tested numeric operands with a decimal-only pattern, so `amount: 0x1`, `1e0`, `1_00`, `+1` and `(1)` classified as `other` — read as a *reference*, which is exactly the verdict that lets a mutating call through. The live defect's sentinel was `amount: 1`; four keystrokes would have walked around the gate.

The classifier now recognises signed, parenthesised, separator-bearing, exponent and hex/octal/binary literals, with a `Number()` round-trip as the final arbiter so the deliberately loose pattern cannot admit a non-number. An arithmetic expression (`count + 1`) is still a reference and still passes — that control is a test, not a claim. 10 added counter-examples; `capability-substitution.test.ts` is 20 → 30 tests, all green.

yousefh409:
Superseded by #648 — GitHub auto-closed this PR when its base branch (`yousefh409/genui-bench-2`) was deleted on the #645 squash-merge, and it cannot be reopened once the base is gone.

#648 is the same work rebuilt on `yousefh409/deversion-2` (#644), plus the two Greptile P1s that were open here: the numeric-spelling bypass and the follow-up spaced-unary-sign bypass. The gate's literal classifier now peels wrappers structurally instead of enumerating spellings; `capability-substitution.test.ts` is 20 → 44 tests.

PR #648, opened 2026-07-27

yousefh409:
> **Re-land of #639, rebased onto current `main`.** The original sat three-deep on a stack whose base was 21 PRs stale. That stack has now landed: #644 (the fresh de-versioning sweep, carrying the squash-merged playground from #645) is on `main`, and the three generation-redesign spec commits that used to ride along here went in via #643. This PR is the last piece — five commits, 20 files, based directly on `main`.
>
> It also replaces #646, which GitHub auto-closed when its base branch was deleted on the #645 squash-merge and which cannot be reopened once the base is gone.

## What is in this PR beyond #639

Three defects the review bots found in the gate itself, plus one gate-ergonomics fix:

- **`fix(apps): the substitution gate reads every spelling of a hand-typed number`** — Greptile P1, raised on #639 and never triaged there. `sourceArgValue` tested numeric operands with a decimal-only pattern, so `amount: 0x1`, `1e0`, `1_00`, `+1` and `(1)` classified as `other` — read as a *reference*, the verdict that lets a mutating call through. The live defect's sentinel was `amount: 1`; four keystrokes walked around the gate.
- **`fix(apps): peel every wrapper before judging a hand-typed number`** — the follow-up P1. Enumerating spellings left the next combination open (`amount: + 1`, a space after the sign). The classifier now *peels* the syntax that can wrap a bare number without making it anything else — parens and unary signs, in any order, any spacing — then tests the core. That closes the class instead of the case. Controls are tests too: `count + 1`, `-count`, `(count)` and `(a) + (1)` stay references. 
- **`fix(apps): a type assertion does not launder a hand-typed number`** — a third Greptile P1, reproduced before fixing. Island source is TSX, so `amount: 1 as number`, `1 as const`, `1 satisfies number` and `1 as unknown as number` are legal and all read as references. Assertions now peel in the same loop, one per pass so chains unwind; `count as number` and `row.amount as const` are controls. `capability-substitution.test.ts` is 20 → 51 tests.
- **`fix(genui-bench): typecheck ignores the dev server's racy type scaffold`** — running the repo's documented gate in its documented order (`build && test && typecheck`) went red locally on genui-bench alone: its tests boot `next dev`, whose typed-route scaffold is written concurrently by several vitest workers and lands corrupt, and Next re-adds that path to tsconfig's `include` on every run. Excluding `.next/dev` wins over the include Next keeps restoring; `next build` route types are still checked. This belongs to #645's app but that PR was already merged, so it rides here.

Ships **D5** of the generation pipeline redesign spec (`docs/superpowers/specs/2026-07-27-generation-pipeline-redesign-design.md`, merged via #643) — signed to go first, on its own PR, ahead of the rest of the wave. It is a safety fix, not a quality tradeoff.

## The defect

```js
  amount: 1,
});
```

| case | verdict |
| --- | --- |

## Measurement

| --- | --- | --- |

- **An amount that coincides with any number in the request.** `requestNumberValues` records every number the user typed and its cent-scaled form, so "show the first 10 transactions" makes `10` and `1000` user-supplied for the whole app. Deliberate: this is the same definition of "a number the user named" that law 1's island math uses, and a second stricter copy living only in this gate is the parallel mechanism the spec forbids. It is also not a bypass on its own — a substitution must still ground its target (recipient/id) in tool data, user input, or the acted-on row, and the live defect failed on both operands. Tightening it is a quality tradeoff that needs its own 58-app sweep: the previous attempt at tightening grounding here regressed failure rate 7% → 28% and was thrown away.

## Gates

Re-verified on the rebuilt branch, in the documented order: `pnpm build` 23/23 · `pnpm test` 53/53 · `pnpm typecheck` 41/41 · `pnpm lint` 6/6, plus `pnpm --filter @vendoai/apps test` (771 passed / 18 skipped).


yousefh409:
Greptile triage — one fixed, one rejected with reasoning.

**FIXED — "Type assertions hide numeric literals" (`b9067c2a4`).** Real, and I reproduced all four forms before fixing: island source is TSX (esbuild `tsx` loader in `islands.ts`), so `amount: 1 as number`, `1 as const`, `1 satisfies number` and `1 as unknown as number` are legal syntax and every one of them read as a *reference* — the verdict that lets the mutating call through. Assertions now peel in the same loop as parens and unary signs, one per pass so chains unwind. A mis-peel is harmless by construction: whatever peeling yields still has to pass the numeric-core test, so the worst case is the behaviour without peeling. `count as number` and `row.amount as const` are controls and still pass. 51 tests in the file, up from 20 on #639.

**REJECTED — "Request-wide numbers bypass grounding".** This is the gate's signed design, not a defect, and narrowing it here would be the wrong trade:

1. `requestNumberValues` is deliberately the *same* definition of "a number the user named" that law 1's island-math carve-out already uses. A second, stricter definition living only in this gate is exactly the parallel mechanism the spec forbids — two answers to one question, drifting apart on the next edit.
2. The finding understates what a bypass costs an attacker-model. The amount alone passing is not enough: the gate flags *any* fabricated target-or-amount operand, so a substitution still has to ground its **recipient/id** in tool data, user input or the acted-on row. The live defect failed on both operands.
3. This area has a measured history. The previous attempt to tighten grounding here regressed failure rate 7% → 28% and was thrown away; the hard ceiling on this PR is *never worse than baseline*, proven over a 58-app sweep. Tightening the number carve-out is a quality tradeoff I cannot claim is safe without re-running that sweep against live models — and a plausible-looking narrowing that silently costs 4× the failures is precisely how the last one shipped.

The residual is now written into the PR body under **Known residual, out of scope** so it is a recorded decision rather than an oversight.

yousefh409:
⚠️ **`main` is currently red, and this PR carries the fix.**

`pnpm install --frozen-lockfile` fails on main at `0e3bc0a9c`, so **every** CI job dies in setup — the last three runs on main are all failures. It is not a flake:

```
ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY: Broken lockfile: no entry for
'next@16.2.11(@babel/core@7.29.7)(@opentelemetry/api@1.9.1)(@playwright/test@1.61.1)
 (@types/node@20.19.43)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)'
```

That is genui-bench's exact `next` peer-set. Cause: GitHub's **squash**-merge of #644 applies the branch's `pnpm-lock.yaml` wholesale against the base. #644's branch last took main at `1ddba3456`; #641 and #647 landed after that, and the squash discarded the resolution entries they had added while their importers stayed behind. A merge commit would not have done this — a squash of a long-lived branch touching a lockfile will, every time.

`9037f428c` regenerates the lockfile against main's current package.json graph. `pnpm install --frozen-lockfile` is clean, and the full gate is green: build 23/23 · test 53/53 · typecheck 41/41 · lint 6/6.

**Merging this PR repairs main.** If you would rather unblock main immediately without waiting on this review, the lockfile commit stands alone and can be cherry-picked straight to a hotfix branch.

PR #649, opened 2026-07-27

yousefh409:
Re-land of #639/#646 against current main (their stacked bases were squash-merged, so the originals could not be reopened cleanly). Same five commits, cherry-picked; full local gate green.

## The defect

Found live in the genui-bench playground. Asked for "a button to approve all requests and forward to Slack" on a host with **no Slack tool**, the engine produced an island that called:

```js
tools.host_transferMoney({ amount: 1, recipient_name: 'Slack Forwarding Bot',
                           memo: 'APPROVED TRANSACTIONS: ' + summary })
```

A payments API repurposed as a messaging channel — its own generated comment says so. Measured incidence of islands reaching a mutating tool: 9% of baseline apps.

## The rule — provenance, not intent

A mutating tool (`risk: write | destructive`) invoked with a bare literal in a **target-or-amount** operand, where that literal is neither something the user named nor a closed vocabulary the tool declares, is invalid. Deliberately mechanical: no LLM intent-matching. It reuses `literals.ts`'s fabricated-literal machinery rather than adding a parallel mechanism.

Why it discriminates the real cases: the Slack call hand-types both the *who* and the *how much* — nothing chose them. Legitimate mutating calls reach their operands through a reference (a bound row id, form state, a value the user typed). Content keys (`memo`, `body`, `message`) are excluded — canned copy is legitimate.

The issue message routes through repair and names the honest `Disclaimer` path as the correct answer when a host lacks a capability.

## Proof

- **20 counter-example tests.** FAIL: the Slack case; fabricated recipient alone; fabricated amount alone; fabricated id nested in an object. PASS: the Cadence "chase every overdue invoice" app calling `host_sendClientMessage` with invoice-derived args; form submit with field-derived values; row action with a bound id; enum/flag literals; and the real "Send $25 to Alex Rivera" app — the user named both amount and recipient. Plus an end-to-end replay of the recorded defect through the real create loop.
- **Replay over 187 recorded apps** across four corpora, including all 12 legitimate mutating calls: **exactly 1 flag — the real defect. Zero false positives.**
- **Measured 58 apps, both hosts, all packs:** failure rate 7% (baseline) → **2%** (gated). The hard ceiling was "never worse"; it improved.

## Out of scope (stated, not hidden)

A mutating tool used for the wrong purpose with *properly grounded* operands needs semantic judgement this gate refuses. Tools with no target/amount parameters are invisible to a provenance rule. Both noted for the D2/D3 work.

Spec: `docs/superpowers/specs/2026-07-27-generation-pipeline-redesign-design.md` (D5, signed to ship first).

---
## Summary by cubic
Blocks mutating tools from being repurposed when a host lacks a capability. The engine now rejects `tools.*` calls and action payloads with hand‑typed target/amount and routes to the disclaimer instead, fixing the live `host_transferMoney` → “Slack” misuse.

- **Bug Fixes**
  - Capability-substitution gate for mutating tools (`risk: write | destructive`) on both surfaces: `{ action, payload }` bindings and island `tools.*(...)` calls.
  - Allowed operands: tool data, user input, form state, bound row ids, and user‑named values; enums/flags/consts always pass. Content fields like `memo`/`message` are ignored.
  - Robust literal detection: peels parens, unary signs, and TS type assertions; recognizes hex/octal/binary, exponent, and separator forms so `1`, `+1`, `(1)`, `0x1`, `1e0`, `1_00` are all treated as hand‑typed numbers.
  - Violations route to repair with the honest disclaimer: "This part of the request isn't available on this host."
  - Integrated into create and edit validation; repair now binds context (no fabricated ids). Targeted tests cover the live defect and legit cases.
  - Pinned `next` resolution in `apps/genui-bench` lockfile to fix CI frozen installs.

- **New Features**
  - Bench measurement tooling in `apps/genui-bench`: metrics, report generator, and `sweep.sh` for before/after runs. Results (58 apps, both hosts): failure rate 7% → 2%; no apps blocked by this gate; replay over 187 apps flagged exactly the one live defect.

<sup>Written for commit b9140001a9bdea182d065585790a068172e8d5f7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/649?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Duplicate of #648, which landed the same gate plus four hardening fixes (wrapper-peeling literal classifier: 0x1 / +1 / 1 as const no longer read as references) and the lockfile repair. Closing.

PR #651, opened 2026-07-28
fix interpolated text bindings

### LINKED ISSUE #631: generation: models reflexively interpolate bindings into text strings; dialect forbids it → build fails (repair should desugar)
**The single biggest generation-reliability wall for data-bound apps**, surfaced while trying to make console draft-Preview render data widgets (vendo-web #192).

## Symptom
Models — Haiku AND Sonnet — routinely write a data binding *inside* a text string:
```
<Text text="Balance: {summary.totalBalanceCents}"/>
```
The wire dialect forbids interpolation inside strings ('embeds a binding inside a string — string interpolation is unsupported; bind the prop to a single {reference} or split the text'), so the build fails. It's the dominant failure once tool wiring is correct.

## Why it matters / why it's not fixable upstream of the engine
- It's the model's *most natural* output — writing a value into a sentence. Prompt guidance against it doesn't hold, and piling on rules makes generation worse.
- **A stronger model is WORSE, not better**: measured on a fixed data-hungry brief with fully-correct tool wiring (shapes + semantics + host_ names), Haiku ≈ 61% build success, Sonnet ≈ 39% — Sonnet writes richer sentences and trips the rule *more*. So this cannot be bought with model spend.
- The other residual failures are the same shape (wrong prewired-component props like `style`/`whole`, hand-typed literals) — all 'the model wrote plausible markup the strict dialect rejects.'

## Proposed fix (engine-level, helps ALL generation not just preview)
Teach the generation repair pass to **desugar** an interpolated string into the legal form automatically: `text="Balance: {q.x}"` → a Stack of a Text label + a bound data component (or a Text with a single reference). The engine already has an internal attempt/repair loop; it exhausts on this class instead of fixing it. Auto-repairing the most common model instinct would materially raise data-bound generation reliability everywhere (runtime chat, app generation, preview).

## Evidence
Preview sample-catalog experiment (branch `yousefh409/preview-catalog-rewire` in this repo's vendo-web sibling): correct wiring took the data-bound path 44% → 61%, but it plateaus there against this wall. Full measurement trail + harness on that branch.

erichanwang:
fixes #631

what changed:
- split generated `Text` strings like `Balance: {metric.totalCents}` into literal `Text` fragments plus bound `Text` nodes during structured repair
- keep the patch narrow to prewired `Text.text`; other interpolation shapes still fall back to the existing repair/regeneration flow

why:
models commonly put data bindings inside copy strings, but the wire dialect does not support interpolation. this handles the common safe case deterministically before spending another model round.

checks:
- `pnpm --filter @vendoai/apps typecheck`
- `pnpm --filter @vendoai/apps test -- src/engine-laws.test.ts` (ran the full @vendoai/apps suite: 772 passed, 18 skipped)
- `pnpm build`
- `pnpm lint`
- `git diff --check`

### DIFF
--- a/packages/apps/src/engine-laws.test.ts
+++ b/packages/apps/src/engine-laws.test.ts
@@ -152,6 +152,23 @@ describe("law 1 — data-classed props must be bindings", () => {
     expect(nodes(document).some((node) => node.component === "DataTable")).toBe(true);
   });
 
+  it("desugars interpolated Text strings into literal and bound Text nodes", async () => {
+    const wire = '<App name="Balance"><Query id="metric" tool="host_metric"/><Text text="Balance: {metric.totalCents}"/></App>';
+    let calls = 0;
+    const model = scriptedLanguageModel(() => {
+      calls += 1;
+      return wire;
+    });
+    const document = await modelEngine.create({ prompt: "Show balance" }, deps(model));
+    const row = nodes(document).find((node) => node.component === "Row");
+    const textNodes = nodes(document).filter((node) => node.component === "Text");
+
+    expect(calls).toBe(1);
+    expect(row).toBeDefined();
+    expect(textNodes.map((node) => node.props?.text)).toContain("Balance: ");
+    expect(textNodes.map((node) => node.props?.text)).toContainEqual({ $path: "/metric/totalCents" });
+  });
+
   it("splices a literal-data fault through the strict structured-repair fix space", async () => {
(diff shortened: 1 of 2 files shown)


yousefh409:
Thanks — unlucky timing: the generation engine this patches was rewritten hours after the PR opened, so the target files are gone. #631 stays open for a fresh fix against the new engine. Closing.

PR #680, opened 2026-07-28
fix(ui): one visitor, one anonymous identity — consent writes land again

yousefh409:
## The bug

An anonymous visitor's identity **is** the opaque session pointer the door mints on a cookie-less wire request, and the door mints one **per request**. A cold page load mounts several hooks at once — `/status`, `/approvals`, `/automations`, `/activity`, `/connections/catalog`, `/connections` — so every one of them leaves cookie-less and mints its own subject. The browser's jar keeps whichever `Set-Cookie` landed last; the rest are orphaned. Measured live on the demos: **one page load produced four distinct subjects, three orphaned.**

The damage lands on the trust mechanism at the centre of the product:

- the run creates its consent approval under subject A
- the user's **Approve** arrives as subject B
- `packages/guard/src/guard.ts:1037` correctly refuses another subject's approval
- it surfaces as `Approval apr_… was not found`, and the run sits on *"waiting for your approval"* forever

**Every consent-gated action silently fails.** The same split is why the activity feed reads empty during a live run.

This was previously fixed host-side only (mint the pointer once on the page response). Any customer embedding Vendo the documented way still has the bug, because the SDK still mints per request — it is live in the published 0.5.0. This is the SDK-side fix.

## Why the fix is in the client, not the door

The door genuinely **cannot** close this alone: two cookie-less requests are indistinguishable from two visitors, and the cookie is the only identity there is.

But the **browser is the visitor boundary**, and every route the cold burst touches already funnels through `createVendoClient`'s single `request()` helper. So the client is the layer where the race closes honestly: the first request through a client may leave cookie-less, and every request issued before it answers waits for it, then travels with the pointer it established. One extra round trip on a cold load, nothing afterwards. A failed first request **releases** the gate rather than holding it, so today's behaviour is the floor — never something worse.

### Two mechanisms considered and rejected as unsound

Both were plausible-sounding suggestions, and the tests state the reasoning so they don't get "fixed" in later:

1. **Fingerprinting the requester (IP / User-Agent) to single-flight the mint** — would merge two real visitors behind one corporate NAT into a *single* session, sharing threads, grants and approvals. That is a worse bug, and it breaks the invariant `wire/context.ts` documents ("two anonymous visitors never share threads, grants, approvals, or apps").
2. **Deriving the pointer deterministically from request attributes** — would make a live session *guessable*, where today guessing one is a 2^128 search.

## Tested as a race — the detail that matters most

A sequential probe **cannot** see this bug: the first request's `Set-Cookie` is in the jar before the second leaves, so requests issued one at a time always agree. An earlier probe "eliminated" this very hypothesis for exactly that reason and was wrong.

`packages/ui/test/anon-session-race.test.ts` uses a **real `node:http` door** reproducing the mint-per-cookie-less-request contract, a **real browser-style cookie jar**, and fires the measured six-call burst **concurrently**:

| | identities | mints |
|---|---|---|
| before | **6** | 6 |
| after | **1** | 1 |

Red/green was verified by reverting the fix (`expected 6 to be 1`) and restoring it.

The second test in that file is the **control**: it passes both before *and* after by design, demonstrating why a sequential probe misses this.

`packages/vendo/src/anon-session-race.test.ts` pins the **door half** of the contract against silent regression, using the real door and a real store:

- the door mints one pointer per cookie-less request (stated plainly, because that is *why* the race isn't solvable there)
- a burst that already carries a pointer yields **exactly one session and zero re-mints** — which is also what keeps hosts that mint on their document response working unchanged

That second test also caught a real hazard worth knowing: on an `https` request the door reads **only** the `__Host-` prefixed cookie name. Send the plain name and it reads as absent and mints its own — silently back to per-request identities. A browser-side fix never has to know this; anything minting the pointer by hand does.

## Known residual (not fixed here)

`packages/ui/src/hooks/use-vendo-thread.ts:102` and `packages/ui/src/voice/voice-act.ts:156` post to `/threads` through a bare `fetch` (they need raw headers), so they bypass the gate. In practice this is unreachable: on any chrome-mounted page `/status` fires at mount long before a user can send, so the cookie already exists; and on a bare `<VendoThread>` page the POST is the *only* request, so there is no race. Closing it properly means sharing the gate with those transports without widening the public client surface — worth a follow-up, deliberately not bundled here.

## Verification

- `pnpm build` 23/23 · `pnpm test` 53/53 · `pnpm typecheck` 41/41 · `pnpm lint` 6/6
- full suite run **twice**, the second with `TURBO_FORCE=true` (0 cached, 53/53)
- `packages/ui` alone: 80 files / 665 tests green

No UI/visual surface changes — this changes request sequencing only, so there are no screenshots to attach.

---
## Summary by cubic
Fixes an anonymous-session race so one visitor keeps one identity. This restores consent approvals and live activity by making concurrent cold-load requests share the same cookie.

- **Bug Fixes**
  - Gate session minting in `@vendoai/ui`: the first request establishes the cookie; concurrent requests wait and reuse it.
  - On failure, the gate releases (falls back to current behavior). No UI changes.
  - Added race-focused tests in `packages/ui` and `packages/vendo` to prevent regressions.

<sup>Written for commit a907dd95a95ac7980bf7dd67ddf71109dffb65bb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/680?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/one-visitor-one-anonymous-identity.md
+++ b/.changeset/one-visitor-one-anonymous-identity.md
@@ -1,0 +1,34 @@
+---
+"@vendoai/ui": patch
+---
+
+One visitor, one anonymous identity — consent-gated actions stop failing silently.
+
+An anonymous visitor's identity IS the opaque session pointer the door mints on a
+cookie-less wire request, and the door mints one PER REQUEST. A cold page load
+mounts several hooks at once (`/status`, `/approvals`, `/automations`,
+`/activity`, `/connections/catalog`, `/connections`), so every one of them left
+cookie-less and minted its own subject; the browser's jar kept whichever
+`Set-Cookie` landed last and the rest were orphaned. Measured live: one page load
+produced four distinct subjects, three orphaned.
+
+The damage lands on the trust mechanism at the centre of the product. An agent
+run created its consent approval under one subject, the user's Approve arrived as
+another, and guard correctly refused another subject's approval — surfacing as
+`Approval apr_… was not found` and a run stuck on "waiting for your approval"
+forever. Every consent-gated action failed this way, and the same split emptied
+the activity feed mid-run.
+
+The browser is the visitor boundary, so `createVendoClient` is the layer that can
+close the race honestly: the first request through a client may leave
+cookie-less, and every request issued before it answers now waits for it and
+travels with the pointer it established. Costs one extra round trip on a cold
+load and nothing afterwards; a failed first request releases the gate rather than
+holding it, so the old behaviour is the floor, never something worse.
+
+Deliberately NOT solved by fingerprinting the requester (IP/User-Agent would
+merge two real visitors behind one NAT into a single session, sharing threads,
+grants and approvals) nor by deriving the pointer from request attributes (that
+would make a live session guessable, where today it is a 2^128 search). Hosts
+that already mint the pointer on their document response keep working unchanged —
+the door treats a pre-established pointer as canonical.
--- a/packages/ui/src/client-impl.ts
+++ b/packages/ui/src/client-impl.ts
@@ -80,7 +80,7 @@ export function createVendoClient(config: VendoClientConfig): VendoClient {
   const baseUrl = config.baseUrl ?? "/api/vendo";
   const headers = { ...(config.headers ?? {}) };
 
-  async function request(path: string, init?: RequestInit): Promise<Response> {
+  async function send(path: string, init?: RequestInit): Promise<Response> {
     return ensureOk(
       await fetch(route(baseUrl, path), {
         ...init,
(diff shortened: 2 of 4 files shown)


yousefh409:
## Real-browser proof (added after opening)

Run in real Chromium against the `packages/ui` e2e harness (`/page-chat`), whose `/api/vendo` is proxied same-origin to a real `node:http` wire server.

**Metric:** how many wire requests leave the browser *before the first one answers*. Only a request in that window can be cookie-less, so that count **is** the number of identities the door would mint.

| | racing cookie-less requests |
|---|---|
| without the fix | **4** |
| with the fix | **1** |

8 wire requests were observed either way: `/status`, `/approvals`, `/connections/catalog`, `/threads`, `/connections`, `/threads`, `/threads/thr_1`, `/connections`.

The **4** is worth noting: it independently reproduces, in a browser, the *four distinct subjects* measured live on `demos.vendo.run` from a single page load.

### The control that mattered

With **no injected latency** the same measurement reads `1` in *both* directions — the local fixture answers faster than React mounts the next component, so every request looks serialized. A probe run that way "confirms" the fix while being completely blind to the bug.

That is the same class of error as the earlier sequential probe that "eliminated" this hypothesis and was wrong. I only caught it because I ran the negative control (fix reverted) before trusting the positive result. 300ms of injected latency opens the window a real cold load over the internet actually has.

The spec was run as **evidence and deliberately not committed**: 300ms × 8 requests is timing-sensitive on a permanent CI gate, and `packages/ui/test/anon-session-race.test.ts` already asserts the same thing deterministically. Happy to add it behind a tag if you'd rather have it in the suite.

PR #686, opened 2026-07-28
fix(bench): stop generating an automation beat, and reject any pill that promises recurrence

yousefh409:
The generated arc required an `"automation"` beat — *"a recurring version of the work ('every Monday, ...')"*. Nothing registers that job, so every demo shipped a pill whose promise died in front of the prospect.

## What I measured on the live shared host

| probe | result |
|---|---|
| the beat **verbatim** | agent built a static view and **dropped the recurrence silently** — never said nothing was scheduled |
| explicit *"Set up a recurring scheduled job…"* | *"Confirmed — I don't have a true cron/recurring-job scheduler."* |
| an app prompt containing `cron`, `Recurring`, `scheduled` verbatim — the very vocabulary `SERVER_INSTRUCTION` matches | still a plain tree app |

`GET /automations` was empty after all three. The silent case is the worse one: the prospect believes it worked.

## Why the pipeline stops asking instead of the demo being wired

The capability is **not** missing. The automations engine is composed in `createVendo`, the `armAutomation` seam is wired, and the deployed host's own log confirms Cloud is the firing authority for schedule triggers under the hosted store. What is missing is the **path from a chat request to an armed automation**:

1. `packages/apps/src/agent-tools.ts` — `vendo_apps_create`'s description never tells the agent that app creation is the door to a schedule, so it refuses up front instead of trying.
2. `packages/apps/src/runtime.ts` (create path, `rung === "steps"`) — the escalation is best-effort behind a bare `catch {}`, so a failure degrades to a static tree app with **no signal to the user and none in the logs**.

Both are in `@vendoai/apps` and want their own change plus a release. The demo host pins `@vendoai/*` at `0.5.0`, so until that ships the pipeline must not ask for the beat — a pill that says "scheduled!" while nothing fires is exactly the hollowness this wave removes.

## The change

- **`requiredBeatKeys` drops `"automation"`.** Four kinds; `mergeBeats` fills the fifth chip slot with a **derived** pill instead — grounded in the demo's real tools, so the strip stays full without promising anything.
- **`readsAsRecurring` moves to `demo-folder.ts`** (where beat rules live) and **inverts**: recurrence was a requirement for one beat, and is now a *problem* on any beat. It immediately caught three `connect-account` pills in shipped demos — `"every Monday"`, `"automatically"`, `"whenever I star a dashboard"` — the same unregistered job under a different key.
- The beats agent is told not to write one, and `regroundBeats` will never stand a recurring pill in for an ungrounded beat.

## Side effect worth naming

Removing the requirement **unbreaks `demo:fix`** for `atlas-freight` and `linear-issues`, which predate the five-kind arc: `demo:fix --id atlas-freight` failed with `REJECTED: missing beat(s): automation, connect-account`. `beatSetProblems` now returns `[]` for all seven live demos (the `connect-account` half is fixed in `runvendo/vendo-demos` #12/#13).

## Test changes, stated out loud

Fixtures encoded the old contract, so they moved with it: `fullArc()` is four kinds; the two `regroundBeats` tests that asserted the automation special-case now assert the general rule (recurring pill never substitutes, one-off does); the three `toEqual([...requiredBeatKeys])` assertions became `[...requiredBeatKeys, "overdue"]` because the fifth slot is now a derived pill; and the duplicate/too-many fixtures switched off the removed key.

## Gates

`pnpm build`, `pnpm test` (529, bench ran fresh — the rest cached and untouched), `pnpm typecheck` (41/41), `pnpm lint --force` (0 cached) — green twice, second time after fetch/rebase onto `origin/main` (`a4a79d91e`).

---
## Summary by cubic
Stop generating the "automation" beat and reject any recurring/scheduled pills so demos don’t promise schedules the host can’t register. The chip strip still shows five pills by deriving a non-recurring one.

- **Bug Fixes**
  - Removed "automation" from `requiredBeatKeys` (arc is now four beats).
  - Moved `readsAsRecurring` into beat rules and inverted it to flag any recurring promise; build fails such pills.
  - Updated the build prompt and `regroundBeats` to never accept recurring pills as stand-ins.
  - Adjusted tests to the new arc and expectations (e.g., expected keys are `[...requiredBeatKeys, "overdue"]`; error now omits "automation").
  - Unblocks `demo:fix` for `atlas-freight` and `linear-issues`.

<sup>Written for commit bae494ee18b81019fae66dafa17291753122d599. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/686?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Target files were deleted in #831; superseded.

PR #689, opened 2026-07-29
fix(cli): onboarding safety and honesty — gitignore warn, single askYesNo, honest keyless line, pages-router/monorepo guidance

yousefh409:
Four onboarding fixes, all in `packages/vendo/src/cli/`. Every one is a thing the first `vendo init` got wrong in a way the dev could not see.

## 1 — A secret written into a committed file now says so

`vendo login` and `vendo init --cloud-key` land `VENDO_API_KEY` in `<root>/.env.local`. Nothing checked whether git would commit it.

Both call sites now follow the write with `git check-ignore -q .env.local` from the target root and print, when it is NOT ignored:

```
warning: .env.local holds a secret and is NOT gitignored — add `.env.local` to .gitignore before you commit.
```

- **Never blocks the write.** The key is already minted and is never printed anywhere, so failing the write would lose it. Warn only.
- **Silent when there is nothing to leak into.** `git check-ignore` exit 1 is the only "not ignored" answer; exit 128 (not a repo) and a failed spawn (no git) stay quiet.
- The warning is emitted at the call sites, after the "wrote it" line, so it reads as part of that narration and can never turn a successful mint into device-login's write-failure copy.

## 2 — Exactly one `askYesNo`

The unguarded duplicate in `extract/extraction.ts` — identical body, no `if (!stdin.isTTY || !stdout.isTTY) return false`, so it hung every non-TTY caller — was already deleted upstream with #671. This PR pins the invariant with a test: exactly one definition, in `shared.ts`, with its guard intact. A second copy cannot come back unnoticed.

## 3 — The keyless closing line stopped over-promising

`init` ended every run with "the agent is live in your app", including runs where no model key resolved — the one case where nothing happens.

| Credential | Closing line |
| --- | --- |
| keyless (`rung === "none"`) | `Then start your dev server — the agent is live once you add a model key.` |
| any key | `Then start your dev server — the agent is live in your app.` (unchanged) |

## 4 — Pages-router honesty + a monorepo hint

**(a)** A pages-only Next host was told to `import { VendoRoot } …` into `app/layout.tsx` and wrap `{children}` — a file it does not have. Both the manual paste and the agent tail now name the file it does have:

```
In pages/_app.tsx:
  … then wrap: <VendoRoot><Component {...pageProps} /></VendoRoot>
  (vendo/vendo-root.tsx mounts <VendoOverlay />, the visible launcher + panel)
```

The generated `vendo-root.tsx` is a client component, so it mounts in `_app.tsx` unchanged.

Two notes on scope:
- **Where the route segment is scaffolded is untouched.** Handing a pages-only host an `app/api/vendo/[...vendo]` segment is valid in Next as long as both routers share one base, and that placement is a deliberately deferred design question (see `appDirectory`'s docstring). Only the instruction text changed.
- **The client root is keyed on whether `app/layout.tsx` exists**, not on `detectRouter` — the scaffold creates `app/` mid-run, so a `detectRouter`-based answer flipped between the manual paste (computed pre-write) and the agent tail (post-write). Caught by the test.

**(b)** `detectFramework` reads only the target dir's `package.json`, so an interactive `vendo init` at a monorepo root found no `next`/`express` and silently scaffolded the runtime-neutral `custom` target one level too high. It now names the candidates:

```
warning: no next or express dependency here, but apps/web look like hosts — did you mean apps/web?
Re-run there (vendo init apps/web) or pass --framework to scaffold this directory anyway.
```

Candidates come from `apps/*` and `packages/*` whose own `package.json` declares `next` or `express` — a hint that names a candidate, not a workspace-glob resolver. Non-interactive runs already errored with the exact flag and are unchanged; the hint is advisory and never changes the exit code.

## Proof

New file `packages/vendo/src/cli/onboarding-safety.test.ts` (11 tests, no existing test file touched):

- gitignore warning present / absent / silent-outside-git, plus the `--cloud-key` path end to end (warning surfaced, key still written)
- exactly one `askYesNo` definition, in `shared.ts`, guard string present
- keyless vs keyed closing line, both exact strings
- pages-only host: output contains `pages/_app.tsx` and the `<Component {...pageProps} />` wrap, and does NOT contain `app/layout.tsx`
- monorepo root: `workspaceHostCandidates` picks `apps/web` and skips `packages/ui`; the interactive run emits the hint; the non-interactive run keeps its exact-flag error and emits no hint

## Gate

`pnpm build` ✅ · `pnpm typecheck` ✅ (41 tasks) · `pnpm lint` ✅ (dependency-guard + portability-gate + eslint) · `pnpm test` — 1499 tests, 1482 pass, 2 fail

The 2 failures are `src/dev-creds/model.test.ts` (`@ai-sdk/anthropic` module resolution under `withoutGlobalPaths`) and are **pre-existing on `origin/main`** — verified by stashing this branch's changes and re-running the same file: identical 2 failed / 23 passed.

No UI surface in this change (CLI strings only), so no screenshots apply.

---
## Summary by cubic
Make `vendo init` safer and more honest. Adds git warnings for `.env.local` secrets, fixes a non‑TTY prompt hang, corrects Next pages‑router guidance, improves monorepo hints, and clarifies the closing “live” line.

- **Bug Fixes**
  - After writing `VENDO_API_KEY`, warn if `.env.local` would commit (covers tracked files and symlinks); stay quiet outside git; on repo/git failures say “git could not answer” with the reason.
  - Ensure exactly one `askYesNo` with a non‑TTY guard; tests pin this invariant.
  - Closing line now has three states: usable key → “live in your app”; scaffolded this run with no key → “live once you add a model key”; re‑run over an existing composition with no resolved key → “live only if your composition passes its own model.”
  - Pages‑only Next hosts get instructions for `pages/_app.tsx`, wrapping `<Component {...pageProps} />`; agent tail updated.
  - At monorepo roots, suggest likely hosts (e.g., `apps/web`); rerun hint resolves from the caller’s cwd, prints `.` when already there, and is POSIX single‑quoted; non‑interactive behavior unchanged.

- **Refactors**
  - Simplified git probe flow (`check-ignore` first, index check only if needed), unified closing‑line log, streamlined framework fall‑through, and inlined small helpers; consolidated tests. No behavior change.

<sup>Written for commit e4e705653eb78012df09bbd4d63f35d516cbb9b2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/689?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/onboarding-cli-safety-and-honesty.md
+++ b/.changeset/onboarding-cli-safety-and-honesty.md
@@ -1,0 +1,23 @@
+---
+"@vendoai/vendo": patch
+---
+
+Onboarding safety and honesty: four fixes to the first `vendo init`.
+
+- **A secret written into a committed file now says so.** `vendo login` and
+  `vendo init --cloud-key` land `VENDO_API_KEY` in `.env.local`. When
+  `git check-ignore` says that file is not ignored, both print one line naming
+  the fix. The write is never blocked — the key is already minted.
+- **The keyless closing line stopped over-promising.** A run that resolved no
+  model key now ends "the agent is live once you add a model key" instead of
+  claiming it is live in your app.
+- **A pages-only Next host gets instructions that work.** The manual wiring
+  paste and the agent tail named `app/layout.tsx`, a file such a host does not
+  have. They now name `pages/_app.tsx` and wrap `<Component {...pageProps} />`
+  (the generated `vendo/vendo-root.tsx` is a client component, so it mounts
+  there unchanged). Where the API route segment is scaffolded is unchanged.
+- **An interactive init at a monorepo root names the real host.** Detection
+  finds no `next`/`express` at a workspace root and falls through to the
+  runtime-neutral `custom` scaffold — silently one level too high. It now names
+  the workspace packages that do look like hosts ("did you mean apps/web?").
+  Non-interactive runs already errored with the exact flag; unchanged.
--- a/.changeset/onboarding-cli-safety-and-honesty.md
+++ b/.changeset/onboarding-cli-safety-and-honesty.md
@@ -9,14 +9,18 @@ Onboarding safety and honesty: four fixes to the first `vendo init`.
   line about whether git will commit it, with the remediation that actually
   works: `git rm --cached` when the file is already tracked (where .gitignore
   cannot help), the .gitignore line when it is untracked and unignored, and an
-  explicit "git could not answer" when a live repo errors. Silent when the file
-  is ignored, and when there is no repository or no git at all. The write is
-  never blocked — the key is already minted.
-- **The closing line stopped over-promising.** It claimed "the agent is live in
-  your app" whenever a rung resolved — including a malformed `VENDO_API_KEY` or
-  `VENDO_DEV_CREDENTIAL=vendo-cloud` with no key, both of which resolve a rung
-  and cannot serve a turn. It now says "the agent is live once you add a model
-  key" unless the credential is actually usable.
+  explicit "git could not answer" when a live repo errors. Symlinks are resolved
+  first, so a gitignored `.env.local` pointing at a tracked file is judged by
+  the file the write really lands in. Silent when the file is ignored, and when
+  there is no working tree or no git at all. The write is never blocked — the
+  key is already minted.
+- **The closing line stopped guessing in both directions.** It claimed "the
+  agent is live in your app" whenever a rung resolved — including a malformed
+  `VENDO_API_KEY` or `VENDO_DEV_CREDENTIAL=vendo-cloud` with no key, neither of
+  which can serve a turn. Now: a usable credential says live; a composition
+  scaffolded this run with no key says "live once you add a model key"; and a
+  re-run over a composition Vendo did not write states the condition, because
+  that composition may pass its own `model` and nothing here can see it.
 - **A pages-only Next host gets instructions that work.** The manual wiring
   paste and the agent tail named `app/layout.tsx`, a file such a host does not
   have. They now name `pages/_app.tsx` and wrap `<Component {...pageProps} />`
@@ -26,6 +30,6 @@ Onboarding safety and honesty: four fixes to the first `vendo init`.
   finds no `next`/`express` at a workspace root and falls through to the
   runtime-neutral `custom` scaffold — silently one level too high. It now names
   the workspace packages that do look like hosts ("did you mean apps/web?") and
-  suggests a path that resolves from the caller's own cwd, quoted when it
-  contains a space. Non-interactive runs already errored with the exact flag;
-  unchanged.
+  suggests a path that resolves from the caller's own cwd, single-quoted when the
+  shell would otherwise mangle it. Non-interactive runs already errored with the
+  exact flag; unchanged.
--- a/.changeset/onboarding-cli-safety-and-honesty.md
+++ b/.changeset/onboarding-cli-safety-and-honesty.md
@@ -5,12 +5,18 @@
 Onboarding safety and honesty: four fixes to the first `vendo init`.
 
 - **A secret written into a committed file now says so.** `vendo login` and
-  `vendo init --cloud-key` land `VENDO_API_KEY` in `.env.local`. When
-  `git check-ignore` says that file is not ignored, both print one line naming
-  the fix. The write is never blocked — the key is already minted.
-- **The keyless closing line stopped over-promising.** A run that resolved no
-  model key now ends "the agent is live once you add a model key" instead of
-  claiming it is live in your app.
+  `vendo init --cloud-key` land `VENDO_API_KEY` in `.env.local`, and now say one
+  line about whether git will commit it, with the remediation that actually
+  works: `git rm --cached` when the file is already tracked (where .gitignore
+  cannot help), the .gitignore line when it is untracked and unignored, and an
+  explicit "git could not answer" when a live repo errors. Silent when the file
+  is ignored, and when there is no repository or no git at all. The write is
+  never blocked — the key is already minted.
+- **The closing line stopped over-promising.** It claimed "the agent is live in
+  your app" whenever a rung resolved — including a malformed `VENDO_API_KEY` or
+  `VENDO_DEV_CREDENTIAL=vendo-cloud` with no key, both of which resolve a rung
+  and cannot serve a turn. It now says "the agent is live once you add a model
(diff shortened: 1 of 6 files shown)


yousefh409:
## Review round (commit 302bf58) — all five findings were real

**1. The closing line was keyed on the rung, not on a usable credential.**
`resolveDevCredential` only checks that `VENDO_API_KEY` is non-blank, and `VENDO_DEV_CREDENTIAL=vendo-cloud` pins the rung with no key at all — so a malformed key printed `VENDO_API_KEY is set but not usable` and `the agent is live in your app` in the same run. The line now keys on a credential that can actually serve a turn: an env key, or a Cloud rung the cloud step validated (which includes a key minted mid-run).

**2. The rerun hint suggested a path relative to the wrong directory.**
`vendo init monorepo` from `/work` suggested `vendo init apps/web`, which resolves to `/work/apps/web`. The suggestion is now relative to the caller's cwd while it stays inside it, absolute otherwise, and quoted when it contains a space.

**3. Every git failure except exit 1 was silently treated as "ignored".**
A live repo that errors — dubious ownership, unreadable config, corrupt index — swallowed the warning entirely. The failure mode now decides: no git and no repository anywhere stay silent (nothing to leak into), but a repo that cannot answer says so and names git's own reason:

```
warning: .env.local holds a secret and git could not say whether it is ignored (fatal: bad config line 1 in file .git/config) — check it yourself before you commit.
```

**4. An already-tracked `.env.local` got advice that cannot work.**
Verified against real git: `check-ignore` reports a tracked file as **not ignored even when a pattern matches it**, so the tracked case was reachable and was told to "add `.env.local` to .gitignore" — which does not untrack it. `git ls-files --error-unmatch` now runs first:

```
warning: .env.local holds a secret and is TRACKED by git — .gitignore alone will NOT stop the next commit: run `git rm --cached .env.local`, then add `.env.local` to .gitignore.
```

**5. The "not ignored" test inherited the developer's own git excludes.**
Anyone with `.env.local` in `~/.config/git/ignore` would have seen this flip. The git tests now isolate `HOME`, `XDG_CONFIG_HOME`, `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM`. Verified load-bearing, not cargo-culted: a hostile excludes file makes `check-ignore` return "ignored" without the isolation and "not ignored" with it.

Tests: 11 → 17. New cases cover the unusable-key and pinned-rung closing lines, a mid-run mint staying "live", the tracked file, a repo whose config git cannot parse, and the space-quoted rerun path.

## Gate (re-run after the review round)

`pnpm build` ✅ · `pnpm typecheck` ✅ (41 tasks) · `pnpm lint` ✅ · `@vendoai/vendo` tests: **121 files pass, 1 fails** — the pre-existing `src/dev-creds/model.test.ts` pair, confirmed identical on clean `origin/main`.

One note on the full `pnpm test` run: four other packages (`demo-accounting`, `fixtures/mcp-e2e`, `fixtures/integration`, `fixtures/redteam`) failed when turbo ran them concurrently after the cache invalidated, plus one `init.test.ts` case that shells out to a real `npm install`. Each was re-run standalone with this branch applied and **all pass** (`mcp-e2e` 19/19, `integration` 35/35, `redteam` 21/21, `init.test.ts` 85/85) — the failures were contention on live-service and network-dependent suites, not regressions. `demo-accounting`'s two `login-e2e` failures (live Supabase, 401) reproduce on clean `origin/main`.

yousefh409:
## Review round 2 (commit 1a4e3d5) — all five findings were real

**1. `modelReady` ignored host-passed model objects.** My round-1 fix traded a false positive for a false negative: `createVendo({ model })` needs no env key, so a keyless re-run told a working install it was not live. Rather than add machinery to detect a model object, the line now has three honest states:

| Situation | Line |
| --- | --- |
| usable credential (env key, or a Cloud key the cloud step validated) | `the agent is live in your app.` |
| composition scaffolded **this run**, no key | `the agent is live once you add a model key.` |
| re-run over a composition Vendo did not write, no key | `no model key resolved here, so the agent is live only if your composition passes its own model.` |

The middle case is definitive because the composition init just generated passes no model (asserted by the existing suite). The third states the condition instead of guessing, which is the honest answer when nothing in reach can see a host's model object.

**2. `relative()` returns `""` when the target IS the cwd.** `cd repo/apps/web && vendo init ../..` fell through to an absolute path; it now suggests `vendo init .`.

**3. `JSON.stringify` is not shell escaping.** Correct — inside double quotes, a directory named `$(…)` is command-substituted by the shell that pastes the suggestion, so the "safe" quoting would have *executed* it. Replaced with POSIX single-quote escaping (`'` → `'\''`), applied only to paths that need it. Test asserts a path containing `$(printf SUBSTITUTED)` comes out single-quoted and never double-quoted.

**4. The probes only saw the lexical `.env.local`.** Real: with the link itself gitignored and pointing at a tracked `config/dev.env`, the write landed in a tracked file and nothing was said. The real path is now resolved before the probes, which run in the repo holding the secret.

Worth flagging: resolving one side was not enough. Every macOS temp dir lives under `/var` → `/private/var`, so `relative(root, realpath(file))` turned the file's own name into a `../../../../../../private/var/...` chain in the *normal* case. Both sides are resolved now — the existing tests caught it immediately.

**5. Locale-dependent stderr matching.** Real. No decision reads stderr any more, taking the exit-code approach you preferred and one step further:

- a `.git` entry found by walking up decides whether a working tree exists (no git invocation, so nothing to translate — and it distinguishes "no repository" from "broken repository", which `rev-parse` alone cannot: a corrupt config makes `rev-parse` fail too, which would have silently undone the round-1 fix)
- exit 0 / 1 from `check-ignore` decide ignored / not-ignored
- any other exit is "git could not answer", with git's own words quoted **as display, never as logic** (and `ENOENT` reads "git is not installed")

Tests: 17 → 20.

## Gate

`pnpm build` ✅ · `pnpm typecheck` ✅ (41 tasks) · `pnpm lint` ✅ (dependency-guard + portability-gate + eslint) · `@vendoai/vendo`: **121 files pass, 1 fails** — the pre-existing `src/dev-creds/model.test.ts` pair, confirmed identical on clean `origin/main`. `onboarding-safety.test.ts` 20/20.

PR #690, opened 2026-07-29
docs(quickstart): match the shipped surfaces, add drift-parity test

yousefh409:
`docs/quickstart.md` is the first code a Vendo host ever pastes, and it had drifted from every surface it documents. This fixes the doc and adds a test so it cannot drift silently again.

## The rot, and what it is now

**1. The registry example imported a package a host cannot resolve.**
`import type { ComponentRegistry } from "@vendoai/core"` — `@vendoai/core` is transitive, so under pnpm strict linking a host that installed only `@vendoai/vendo` gets TS2307. The CLI's own scaffold imports it from `@vendoai/vendo` for exactly this reason ([`init-scaffolds.ts:46-52`](https://github.com/runvendo/vendo/blob/main/packages/vendo/src/cli/init-scaffolds.ts#L46-L52)). The doc now shows the scaffold's import and says why in one paragraph.

**2. The primary route example pinned a deprecated key.**
It pinned `model: anthropic(...)` and mentioned `paint:`; `CreateVendoConfig` marks both `@deprecated`, superseded by `models: ModelsConfig`. The primary example is now the init scaffold **verbatim** (no model line — the composed default resolves a key from the environment), with the current pin surface beside it:

```ts
models: { agent: "claude-sonnet-4-6" },
```

The `@ai-sdk/anthropic` install requirement was also wrong: `@ai-sdk/anthropic` is a real dependency of `@vendoai/vendo`, and `importHostModule` falls back to vendo's own copy for `@ai-sdk/*` specifiers. So the string form needs no host install on the Anthropic and Cloud rungs; only a BYO model **object** does. The doc now says that.

**3. The config-surface listing was wrong in three ways at once.**
It omitted nine real keys — including `knowledge` and `apps`, which its own prose referenced 100+ lines later — pulled types from six transitive packages a host cannot import, and dropped `guardedTools` from the `Vendo` interface. It is now regenerated from the actual `CreateVendoConfig` (all 30 keys, `model`/`paint` marked deprecated), with an import block that names only `@vendoai/vendo`, `@vendoai/vendo/server`, and `ai`.

**4. The intro described an init that no longer exists.**
It said init "writes two code files", "prints the one line that is yours to paste", and "never edits code you wrote". Init writes three code files plus the server-action map, makes one bounded `app/layout.tsx` edit itself, and the paste is the *degraded fallback* — which the doc's own §322 already admitted. The intro now matches `init.ts`, including the `.vendo/` artifacts and the two `package.json` hooks.

**5. `vendo try` was buried at line 61.** The zero-install look is now the first section.

Also: `--framework <next|express>` → `<next|express|custom>` in the non-interactive section, and the doc's client-mount section is now the generated `vendo/vendo-root.tsx` instead of a hand-rolled `Root` that nothing writes.

## The parity test (the load-bearing part)

New `packages/vendo/src/cli/quickstart.docs.test.ts`, patterned on `doctor-codes.docs.test.ts`. Seven checks, all mechanical:

| Check | Fails when |
| --- | --- |
| registry block vs `registrySource("tsx")` | the scaffold's import specifier changes |
| route block vs `routeSource(...)` | any load-bearing scaffold line changes (exact line-for-line) |
| client-mount block vs `vendoRootWrapperSource(...)` | same, for the wrapper |
| config listing vs `CreateVendoConfig` | a key is added, removed, reordered, or its `@deprecated` mark moves |
| `Vendo` listing vs the `Vendo` interface | a member is added or removed |
| every `@vendoai/*` specifier in a code block | the doc tells a host to import a transitive package |
| every composition example | a `model:` or `paint:` pin comes back |

Plus the doc's import block is **compiled inside the test file**, so a renamed or removed umbrella export breaks `pnpm typecheck` rather than the reader.

### Mutation check

Each check was proven to bite, then reverted:

| Mutation | Result |
| --- | --- |
| `routeSource`: `policy: {}` → `policy: "cautious"` | ❌ route block test fails, diffing the exact line |
| `registrySource`: import from `@vendoai/core` | ❌ two tests fail (parity + transitive-import) |
| `server.ts`: add a `brandNewKey?: string` config key | ❌ config-key test fails, 30 vs 31 |
| `server.ts`: delete `guardedTools` from `Vendo` | ❌ `Vendo` member test fails, 10 vs 9 |
| doc: drop the `@deprecated` mark on `model` | ❌ config-key test fails on the flag |
| doc: reintroduce `model: anthropic(...)` | ❌ deprecated-pin test fails, naming the line |

All reverted; `git status` clean of mutations before commit.

## Two CLI copy fixes

- `--yes` claimed only "skip the cloud-login offer". It also accepts the detected auth preset, skips AI polish and the theme review, and swaps the interactive success screen for the agent tail.
- `--framework` help listed `next, express`; `custom` has been valid all along (`INIT_FRAMEWORK_VALUES`).
- `vendo login`'s transient-failure catch printed the raw `error.message` and nothing else, so the reader assumed the ceremony was lost and started over — abandoning an approval that would still have landed. It now names the surviving pairing code and says a re-run resumes the same request, gated on a claim that actually survived and has not expired (every terminal outcome already deletes it, so the line never lies).

## Gate

- `pnpm build` ✅
- `pnpm typecheck` ✅ (41/41)
- `pnpm lint` ✅ (dependency-guard, portability-gate, eslint — force-run, not cached)
- `pnpm test` — 51/53 tasks green. Two pre-existing environmental failures, **identical on `origin/main` in this worktree**, unrelated to this diff:
  - `@vendoai/vendo` — 2 tests in `src/dev-creds/model.test.ts` ("provider module resolution"): `@ai-sdk/anthropic` does not resolve from vendo's own module context in this worktree's node_modules layout. Measured failing at baseline with the branch stashed. The other **1478 tests pass**, including the 7 new ones.
  - `demo-accounting` — `src/vendo/login-e2e.test.ts` needs a live Supabase/GoTrue; it passes when the container is up and fails when it is not, independent of this branch (which touches nothing demo-accounting runs).

No test files were edited — only the new one added.

---
## Summary by cubic
Brings the quickstart doc back in sync with the shipped scaffolds and types, compiles the published config surface to prevent drift, and makes `vendo init --yes` non-blocking by queueing loosening proposals instead of prompting.

- **Bug Fixes**
  - Quickstart examples now match `init` scaffolds verbatim: registry imports `ComponentRegistry` from `@vendoai/vendo`, the route uses `models` (no deprecated `model`/`paint`), and the client mount shows `vendo/vendo-root.tsx`.
  - Config listing rebuilt from the real `CreateVendoConfig`/`Vendo`: all keys included (adds `knowledge`, `apps`, `theme`, `brief`, `connectorApps`, `approvals`, `profileDir`, `profile`, `fetch`), `guardedTools` present, and closed shapes for `mcp.remoteAs`/`federation` and `apps.pipeline`.
  - Intro matches current `vendo init`: consent prompts listed (auth preset, Cloud offer, AI polish, loosenings review, theme value, zod bump, star), one bounded layout edit, full file set, and `createVendo({})` boots (no-arg throws).
  - CLI help: `--yes` describes all effects and now never blocks on the loosening review; unattended runs queue loosenings as pending and print `vendo sync --review` with a reason. `--framework` lists `custom`. `vendo login` prints the surviving pairing code and resume guidance on transient failures.

- **Refactors**
  - Added `packages/vendo/src/cli/quickstart.docs.test.ts` and compiled fixture `quickstart-config-surface.docs-check.ts`: the doc’s config block is typechecked against real types; the test asserts the doc stays byte-identical, forbids transitive `@vendoai/*` imports and deprecated `model`/`paint` pins, and anchors import rewrites to declarations. New `tsconfig.docs-check.json` typechecks fixtures without emitting; `typecheck` runs both programs.
  - Simplified the parity gate: replaced the interface parser with regex scans, merged per-interface key-set tests, dropped a redundant “must import from <specifier>” check (byte identity over anchored import lines already proves it), read markdown once per test file, and folded the queued-loosening message assertion into the `--yes` test. Minor `tsconfig.docs-check.json` cleanup.
  - Added `packages/vendo/src/cli/init-judgment.unattended.test.ts` to pin that `--yes` and non-TTY runs queue loosenings and never forward a confirm seam.
  - `@vendoai/vendo` re-exports `ServerActionHandler` so the copied config block typechecks.

<sup>Written for commit 4693681ccf0b23229e356f8027364521c1b6518c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/690?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/quickstart-parity-and-cli-copy.md
+++ b/.changeset/quickstart-parity-and-cli-copy.md
@@ -1,0 +1,19 @@
+---
+"@vendoai/vendo": patch
+---
+
+Three CLI help and error lines now say what the code actually does.
+
+`--yes` claimed only "skip the cloud-login offer". It also accepts the detected
+auth preset, skips the AI polish pass and the theme review, and swaps the
+interactive success screen for the agent tail — an agent reading the old line
+could not predict any of that. `--framework` listed `next, express` while
+`custom` (the runtime-neutral scaffold for Workers, Bun, Deno, Hono, and Lambda
+adapters) has been accepted all along.
+
+When `vendo login` dies on a transient failure — network, DNS, a killed fetch —
+it printed the raw error and nothing else, so the reader assumed the ceremony
+was lost and started over, abandoning an approval that would still have landed.
+It now names the surviving pairing code and says that re-running `vendo login`
+resumes the same request. The line appears only when a resume can actually
+succeed: every terminal outcome already deletes the claim.
--- a/.changeset/quickstart-parity-and-cli-copy.md
+++ b/.changeset/quickstart-parity-and-cli-copy.md
@@ -2,7 +2,19 @@
 "@vendoai/vendo": patch
 ---
 
-Three CLI help and error lines now say what the code actually does.
+`vendo init --yes` no longer blocks on the loosening review, and three CLI help
+and error lines now say what the code actually does.
+
+`--yes` promises every question is already answered. It kept that promise for
+the AI-polish consent and broke it one step later: with `--ai-polish` granting
+consent, a run in a terminal reached the aggregated loosening review and waited
+for a human the moment the judgment pass proposed waking a disabled tool or
+lowering a risk grade — so `vendo init --yes --ai-polish` could hang in CI or
+under an agent. Unattended runs now queue loosenings instead: held as `pending`,
+nothing applied, printed with `vendo sync --review`. Auto-applying was never an
+option — risk is not lowered without a human — and no `confirm` seam is handed
+to the pass at all when the run is unattended, so nothing downstream can block
+either.
 
 `--yes` claimed only "skip the cloud-login offer". It also accepts the detected
 auth preset, skips the AI polish pass and the theme review, and swaps the
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -17,8 +17,14 @@ npx vendo init
 ```
 
 The `vendoai` package is a thin alias. The scoped package is the canonical
-install. `vendo init` asks nothing on the happy path. It writes the whole
-server side and mounts the visible surface for you:
+install. `vendo init` asks for no data — no interview, no keys to paste, no
+per-diff approvals. What an interactive run does ask are consent questions you
+answer with Enter or y/n: the detected auth preset, the Vendo Cloud starter-key
+offer (only when no provider key is set), consent for the AI polish pass, a
+review of any theme slot the extractor was unsure about, a `zod` bump when your
+pin is below 3.25, and a closing "star the repo?". `--yes` and non-interactive
+runs answer all of them silently. Init writes the whole server side and mounts
+the visible surface for you:
 
 - the catch-all route `app/api/vendo/[...vendo]/route.ts` holding the entire
   `createVendo` composition (on Express or any other Web-standard runtime,
@@ -45,10 +51,8 @@ like everything else. When the layout can't be edited unambiguously, init
 degrades to printing the paste lines instead — see
 [the client mount](#vendovendo-roottsx--the-client-mount).
 
-When it detects your auth provider in package.json it asks once to confirm the
-preset (consent-style, Enter accepts; `--yes` and non-interactive runs accept
-it silently). Then start your dev server — the agent is live in your app — and
-run `npx vendo doctor` to verify everything with one real model turn.
+Then start your dev server — the agent is live in your app — and run
+`npx vendo doctor` to verify everything with one real model turn.
 
 ## Non-interactive runs (agents)
 
@@ -162,8 +166,10 @@ see [actAs preset recipes](./act-as-presets.md) for the full list, the
 hosts without a shipped preset.
 
 `models` and `catalog` are the only other keys most hosts touch on day one.
-Every key is optional — a bare `createVendo()` legitimately boots, with an
-env-resolved model, anonymous ephemeral sessions, and PGlite persistence.
+Every key is optional — `createVendo({})` legitimately boots, with an
+env-resolved model, anonymous ephemeral sessions, and PGlite persistence. (The
+config object itself is required: `createVendo()` with no argument does not
+typecheck and throws at runtime.)
 The `policy: {}` line activates the `.vendo/policy.json` file init wrote, so
 the scaffolded default posture is: destructive tools ask, reads run. Remove
 the `policy` key entirely and every call auto-runs (audited, with the
@@ -428,10 +434,11 @@ absent (see [Model keys](#model-keys)), and with neither `auth` nor
 validation error at compose time. Pick the preset or hand-wire the three
 seams, never both.
 
-Every type below is importable by a host: the umbrella's root entry re-exports
-the contract types, and `@vendoai/vendo/server` carries the composition ones.
-(`ServerActionHandler` is the only exception — the generated `vendo-actions.ts`
-map is the surface you use, never the type.)
+`createVendo(config: CreateVendoConfig): Vendo`, in full. Every type is
+importable by a host — the umbrella's root entry re-exports the contract types
+and `@vendoai/vendo/server` carries the composition ones — and this block is
+compiled against the real `CreateVendoConfig` in
+`packages/vendo/src/cli/quickstart-config-surface.ts`, so it cannot drift:
 
 ```ts
 import type {
@@ -441,10 +448,10 @@ import type {
   PolicyFile, Principal, RunId, SandboxAdapter, SecretsProvider, ToolRegistry,
   VendoAgent, VendoGuard, VendoStore, VendoTheme,
 } from "@vendoai/vendo";
-import type { ConnectionsService, HostAuthPreset, ModelsConfig } from "@vendoai/vendo/server";
+import type {
+  ConnectionsService, HostAuthPreset, ModelsConfig, ServerActionHandler,
+} from "@vendoai/vendo/server";
 import type { LanguageModel } from "ai";
-
-export function createVendo(config: CreateVendoConfig): Vendo;
 
 export interface CreateVendoConfig {
   /** @deprecated superseded by `models.agent`. */
@@ -481,7 +488,11 @@ export interface CreateVendoConfig {
     policy?: PolicyFile;
     designRules?: string;
   };
-  mcp?: boolean | { baseUrl?: string; remoteAs?: object; federation?: object };
+  mcp?: boolean | {            // the door; `baseUrl` is its PUBLIC base URL
+    baseUrl?: string;
+    remoteAs?: { issuer: string; jwksUri?: string; audience: string };
+    federation?: { secret: string };
+  };
   oauth?: HostOAuthAdapter;   // escape hatch; required when `mcp` is true and `auth` is absent
   agent?: {
     instructions?: string;
@@ -498,7 +509,14 @@ export interface CreateVendoConfig {
   apps?: {
     experimentalServedApps?: boolean;
     experimentalMachines?: boolean;
-    pipeline?: Record<string, boolean>; // generation-pipeline opt-ins
+    pipeline?: {              // generation-pipeline knobs, measured before default-on
+      structuredRepair?: boolean;
+      regionParallel?: boolean;
+      endPass?: boolean;
+      exemplarContract?: boolean;
+      smokeRender?: boolean;
+      rebind?: boolean;
+    };
     designRules?: string;
   };
 }
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -1,62 +1,6 @@
 # Vendo quickstart
 
-Install the default composition and run setup:
-
-```bash
-npm install @vendoai/vendo
-npx vendo init
-```
-
-The `vendoai` package is a thin alias. The scoped package is the canonical
-install. `vendo init` asks nothing on the happy path. It writes two code
-files — an empty `vendo/registry.tsx` and the catch-all handler wired to
-it — plus two package.json script hooks, extracts your tools and brand
-theme into `.vendo/`, resolves a model key, and prints the one line that is
-yours to paste: the `<VendoRoot components={registry}>` wrap in your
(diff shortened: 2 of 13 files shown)


yousefh409:
## Codex review round 1 — all six findings fixed (`eabcbb0`)

Every finding was real. Verdicts and fixes:

**1. "asks nothing on the happy path" — CONFIRMED.** A fresh keyless interactive run prompts for the Cloud starter key ([`cloud-init.ts:165`](https://github.com/runvendo/vendo/blob/main/packages/vendo/src/cli/cloud-init.ts#L165)), AI-polish consent, any uncertain theme slot, a `zod` bump below 3.25, and the closing star ask. The claim was inherited from the old doc and I preserved it. The intro now says what is actually true — init asks for no *data* (no interview, no keys to paste, no per-diff approvals) — and names every consent question plus the fact that `--yes` answers them all.

**2. `createVendo()` does not boot — CONFIRMED.** `config` is a required parameter and the body reads `config.auth` at line 1206, so TS rejects the call and JS throws. Fixed to `createVendo({})`, **verified booting** with a throwaway test (`typeof vendo.handler === "function"`), plus an explicit note that the no-argument form does not typecheck. (Worth flagging: [`09-vendo.md:214`](https://github.com/runvendo/vendo/blob/main/docs/archive/contracts/09-vendo.md#L214) still says "bare `createVendo()` legitimately boots" — that is where the claim came from. Archived contract, left alone.)

**3. `mcp.remoteAs`/`federation` and `apps.pipeline` — CONFIRMED.** Both now show the real closed shapes: `remoteAs?: { issuer: string; jwksUri?: string; audience: string }`, `federation?: { secret: string }`, and `pipeline` expanded to its six named boolean flags.

**4. `ServerActionHandler` unimported → TS2304 — CONFIRMED.** My previous "the only exception" caveat papered over a block that would not compile. It is the value type of a documented config key, so `server.ts` now re-exports it beside `CatalogFile`/`ExtractedTool`/`OverridesFile` (one identifier added to an existing export line — a host must be able to name it without a direct `@vendoai/actions` dependency), and the listing imports it.

**5. + 6. Shallow comparison and a hand-written import list — CONFIRMED, and worse than reported.** `tsconfig.json` excludes `src/**/*.test.ts`, so **nothing in the test file was ever typechecked** — my "the import block is compiled in the test" claim was empty. Finding 3's inaccuracies passing while green is exactly the blind spot.

Rather than deepen the string comparison, the fix is a compiled fixture: **`packages/vendo/src/cli/quickstart-config-surface.ts`** holds the doc's block *verbatim* (the only rewrite: `@vendoai/vendo` → `../index.js`, since a package cannot resolve itself by name) and asserts invariant type identity against the real types:

```ts
type Identical<A, B> = (<T>() => T extends A ? 1 : 2) extends (<T>() => T extends B ? 1 : 2) ? true : false;
export type ConfigSurfaceMatchesTheDoc = Assert<Identical<CreateVendoConfig, RealCreateVendoConfig>>;
export type VendoSurfaceMatchesTheDoc = Assert<Identical<Vendo, RealVendo>>;
```

That file is **not** excluded from tsconfig, so nested members, their types, and their optionality are all checked by `pnpm typecheck`, and the imports being compiled are the doc's own text. Invariant identity (not mutual assignability) is deliberate — mutual assignability would have let `remoteAs?: object` pass in one direction. The test now asserts doc and fixture stay byte-identical, and the hand-maintained import list is deleted.

## Mutation checks (all reverted)

| Mutation | Result |
| --- | --- |
| fixture: `remoteAs?: {...}` → `object` (finding 3's exact drift) | typecheck fails — TS2344 `Type 'false' does not satisfy the constraint 'true'` |
| fixture: drop nested `pipeline.smokeRender` | typecheck fails — TS2344 |
| doc only: rename `VendoTheme` → `VendoThemeTypo` (finding 6's scenario) | byte-identity test fails |
| doc **and** fixture: same rename | typecheck fails — TS2724 "has no exported member named 'VendoThemeTypo'" |

Round-1 mutations (scaffold line, core import, added config key, deleted `guardedTools`, dropped `@deprecated`, reintroduced `model:`) all still bite.

## Gate

- `pnpm build` green · `pnpm typecheck` green (41/41) · `pnpm lint` green (dependency-guard, portability-gate, eslint)
- `pnpm test` — 51/53. **Same two pre-existing environmental failures as `main` in this worktree**, unchanged by this batch:
  - `@vendoai/vendo` — 2 tests in `src/dev-creds/model.test.ts` (`@ai-sdk/anthropic` does not resolve from vendo's own module context here). **1479 pass**, including all 8 parity tests.
  - `demo-accounting` — `src/vendo/login-e2e.test.ts` needs a live Supabase/GoTrue.

One note on measurement: an intermediate run showed ~10 extra failures at 30–40s each. Machine load average was 50 at the time; re-run clean it is back to the 2 above. Load-induced timeouts, not regressions.

yousefh409:
## Codex review round 2 — all five findings fixed (`35f0c9b`)

Every finding was real, including a genuine CLI bug. Verdicts:

### 1. Consent inventory omitted the loosening review — CONFIRMED

`reviewLoosenings()` shows a diff and asks `Apply N loosenings across M tools` before anything is applied. It is now in the list, described as what it is: one aggregated question covering every proposed wake or risk downgrade.

### 2. `--yes` blocked on the loosening review — CONFIRMED, real bug, fixed

`init-judgment.ts` keyed the review on TTY alone:

```ts
loosenings: interactive ? "review" : "queue",
```

With `--ai-polish` supplying consent, `options.consent === true` skips the early return, so a TTY run with `--yes` reached `"review"` and **waited for a human** the first time the pass proposed waking a disabled tool or lowering a risk grade. `vendo init --yes --ai-polish` could hang in CI or under an agent — the exact promise `--yes` makes is that it does not.

Fixed as directed — unattended runs **queue**, never auto-apply:

```ts
const attended = interactive && !options.yes;
// loosenings: attended ? "review" : "queue"
// confirm is not forwarded at all when unattended
```

Auto-applying was never on the table (risk is not lowered without a human, per this repo's guard law). Withholding the `confirm` seam entirely is belt-and-braces: no future code path inside the pass can acquire a way to block, whatever the mode.

The pass already prints the count and `vendo sync --review`; init now adds the *why*, so a queued result cannot read as a refusal or a silent apply:

```
  2 loosenings queued — review with `vendo sync --review`
  (held, not applied: this run had no one to ask — re-run `vendo init` in a terminal to review them inline)
```

**New tests** (`init-judgment.unattended.test.ts`, 4 cases): `--yes` in a TTY, non-TTY, the note's presence, and that an attended run *still* reviews inline with the confirm forwarded. Regression proof: restoring `const attended = interactive` fails 2 of the 4.

### 3. "Enter or y/n" misdescribed the theme review — CONFIRMED

`init.ts:238` asks `Replacement value, or Enter to keep:` — a hex colour or font stack, so `y` is invalid there. The doc now separates it out explicitly as the one question that takes a **value**.

While in that section I also found and fixed two related untruths I had left standing: the AI-polish section claimed "risk can be raised but never lowered" (loosenings *can* be applied — after human approval) and that output "lands in the override channel" (judgments land in `.vendo/judgments.json`; `overrides.json` means only "what a person decided"). And "skipped silently in non-interactive runs" is only true *without* `--ai-polish`, which it now says.

### 4. Unanchored import localization — CONFIRMED, and I verified the hole was exploitable

Reproduced the exact attack: a comment carrying `from "@vendoai/vendo";` above an import whose real specifier reads `from "../index.js";`.

- Under the **old** logic: **all 8 tests passed** and `tsc` was clean — a quickstart telling hosts to import from a relative path, fully green.
- Under the **new** logic: fails with `the listing must import from @vendoai/vendo: expected [ '../index.js', …(2) ] to include '@vendoai/vendo'`.

Localization is now anchored to an import declaration's own closing line (`/^\} from "…";$/gm`), the "names the public specifier" assertion reads real declarations only via a comment-skipping `importSpecifiers()`, and the listing may contain **no** relative import at all. The host-installability sweep across every code block now uses the same declaration-only scan, so a specifier quoted in prose can never stand in for one a host pastes.

### 5. Test-only fixture shipped in the tarball — CONFIRMED, excluded

`files: ["dist", …]` ships `dist` wholesale, and build + typecheck shared one tsconfig, so keeping it typechecked meant keeping it emitted. Split:

- renamed to `quickstart-config-surface.docs-check.ts`
- `tsconfig.json` excludes `src/**/*.docs-check.ts` (build)
- new `tsconfig.docs-check.json` (`noEmit`) includes exactly that pattern
- `typecheck` runs both programs

Verified three ways: a fresh emit into a scratch `outDir` contains no `quickstart-config-surface*`; the main program stays clean when the fixture is deliberately broken; and the docs-check program still catches that break.

## Mutation checks (all reverted)

| Mutation | Result |
| --- | --- |
| `attended = interactive` (restore the bug) | 2 of 4 unattended tests fail |
| comment-absorption attack, anchored fixture | anchored specifier assertion fails; **passed** under old logic |
| fixture: `remoteAs?: {…}` → `object` | `tsconfig.docs-check` TS2344; main program stays clean |
| fixture: drop nested `pipeline.smokeRender` | `tsconfig.docs-check` TS2344 |
| `server.ts`: add a config key | runtime key test fails **and** docs-check TS2344 |
| `routeSource`: change the `policy` line | route parity test fails |

## Gate

- `pnpm build` green · `pnpm typecheck` green (41/41, now including the docs-check program) · `pnpm lint` green
- `packages/vendo`: **1483 passed**, 12 of them the parity + unattended tests. The only failures are the 2 pre-existing `src/dev-creds/model.test.ts` provider-resolution tests, which fail identically on `main` in this worktree.
- Full workspace: same known set — `@vendoai/vendo` (those 2) and `demo-accounting`'s `login-e2e.test.ts`, which needs a live Supabase/GoTrue.

PR #691, opened 2026-07-29
demo-creator: stop guessing motion from static screenshots

yousefh409:
## Summary
- The brief's vision prompt asked the model to read `"motion": full|reduced` off a static screenshot — there's no motion signal in a static image, so it was guessing. Across the demo fleet's `theme.json`s it guessed `"reduced"` more than half the time (grafana, numbers, plausible, ramp-licenses, toggletravel, zelty).
- `"reduced"` sets `--vendo-motion-duration: 0ms`, which silences every built-in `@vendoai/ui` animation (`VendoOverlay` open/close, `MorphToast`, etc.) sitewide. Nothing was actually missing — it was switched off by a hallucinated read.
- `motion` is now hardcoded `"full"` and dropped from the vision prompt's schema entirely. Real `prefers-reduced-motion` accessibility support is already handled at runtime by the components themselves, independent of this per-brand guess.

## Test plan
- [x] `vitest run src/demo-creator/` — 464 passed, 1 pre-existing skip (`judge.live.test.ts`)
- [x] `tsc --noEmit` on `bench/`
- [x] Added a regression test asserting `motion` is hardcoded to `"full"` even if a model reply volunteers `"reduced"`
- [ ] Already-shipped demos with `motion: "reduced"` in their `theme.json` are unaffected by this PR (theme.json is fenced from `demo:fix` by design) — separately patched `toggletravel`'s `theme.json` by hand in the demos repo; the other 5 (grafana, numbers, plausible, ramp-licenses, zelty) are untouched and still reduced pending a similar one-off fix if wanted

---
## Summary by cubic
Stop guessing motion from static screenshots by hardcoding `motion: "full"` so demos no longer disable `@vendoai/ui` animations due to bad reads.

- **Bug Fixes**
  - Removed `motion` from the vision prompt schema and guidance; only `density` is read from screenshots.
  - In `parseBriefReply`, set `motion` to `"full"` and update `themeNotes` to reflect it is not derived from screenshots.
  - Added a regression test to keep `motion` at `"full"` even if a model reply includes `"reduced"`.
  - Prevents sitewide animation suppression (`VendoOverlay`, `MorphToast`, etc.) caused by previous `"reduced"` guesses.

<sup>Written for commit 77107510c40153c4a8436402c878f77cca7750c8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/691?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/bench/src/demo-creator/brief.test.ts
+++ b/bench/src/demo-creator/brief.test.ts
@@ -208,11 +208,22 @@ describe("parseBriefReply colours", () => {
     expect(theme.colors.surface).toBe("#FFFFFF");
   });
 
-  it("produces a theme parseDemoTheme accepts, with the model's density and motion", () => {
+  it("produces a theme parseDemoTheme accepts, with the model's density", () => {
     const { theme } = parseBriefReply(reply(), parseOptions());
     expect(() => parseDemoTheme(theme)).not.toThrow();
     expect(theme.colors.accent).toBe("#1E6BFF");
     expect(theme.density).toBe("compact");
+    expect(theme.motion).toBe("full");
+  });
+
+  // A static screenshot carries no motion signal — the vision call is never
+  // asked for one, and even a reply that volunteers "reduced" (e.g. a stale
+  // fixture, or a model that ignores the schema) must not silence every
+  // built-in @vendoai/ui animation sitewide.
+  it("hardcodes motion to full regardless of what the model reply says", () => {
+    const parsed = JSON.parse(reply()) as Record<string, unknown>;
+    parsed["motion"] = "reduced";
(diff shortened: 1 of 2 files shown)


yousefh409:
bench demo-creator was deleted in #831; superseded.

PR #694, opened 2026-07-29
fix(vendo): the model ladder surfaces the resolved model id, so capability params see the real Claude family

### LINKED ISSUE #692: vendoModel()'s lazy wrapper reports modelId "vendo-env", defeating model-params' Claude 5 allowlist
Found while building a prospect demo on the published 0.5.0.

**Symptom**: pin any Claude 5 model through the ladder (`VENDO_MODEL=claude-sonnet-5` with an explicit `ANTHROPIC_API_KEY`) and every generation fails instantly:

```
[vendo] gen pipeline full attempt=0 valid=false issues=model generation failed: `temperature` is deprecated for this model. ms=268
```

**Cause**: `packages/apps/src/model-params.ts` decides sampling support from `model.modelId`, but `vendoModel()` returns `lazyModel(...)` whose `modelId` is the literal `"vendo-env"` (`packages/vendo/src/dev-creds/model.ts` — `lazyModel(controller, "vendo", name ?? "vendo-env")`). `claudeToken("vendo-env")` finds no claude token → `acceptsSamplingParams` returns true (non-Claude branch) → `temperature: 0` is sent → the resolved Claude 5 rung 400s. The b6fdfc106 fix is correct at the call sites but never sees the real model id behind the ladder.

Sonnet 4.6 masks it (accepts temperature), which is why default-ladder hosts don't notice.

**Repro**: vendo-demos host, `ANTHROPIC_API_KEY` set, `VENDO_MODEL=claude-sonnet-5`, any generate-ui beat.

**Fix direction**: the wrapper should either surface the resolved rung's model id, or model-params should resolve through the controller (ask the lazy model, not the wrapper) before deciding sampling params.

yousefh409:
Fixes #692.

## What broke

`vendoModel()` hands back a lazily-resolving model. Its `modelId` was the seam's
own placeholder — the literal `vendo-env` — and it stayed that way forever,
because nothing ever wrote the resolved rung's id back onto the wrapper.

That id is not decoration. `packages/apps/src/model-params.ts` reads
`model.modelId` to decide HOW to make the call: the Claude 5 line removed the
sampling parameters and rejects `temperature` outright. `claudeToken("vendo-env")`
finds no Claude token, so the rule took the non-Claude branch, kept
`temperature: 0`, and the resolved Claude 5 rung answered:

```
[vendo] gen pipeline full attempt=0 valid=false issues=model generation failed: `temperature` is deprecated for this model. ms=268
```

Every create, edit, repair and verify pass, on the first request. Observed live
2026-07-29 on a prospect demo with `VENDO_MODEL=claude-sonnet-5` and an explicit
`ANTHROPIC_API_KEY`. Sonnet 4.6 masks it by still accepting `temperature`, which
is why default-ladder hosts never saw it.

## The fix

The wrapper reports the rung it will actually call.

`modelId` is now a getter over the controller. It answers the resolved model's
own id once resolution has run, and before that the id resolution WILL pick —
which the ladder can answer without waiting, because credential detection is
pure env reading with no network and no disk. `resolveDevCredential` keeps its
awaitable seam and delegates to a new synchronous `detectDevCredential`.

That mattered for the timing: capability params are computed while COMPOSING the
first call, before any resolution has happened. A getter that only became honest
after the first call would still have 400'd on that first call — the only one
that matters here.

Fixing the id rather than stripping parameters inside `doGenerate` was the call
because the id is what everything downstream reasons from. Stripping would put a
second copy of the sampling policy behind the ladder, would silently drop a
caller's explicit `temperature` (which `explicitSamplingParams` deliberately
throws on), and would leave every other reader — error messages, the output cap,
logs — still looking at `vendo-env`.

## What does not change

- The same precedence picks the id (env pin → deprecated pin → configured string
  → verbatim name → per-rung default); only the reporting is new.
- `describeDevCredential` announce lines still print the real id, from the same
  code path as before.
- A rung with no credential keeps the placeholder — there is no real id, and no
  call can succeed anyway.
- Non-ladder callers are untouched: nothing in `model-params.ts` moved. Ladder
  hosts on the default rungs resolve to `claude-sonnet-4-6` / `vendo` / `gpt-5`,
  all of which take sampling, so they get exactly today's `temperature: 0`.

## Tests

Both halves of the chain, each pinned in its own package (`model-params` is
internal to `@vendoai/apps`, and widening that package's public API for a test
is not worth it):

- `packages/vendo/src/dev-creds/model.test.ts` — the ladder reports
  `claude-sonnet-5` BEFORE the first call under
  `ANTHROPIC_API_KEY` + `VENDO_MODEL=claude-sonnet-5`, keeps reporting the id it
  actually called the provider with, answers a bound explicit model object's own
  id, and still falls back to the placeholder with no credential. This is the
  test that fails on `main`.
- `packages/apps/src/model-params.test.ts` — the REAL engine, driven with a
  recording model shaped like the ladder wrapper (id readable only through a
  getter): a Claude 5 rung sends NO temperature and an explicit output cap, a
  Sonnet 4.6 rung still sends `temperature: 0`.

No existing test was modified.

## Gates

- `pnpm --filter @vendoai/vendo test` — 1511 passed, 15 skipped
- `pnpm --filter @vendoai/apps test` — 773 passed, 18 skipped
- `pnpm build`, `pnpm typecheck`, `pnpm lint` (dependency guard + portability
  gate included) — green

No UI surface is touched.

---
## Summary by cubic
Fixes a bug where `vendoModel()` exposed the placeholder id `vendo-env`, causing `@vendoai/apps` to treat Claude 5 as non-Claude and send `temperature: 0` (400 errors). The ladder now surfaces the resolved rung’s model id before the first call so sampling params are correct for the Claude 5 family.

- **Bug Fixes**
  - `modelId` on the ladder wrapper is now a getter that returns the resolved rung id, or the id chosen via sync env-based detection before the first call.
  - Introduced `detectDevCredential` (sync); `resolveDevCredential` now delegates to it so the id can be known without awaiting.
  - If no credential is present, the placeholder remains and no call succeeds. Tests in `@vendoai/vendo` and `@vendoai/apps` pin the pre-call id and parameter behavior (no `temperature` for Claude 5; `temperature: 0` for Sonnet 4.6).

<sup>Written for commit ca39ad6bb2c06512870ca6d31b0bcbad9814e135. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/694?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/ladder-surfaces-resolved-model-id.md
+++ b/.changeset/ladder-surfaces-resolved-model-id.md
@@ -1,0 +1,22 @@
+---
+"@vendoai/vendo": patch
+---
+
+A model pinned to the Claude 5 line through the ladder generates again.
+
+`vendoModel()` hands back a lazily-resolving model, and its `modelId` was the
+seam's own placeholder — the literal `vendo-env` — until the rung resolved. But
+the id is not decoration: the generation engine reads it to decide HOW to make
+the call, and the Claude 5 line rejects `temperature` outright. A placeholder
+reads as "not a Claude model", so `VENDO_MODEL=claude-sonnet-5` with a provider
+key kept `temperature: 0` and every single generation came back
+``400 `temperature` is deprecated for this model``. Sonnet 4.6 hid it by
+accepting the parameter, which is why the default ladder never showed it.
+
+The wrapper now reports the rung it will actually call: the resolved model's own
+id once resolution has run, and before that the id resolution will pick, which
+the ladder can answer without waiting because credential detection is pure env
+reading. Nothing else moves — the same precedence chooses the id, the announce
+line still prints it, and a rung with no credential keeps the placeholder
+because there is no real id to give. Hosts on a sampling-era model see exactly
+what they saw before.
--- a/packages/apps/src/model-params.test.ts
+++ b/packages/apps/src/model-params.test.ts
@@ -217,3 +217,39 @@ describe("engine model calls (integration)", () => {
     for (const call of calls) expect(call.temperature).toBe(0);
   });
 });
+
+// ---------------------------------------------------------------------------
+// #692: vendo's model ladder hands the engine a lazily-resolving wrapper whose
+// modelId is a GETTER over the resolved rung — the id is not a fixed string on
+// the object. The rule must read it at call time, per call, or a ladder pinned
+// to a Claude 5 rung keeps temperature: 0 and 400s. (The ladder half — that the
+// getter answers the real rung id — is pinned in
+// packages/vendo/src/dev-creds/model.test.ts.)
+// ---------------------------------------------------------------------------
+
+/** A recording model shaped like the ladder wrapper: the id is only knowable
+ *  through the getter, never as an own value on the object. */
+const laddered = (rungId: string): ReturnType<typeof recordingModel> => {
+  const { model, calls } = recordingModel("vendo-env");
+  Object.defineProperty(model, "modelId", { get: () => rungId });
+  return { model, calls };
+};
+
+describe("engine model calls behind the vendo model ladder", () => {
+  it("sends NO temperature when the ladder resolves to a Claude 5 rung", async () => {
+    const { model, calls } = laddered("claude-sonnet-5");
+    await createWith(model);
+    expect(calls.length).toBeGreaterThan(0);
+    for (const call of calls) {
+      expect(call.temperature).toBeUndefined();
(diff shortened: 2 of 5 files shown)


yousefh409:
Superseded by #703 on main.

PR #702, opened 2026-07-30
fix(ui): share anonymous session gate

### LINKED ISSUE #693: Anonymous wire identity churns per request on plain-http localhost — apps listing, thread follow-ups, and Regenerate all break in local dev
Found running the vendo-demos host locally (`pnpm dev`, no VENDO_API_KEY and also with one — both store rungs affected identically).

**Symptoms**, all in a real Chrome against `http://localhost:3000`:
- An agent turn creates an app (`vendo_apps_create` succeeds, card renders) — but `GET /api/vendo/<slug>/apps` from the same browser session returns `[]` forever. App rows in the store carry a DIFFERENT `anonymous_<id>` subject per turn.
- The SECOND message on any thread fails with `{"error":{"code":"conflict","message":"threadId is already in use"}}` — so Regenerate and every follow-up message are broken locally.
- Heartbeats and by-id thread ops still 200 (not subject-gated), which masks the problem.

**What I verified**:
- The anon cookie (`vendo_anon_session`, Path=/api/vendo, HttpOnly, SameSite=Lax, no Secure) IS minted, IS stored by the browser, and is NOT re-minted on curl requests that carry it (stable across a curl jar).
- Yet subjects on rows written seconds apart from the same browser differ, and list endpoints (subject-filtered) come back empty.
- Production (https, __Host- prefix path) demonstrably holds long multi-turn conversations on live demos, so this looks local/plain-http specific — possibly the `secureRequest`/cookie-name split (`__Host-vendo_anon_session` vs `vendo_anon_session`) disagreeing between mint and read somewhere in the wire (packages/vendo/src/wire/context.ts), or the streaming turn resolving context before/after a Set-Cookie race.

**Why it matters**: any feature that lists the caller's own records (VendoSlot self-discovery via useSlotApp, app pinning, thread history) cannot be developed or demoed locally, and every local multi-turn conversation dies on message 2.

**Repro**: vendo-demos repo, `pnpm dev`, open any demo, send a message via the trigger, then send a second message on the same thread → conflict. Or create an app and `fetch('/api/vendo/<slug>/apps')` from devtools → [].

erichanwang:
## summary

- share the anonymous-session cold-load gate across Vendo client instances for the same wire base
- normalize trailing slashes on `baseUrl` so `/api/vendo` and `/api/vendo/` use the same gate
- add a regression for concurrent first requests from multiple clients

fixes #693

## tests

- `pnpm --filter @vendoai/ui exec vitest run test/anon-session-race.test.ts`
- `pnpm --filter @vendoai/ui exec vitest run test/chrome/workspace-palette-slot.test.tsx`
- `pnpm --filter @vendoai/ui typecheck`
- `pnpm --filter @vendoai/ui build`
- `pnpm --filter @vendoai/vendo... build`
- `pnpm lint`

note: `pnpm --filter @vendoai/ui test -- test/anon-session-race.test.ts` ran the broader UI suite; the new race test passed, but one palette test failed there and passed when rerun directly.

---
## Summary by cubic
Share the anonymous-session cold-load gate across all `@vendoai/ui` client instances for the same base URL to stop races that mint multiple anonymous identities, fixing #693. Normalize `baseUrl` trailing slashes so `/api/vendo` and `/api/vendo/` use the same gate, and add a concurrency regression test.

<sup>Written for commit 96dd4b2ebda77f289c785933e0892e23e5b0eba6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/702?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/client-impl.ts
+++ b/packages/ui/src/client-impl.ts
@@ -75,9 +75,11 @@ function announceApprovalsDecided(detail: ApprovalsDecidedDetail): void {
   window.dispatchEvent(new CustomEvent<ApprovalsDecidedDetail>(APPROVALS_DECIDED_EVENT, { detail }));
 }
 
+const anonymousSessionGates = new Map<string, Promise<void>>();
+
 /** 08-ui §2 */
 export function createVendoClient(config: VendoClientConfig): VendoClient {
-  const baseUrl = config.baseUrl ?? "/api/vendo";
+  const baseUrl = (config.baseUrl ?? "/api/vendo").replace(/\/$/, "");
   const headers = { ...(config.headers ?? {}) };
 
   async function send(path: string, init?: RequestInit): Promise<Response> {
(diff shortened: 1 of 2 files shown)


yousefh409:
Thanks — the underlying bug was root-caused and fixed differently the same day in #704 (anon cookie path scoping). Closing as superseded.

PR #716, opened 2026-08-01
feat(ui): the pin ceremony — the panel dismisses, a ghost flies to the slot, and the slot re-reads on the pin

yousefh409:
Lane 1 of the **Keystone graduates** wave, item **B8**. Spec:
`docs/superpowers/specs/2026-07-31-keystone-graduates-design.md`.

Today a pin is silent. The user clicks "Pin to dashboard", the panel stays open
over the page, and the slot shows the app whenever its next ≤5s poll happens to
fire (`use-slot-app.ts`, `DEFAULT_POLL_MS = 5000`). Nothing connects the click
to the result — on a projector it reads as a dead button.

This graduates the sequence the Keystone demo designed and proved on stage
(`runvendo/vendo-demos` `host/src/vendo-kit/pin-flight.ts`) into the package,
adapted to its seams. Every pin runs one path: the in-thread card bar and the
workspace stage's pin both go through `usePinAction`.

## The sequence

1. **The panel goes first.** A pin's payoff is on the page, and the card being
   pinned sits in a centred panel over a scrim — the flight would otherwise play
   behind the thing hiding it. The ghost is lifted clear of the panel's own
   subtree onto a themed zero-size stage on `document.body` *before* the
   dismiss, or closing the overlay (`display: none` on the portal root) would
   make it vanish mid-flight.
2. **The ghost flies.** 300ms on a settling spring easing, opacity dissolving
   over the last 110ms, then a 180ms ring pulse drawn *over* the slot in the
   slot's own ink — 480ms end to end, deterministic, no randomness. The slot is
   scrolled into view (`block: "nearest"`, a no-op when it is already visible)
   in the frame the scroll lock releases, so the landing is actually watchable.
3. **The slot re-reads on the pin**, not on the poll tick — a module-scope pin
   bus that `useSlotApp` subscribes to. `onPin` is fire-and-forget by contract,
   so the immediate read can beat the host's write; a second read as the ghost
   lands covers that, and the poll stays as the floor.

`prefers-reduced-motion` keeps the dismiss and the pulse and skips the flight.
A slot that is not mounted means no animation rather than a stranded ghost — the
ceremony is presentation only; the pin is still whatever `onPin` writes.

## Intentional behavior: a pin scrolls its slot into view

Step 2 moves the host page's scroll position when the destination slot is off
screen. That is deliberate and ratified, not a side effect. The panel is a modal
the user may have opened from anywhere on a long page, and the entire point of
the ceremony is showing *where the pin landed* — a flight to somewhere above the
fold answers the question by playing off-screen. `block: "nearest"` means a slot
that is already visible is untouched, so the common case scrolls nothing.

This was found in the live proof below, not designed up front: the first capture
had the ghost flying to a slot above the fold with nothing to see.

## API (all optional, nothing existing changes)

- `usePinAction()` — what the built-in affordances call.
- `playPinCeremony({ appId, slot, dismiss })` — the same sequence for a host
  running a pin from its own control (the in-place-apply case).
- `announcePin(appId)` / `onPinAnnounced(listener)` — the bus, for a host that
  pins outside a Vendo surface.
- `pinSlot` on `VendoRoot`/`VendoProvider` — the destination. Only needed by
  hosts mounting **several** slots; with one mounted, the ceremony finds it, so
  `apps/demo-bank` needed no changes at all.

`onPin`'s signature is untouched: widening it to `void | PromiseLike<void>`
compiled here but broke `demo-accounting`, whose handler returns a value (TS's
void-return rule allows that, `void | PromiseLike<void>` does not). Not worth a
breaking change — the settle re-read covers the same ground.

## Tests

`packages/ui/test/chrome/pin-ceremony.test.tsx`, 8 cases: the ghost is lifted
clear and themed before the dismiss; the flight lands centred on the slot with
the tighter axis winning the scale; 300 + 180 < 500ms; the slot is scrolled in
before measuring; reduced motion pulses without flying; a missing destination
strands nothing; the single-mounted-slot fallback; and the slot showing the app
without waiting out the poll. That last one is the discriminating test — with
the bus subscription disabled it fails (`Unable to find "Invoices app surface"`
inside 1500ms against the 5000ms poll).

## Live proof

`apps/demo-bank` (Maple) run locally against the patched packages, driven
headless, real Anthropic model, real generated app, real click. Chromium's own
animation timeline reports the durations; frames captured at 0.12× playback so
the 480ms is visible.

```
signed in → http://localhost:3210/
home-hero slot mounted
sent: Show my spending by category as a chart
app card ready, Pin to dashboard offered
t+   0ms  ghost=1 stage=1 ring=0 panelVisible=false
t+ 350ms  ghost=1 stage=1 ring=0 panelVisible=false  matrix(0.9714, …, 275.8, -196.5)
t+ 900ms  ghost=1 stage=1 ring=0 panelVisible=false  matrix(0.9573, …, 412.8, -294.2)
t+1500ms  ghost=1 stage=1 ring=0 panelVisible=false  matrix(0.9556, …, 428.9, -305.6)
t+2200ms  ghost=0 stage=0 ring=1 panelVisible=false
t+4200ms  ghost=0 stage=0 ring=0 panelVisible=false
durations reported by the browser: 300ms cubic-bezier(0.32,1.04,0.36,1) ·
                                   110ms delay 190 · 180ms ease-out
slot filled 33ms after the click (the default poll cadence is 5000ms)
```

Reduced-motion run: no ghost, no flight, one 180ms ease-out pulse, panel
dismissed, slot filled 32ms after the click.

The scroll-into-view step came out of this proof: the first capture had the
ghost flying to a slot above the fold, so the payoff played off-screen. See
"Intentional behavior" below.

## Gate

`pnpm build && pnpm test && pnpm typecheck && pnpm lint` all green (673 tests
in `@vendoai/ui`, 53 tasks across the monorepo). Changeset included (`minor`).
Per the spec's hard constraint the npm release and every demo-fleet redeploy
queue until after the YC show — this PR is code only, and nothing here touches
the frozen keystone demo.

---
## Summary by cubic
Add the pin ceremony so pinning feels instant: the overlay dismisses, a ghost flies to the dashboard slot, and the slot refreshes on the pin instead of waiting for the 5s poll. Adds live-run verification docs; supports reduced motion and keeps existing `onPin` behavior.

- **New Features**
  - `usePinAction()` — unified pin flow used by the thread card bar and stage.
  - `playPinCeremony({ appId, slot, dismiss })` — trigger the ceremony from host controls.
  - Immediate slot refresh via bus: `announcePin(appId)` / `onPinAnnounced(listener)` exported from `@vendoai/ui` and `@vendoai/vendo`, with a second settle read (~500ms).
  - Optional `pinSlot` on `VendoProvider` to target a specific slot; single-slot hosts auto-detect.
  - Scrolls the slot into view; 300ms flight + 180ms pulse; honors `prefers-reduced-motion`.
  - Verification: added live run frames and transcripts under docs/verification (Keystone B8).

- **Migration**
  - No breaking changes; `onPin` remains `void`.
  - Hosts with multiple slots should set `pinSlot` on `VendoProvider`.
  - If pinning outside a Vendo surface, call `announcePin(appId)` to re-read slots immediately.

<sup>Written for commit f810fd57c0880cf16be54689717386d437e273b1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/716?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Frames from the live run are committed under `docs/verification/keystone-graduates-b8/` (README there maps each one), so they render inline from the diff:

- [01 — panel dismissed, ghost lifted, nothing moved yet](https://github.com/runvendo/vendo/blob/feat/ui-pin-ceremony/docs/verification/keystone-graduates-b8/01-panel-dismissed-ghost-lifted.png)
- [02 — ghost in flight over the bare page](https://github.com/runvendo/vendo/blob/feat/ui-pin-ceremony/docs/verification/keystone-graduates-b8/02-ghost-in-flight.png)
- [03 — settle pulse on the slot](https://github.com/runvendo/vendo/blob/feat/ui-pin-ceremony/docs/verification/keystone-graduates-b8/03-settle-pulse.png)
- [04 — pinned](https://github.com/runvendo/vendo/blob/feat/ui-pin-ceremony/docs/verification/keystone-graduates-b8/04-pinned.png)

Plus `live-run.txt` / `live-run-reduced-motion.txt` with the per-frame DOM state and the durations Chromium's own animation timeline reported.

PR #720, opened 2026-08-01
fix(ui): surface stale app refresh failures

### LINKED ISSUE #717: useApp: a failed reopen leaves an already-rendered app showing stale content, with no error and no recovery
## What happens

An app that has **already rendered** keeps showing stale content, with no error
and no way back, when a later `refresh()` fails every attempt.

`useApp`'s catch arm records the error but leaves `app` and `surface` on the
previous successful load. Both mounting surfaces gate on the surface first:

- `VendoSlot` → `MountedApp`: `if (!surface) { …error / skeleton… }` — a
  non-null stale `surface` short-circuits straight to `AppFrame`.
- `VendoPage` → `OpenApp`: identical shape.

So the terminal-failure state added in #710 is unreachable once a load has
succeeded once. The user sees a view that silently stopped updating.

The realistic trigger is the served-app keepalive (`execution-v2` Wave 7 H2):
`AppFrame`'s `keepalive.reopen` is `useApp`'s `refresh`, and it fires precisely
when the machine reports it *woke* — i.e. exactly when a reopen is most likely to
fail. The retry ladder from #710 (3 attempts, 300ms/600ms) makes the window
narrower, not closed.

## Not a regression

`main` before #710 behaved identically — the pre-#710 `refresh` also did
`setError(...)` in its catch and left `surface` untouched
(`756231d09:packages/ui/src/hooks/use-app.ts`). #710 only added retries and the
never-loaded-yet error state; it did not touch this path. Filing it separately
rather than expanding a merged lane, because the fix needs a design decision
this one doesn't.

## Why it needs design, not a one-liner

Blanking a working view on a failed background refresh is worse than showing
slightly stale data — the app may be perfectly usable and only the reopen
failed. The honest shape is probably a **non-destructive staleness affordance**:
keep the rendered app, mark it out of date, offer the same "Try again" the
terminal state now has. That is new chrome and wants a look from the design
skill, plus a decision on whether a keepalive-driven failure reads differently
from a user-driven `refresh()` failure.

## Provenance

Found by Greptile on #710 (P1), which is the UI-fixes lane of the Keystone
graduates wave — `docs/superpowers/specs/2026-07-31-keystone-graduates-design.md`,
item **A5** ("`useApp` never retries a failed `apps.open`"). A5 is about the
never-loaded skeleton; this is its already-loaded twin and the spec does not
cover it.

## Pointers

- `packages/ui/src/hooks/use-app.ts` — the catch arm that keeps the stale surface
- `packages/ui/src/chrome/vendo-slot.tsx` — `MountedApp`, `SlotLoadFailed`
- `packages/ui/src/chrome/vendo-page.tsx` — `OpenApp`
- `packages/ui/src/tree/frames.tsx` — the `keepalive` contract that calls `reopen`

CoderSufiyan:
Closes #717.

Keeps an already-rendered app visible when a later `useApp.refresh()` exhausts its retries. Both slot and workspace mounts now show an explicit stale-state notice with a retry action instead of silently presenting stale content as current.

Tests:
- `pnpm build`
- `pnpm --filter @vendoai/ui test`
- `pnpm typecheck`
- `pnpm lint`

---

<a href="https://app.devin.ai/review/runvendo/vendo/pull/720" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---
## Summary by cubic
Keeps the app mounted when a `useApp.refresh()` fails after retries and shows a clear “stale” notice with a retry button. Closes #717 by preventing stale content from appearing current without warning.

- **Bug Fixes**
  - Added `AppStaleNotice` and render it in `vendo-page` and `vendo-slot` when `error && !isLoading`, showing the error and a “Try again” action.
  - Preserve the previous `surface` on refresh failure; tests cover surface retention and slot recovery with the stale notice.

<sup>Written for commit dba06942ee93d1db854b166baa03c5029e544063. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/720?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/chrome/app-stale-notice.tsx
+++ b/packages/ui/src/chrome/app-stale-notice.tsx
@@ -1,0 +1,10 @@
+/** A failed refresh must not hide a usable app, but users need an honest way
+ * to see that it may be out of date and ask for another refresh. */
+export function AppStaleNotice({ error, onRetry }: { error: Error; onRetry(): void }) {
+  return (
+    <div role="alert" className="fl-error">
+      This view may be out of date — {error.message}
+      <button type="button" className="fl-error-retry" onClick={onRetry}>Try again</button>
+    </div>
+  );
+}
--- a/packages/ui/src/chrome/vendo-page.tsx
+++ b/packages/ui/src/chrome/vendo-page.tsx
@@ -7,6 +7,7 @@ import { useThreads } from "../hooks/use-threads.js";
 import { AppFrame } from "../tree/frames.js";
 import { ActivityPanel } from "./activity-panel.js";
 import { AutomationsPanel } from "./automations-panel.js";
+import { AppStaleNotice } from "./app-stale-notice.js";
 import { ChromeRoot } from "./chrome-root.js";
 import { ACTIVITY_ANCHOR_ATTRIBUTE, ACTIVITY_BUMP_EVENT } from "./morph-toast.js";
 import { ConnectedAccountsPanel } from "./connected-accounts-panel.js";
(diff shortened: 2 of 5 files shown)


CoderSufiyan:
Addressed the integrated-coverage finding in dba06942. The new slot-level regression mounts a rendered app, triggers the real `keepalive.reopen` callback after forcing `apps.open` to fail, and asserts that the rendered app remains visible beside the stale notice and retry action.

Focused validation: `pnpm --filter @vendoai/ui exec vitest run test/chrome/app-load-retry.test.tsx test/chrome/app-stale-slot.test.tsx`.

PR #728, opened 2026-08-02
test(demo-accounting): fix flaky jwt test by deterministically mutating the signature

### LINKED ISSUE #722: Flaky CI: Cadence session test's "tampered" JWT is byte-identical ~0.1% of runs
`cadence-supabase-auth` fails intermittently on `apps/demo-accounting/src/server/__tests__/session.test.ts:85`:

```
AssertionError: expected { …(3) } to be null
+ Received: { display: "Maya Alvarez", email: "maya@cadence.test", subject: "8d0158a1-…" }
```

## Root cause

The assertion tampers a token by replacing its last two base64url characters:

```ts
await expect(resolveCadenceSession(withCookie(`${token.slice(0, -2)}xx`))).resolves.toBeNull()
```

An HS256 signature is 32 bytes → 43 base64url characters, and 43 characters carry 258 bits. The final character's low 2 bits decode to nothing, so distinct character pairs can decode to the **same** 32 bytes. When that happens the "tampered" token is not tampered, it verifies correctly, and the assertion fails. Nothing is wrong with `resolveCadenceSession` — the test's mutation is not guaranteed to be a mutation.

## Repro (measures the rate directly)

```js
const b64u = c => "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_".indexOf(c);
const dec = s => { let bits = 0, n = 0; const out = []; for (const c of s) { n = (n << 6) | b64u(c); bits += 6; if (bits >= 8) { bits -= 8; out.push((n >> bits) & 0xff); } } return out; };
let collisions = 0; const N = 200000;
for (let i = 0; i < N; i++) {
  const sig = Buffer.from(crypto.getRandomValues(new Uint8Array(32))).toString("base64url");
  if (Buffer.from(dec(sig)).equals(Buffer.from(dec(`${sig.slice(0, -2)}xx`)))) collisions++;
}
console.log(collisions / N);
```

```
identical-bytes rate: 194/200000 = 0.097%
```

So roughly 1 CI run in 1,000 on this job fails for no reason. It cost a red gate on #714 (a `packages/ui` PR that does not touch this file).

## Fix

Tamper the signature deterministically instead of textually — flip a byte of the decoded signature and re-encode, or mutate a character in the **payload** segment (where every character is significant), or assert against a signature built with a known-wrong key (that arm of the same test is already deterministic and correct).

Found while landing #714 (Keystone graduates B7).

Akarshkushwaha:
Fixes #722

## What
Fixes the ~0.1% flaky test in `apps/demo-accounting/src/server/__tests__/session.test.ts` where tampered tokens would occasionally successfully verify. The test now tampers with the JWT deterministically by mathematically flipping the first base64url character of the signature block, rather than slicing off the end.

## Why
Because an HS256 signature is exactly 32 bytes (256 bits), and 43 base64url characters can hold 258 bits, the final character has unused bits. The previous approach of blindly swapping the last two characters (`token.slice(0, -2) + "xx"`) occasionally produced a string that decoded to the exact same 32-byte signature, causing the test to falsely pass a "tampered" token and break the CI gate. 

By flipping the *first* character of the signature (which utilizes all 6 bits of its base64url character), we guarantee that the decoded signature bytes will be altered 100% of the time, while preserving the payload structure so we strictly test cryptographic validation.

## Verification

---

<a href="https://app.devin.ai/review/runvendo/vendo/pull/728" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

---
## Summary by cubic
Fixes a flaky JWT verification test in `examples/demo-accounting/src/server/__tests__/session.test.ts` by deterministically tampering the signature: flip the first base64url character so the HS256 signature always differs and the test reliably rejects tampered tokens. Extracts the tampered token to a local const to keep line length.

<sup>Written for commit d52b4189b602116153a6e47c6afc02122678b61e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/728?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Thanks — this was a correct fix, but #860 deleted the whole Cadence demo including this test three days after the PR opened, so it's moot now. Closing.

PR #736, opened 2026-08-02
Rebuild cutover: harness runtime, claudeCode() live sessions, turn-aware MCP door

yousefh409:
# The rebuild: harness runtime, claudeCode(), and a turn-aware MCP door

This lands the rebuild program staged on `rebuild/cutover` (waves 1–3 + the cc-native and door-ctx lanes). ~320 commits, each wave gated and independently verified before landing here.

## What's in it

- **Harness runtime**: the chat route runs on swappable harnesses — `vendo()` (default), `instant()` (non-agentic specialist), `claudeCode()` (Claude Agent SDK in a real e2b sandbox). One transcript, one guard, one audit log across all three.
- **claudeCode() = one live Claude Code session per conversation**: stock `claude_code` preset + a few appended lines, native skills discovery via `plugins`, workspace materialization with per-conversation diff baselines. The old machine pool / snapshot / token-rotation / relay-bridge machinery is deleted.
- **Turn-aware MCP door**: the door now accepts turn-scoped credentials — an opaque, turn-lifetime bearer minted only inside a live turn. Audit rows carry real `venue`/`presence`, approvals park the turn and render as cards on the user's surface, transcript mirror + commit inherited from the in-process path. Proven **identical** to the in-process path by the flipped parity gate (`mcp-door-parity.e2e.test.ts`). Outside-agent (PKCE) behavior pinned before the change and byte-identical after.
- **Orgs, sharing, sponsorship, org policy** (wave 3) and the wave-1/2 foundations underneath.

## Gates (on this head, uncached)

`pnpm build` 24/24 · `pnpm test --force --concurrency=1` **55/55 tasks, 7,739 tests** · `pnpm typecheck` 43/43 · `pnpm lint` 6/6.

## Live browser proof (recorded on this head, claudeCode() through the public door, real e2b box)

Approval card parks the turn → Approve executes it → generated app fills with live data; a deny is narrated and nothing moves. Frames and logs live on the lane branches (`rebuild/cc-native`, `rebuild/door-ctx`) rather than in the tree — see below.

Cross-take reality check: the account balance in the memory proof is exactly $40 lower than before the approved transfer — the four recordings share one live world. Full evidence (raw logs, frames, close notes) is reachable on the lane branches and at commit `c5f0f0a7e`; it was stripped from the tree in `a7776428c` so this PR is reviewable and the public repo carries product, not process.

## Known limits (documented in close notes)

- Files grown past 8 MiB inside a box read as deleted at turn-end sync (`WALK_SKIP_BYTES`, pre-existing, documented in the cc-native close note §7).
- A box is held (≤90s) while an approval waits — pinned as a test so the cost is visible.
- In-box `Task` subagents run auto-allowed without a hire receipt (SDK limitation).
- Generation-engine quality on binding-heavy asks is a known, separately-tracked issue (#631) — not touched by this PR.

## Follow-up in flight

`rebuild/door-internal`: internal-only door mode so `harness: claudeCode()` needs zero config and `mcp: true` stays purely "outside agents allowed" — lands as its own PR after its independent check.

## Reviewer notes

- Process exhaust (agent memory, raw proof logs, lane contracts, eval run dumps — ~164k lines) was stripped in `a7776428c`. Frozen specs under `docs/superpowers/specs` stay: source comments cite them as contracts.
- Known gap carried in, not introduced here: `/orgs/**` paths are dropped by workspace materialization on the `claudeCode()` harness, so team-shared files are invisible to that harness (arrived with the wave‑3 merge; own lane).


yousefh409:
## Review triage (two agents, since cubic/greptile both bailed on size)

**Blocking, fixes in flight — do not merge yet:**

1. **Unrestricted sandbox egress** (security review, verified in code). `boxMachine()` provisions the conversational box with no `allowedDomains`, and the adapter contract + e2b adapter treat that as `allowInternetAccess: true`. `Bash` is auto-allowed in the box with no guard, so injected content can exfiltrate the materialized workspace with no audit row. The app-builder box one layer over already does this correctly (`runtime.ts` → `boxAllowlist`), and `claude-code/index.ts:85` currently claims deny-by-default egress that isn't wired. Fix lane: `fix/box-egress`.
2. **`/orgs/**` is dropped for the `claudeCode()` harness** (merge-seam review). `pathAccess()` returns `none` for every org path, so team-shared files never reach the box; a second `/user/`-only filter in `turn-routes.mjs`/`local.ts` means edits to them would not come back either. `materialize.test.ts:22` pins the old behavior, so no suite goes red. Arrived with the wave-3 merge, not this PR's two lanes, but this PR carries it to main. Fix lane: `fix/orgs-materialize`, rewiring onto the store's existing `visibleFiles()`/`canCommit()` (which had zero production callers).

**Recorded, own lane, not blocking:** `wire/box.ts:293` rebuilds a RunContext with no memberships, so org policy doesn't bind on `/box/tools` and a promoted app's box authenticates as the org id; `machine: "local"` reuses the sandbox's auto-allow permission hook on a non-sandboxed host process; no test anywhere combines memberships with the turn-bearing door.

**Clean under adversarial check:** turn-credential design (opaque, turn-scoped, unforgeable — 11 abuse negatives, two revert-verified red); both session-space crossings (OAuth↔turn) closed in both directions; approval replay/tamper pinned on 8 fields with single-winner atomic claim; no token values in logs; all four hand-resolved merge files are line-for-line unions with wave-3's fixes intact; the one positional-arg slip was caught by the compiler and fixed in 31ddb2afd.

**Housekeeping:** `a7776428c` strips ~164k lines of factory process exhaust (agent memory, raw proof logs, lane contracts/plans, eval run dumps) and gitignores those paths; frozen specs stay because source comments cite them as contracts.

**Still open on this PR:** it is CONFLICTING against main (main has moved 33 commits ahead — the remix work — with 31 conflicts), and no build/test/typecheck/lint gate runs on the PR itself. Gates are green locally on the head; that is asserted, not proven at PR level.

PR #742, opened 2026-08-03
feat(cli): init never edits your source; sync owns the whole scan

yousefh409:
Five decisions from the founder, implemented as given. Gates green:
`pnpm build && pnpm test && pnpm typecheck && pnpm lint` (53/53 tasks, 1604
tests in `@vendoai/vendo` alone), plus a real end-to-end run of the built CLI
against a scratch Next app (terminal output below).

---

## 1 — Init never writes to a user-authored file

`wireNextLayout` rewrote the host's `app/layout.tsx` to wrap `{children}` in
`<VendoRoot>`. It's gone, along with `LAYOUT_CHILDREN` and
`importInsertionIndex` (its only callers) and the `{ kind: "wired" }` arm of
`LayoutWiring`. No dead exports left behind.

Everything init still writes is a **new** Vendo-owned file (`vendo/registry.tsx`,
`vendo/vendo-root.tsx`, the catch-all route, `.vendo/*`) — plus `package.json`,
which stays in scope. **My judgment, called out as asked:** `package.json` is
config, not source; init only maintains its own two script keys, the edit is
idempotent, and it is what makes `vendo sync` run at all. Removing it would
silently stop extraction on every existing install. Tests agreed — nothing
argued the other way.

The mount is now a first-class printed step:

```

  File: app/layout.tsx

```

Same data in three places, from one source (`mountStep`): the framed block, the
head of `manualSteps`, and a new `mount: { file, lines, why }` on the `--agent`
plan. Express/custom hosts are unchanged — they have no single host file to
name, so they keep their compact wiring lines.

**Doctor** already failed on the missing mount (`E-WIRE-004`) but only
*described* the fix. It now prints the exact paste with the resolved file path
and the resolved import specifier, and says init will never make the edit for
you. `docs-site/agents/verify.mdx` updated to match.

## 2 — AI judgment: no saved consent, flags + interactivity decide

One rule, applied identically to `init` and `sync`:

| | no flag | `--ai` | `--no-ai` |
| --- | --- | --- | --- |
| Interactive (TTY) | **asks, every run** | runs | off |
| Non-interactive (CI, pipe, `--json`, `--yes`) | off, with a note | runs | off |

- `--ai` / `--no-ai` is now the canonical pair on both commands. `--ai-polish`
  (init) and `--no-watermark` (sync) still work — both were documented.
- Nothing is persisted; there was no saved-consent state to remove (I checked
  `.vendo/` and the CLI — the old init already asked per-run; sync simply
  never asked at all and ran the pass unconditionally in CI, which is the drift
  this fixes).
- `--yes` implies non-interactive on both; `--json` does too, and still emits
  exactly one object.
- `--ai` + `--no-ai` together is a loud parse error on both commands.
- The auto-installed hooks carry the flag: `predev: vendo sync --no-ai`,
  `prebuild: vendo sync --strict --no-ai`. Re-running init **upgrades the exact
  hookless entry an older init wrote** (`vendo sync` / `vendo sync --strict`),
  in place, wherever it sits in a `&&` chain, idempotently. Any *other*
  `vendo sync …` in your scripts is yours — init leaves the whole script alone.

⚠️ **Behavior change worth flagging:** `vendo sync` in a terminal now asks a
consent question it never asked before. That is the decision as written
("interactive: ALWAYS ask, every run"), but it is the one thing a user will
notice immediately.

### The provenance design choice

`.vendo/theme.json` is the editable source of truth and is imported straight
into the host's client bundle, so I did **not** put provenance inside it.
Instead there is a sibling merge base, **`.vendo/theme.extracted.json`** — what
the deterministic scan produced last time. That mirrors the split `.vendo/`
already lives by (`tools.json`/`judgments.json` = machine layer,
`overrides.json` = what a person decided) rather than inventing a second
convention, keeps `theme.json` exactly the frozen `VendoTheme` shape, and ships
no dead bytes to the browser. Both files should be committed. Init writes the
base from the **deterministic** summary only — before any model fill or
`--theme` answer, because those are decisions and sync must pin them.

Per slot, on every sync:

- in the base and `theme.json` still equals it → machine-extracted, **updated**
- in the base and `theme.json` differs → **pinned**, and reported
- absent from the base → updated only while `theme.json` still holds Vendo's
  neutral default, pinned otherwise
- `radius.small`/`radius.large` follow a `radius` change only while they still
  hold the values init derived from `radius.medium`

**No base yet** (every install from before this PR): nothing is machine-owned,
so nothing is touched. If the scan and `theme.json` agree everywhere, the base
bootstraps silently; if they disagree, the run warns with the diff **every
sync** until a human resolves it with `--theme-refresh`. I deliberately do not
advance the base while a disagreement stands — that would quietly bake the
stale value in as the new truth.

Reported in the normal output line and in `--json` as
`theme: { updated, pinned }`.

## 4 — Keyed sync pushes pin baselines to Cloud

New `packages/vendo/src/cli/cloud/pin-baselines.ts`. When a key resolves
(`--key`, `VENDO_API_KEY`, or `.env.local`/`.env` via doctor's parser), a
**normal** sync — no `--report` — reconciles `.vendo/remixable/` with the
`vendo_pin_baselines` collection through the ordinary public store door
(`hostedStore`, the same adapter a keyed runtime uses). One record per slot,
`id = slot`, `data` = the baseline, validated by the OSS `pinBaselineSchema` —
matched against the console's consumer (`apps/console/lib/remix-reviews/cloud.ts`,
vendo-web `main`). New/changed slots are pushed, slots pruned locally are
deleted remotely, unchanged slots are skipped (hash + capturedAt + review).

### Exactly what crosses the wire

The baseline file **verbatim**: `slot`, `hash`, `exportable`, `capturedAt`,
`review`, **the wrapped component's full source**, `subSources` (the source of
every module it imports within the capture depth), `sourceImports`, and
`styles` (the captured app-root stylesheets). That is the point — a reviewer
cannot review a fork's diff against source Cloud does not have — but it is now
an explicit, documented fact in `docs-site/reference/cli.mdx`.

Keyless/BYO: unchanged, stays local, **zero network calls** (asserted in a
test with a `fetchImpl` that throws). A Cloud failure is a warning line naming
that the next sync retries; the exit code never moves. Skipped entirely when
`.vendo/remixable/` does not exist.

Reported in `--json` as `baselines: { pushed, pruned }` (`null` when keyless).

## 5 — The stale CLI docs

`docs-site/reference/cli.mdx`'s sync section described a git-diff-from-watermark
enrichment writing `.vendo/overrides.json`, "restrictive-only". Replaced with
what actually happens: an evidence-backed judgment pass writing
`.vendo/judgments.json`, candidates chosen by `srcHash` drift / binding
mismatch, `--full` to judge everything, loosenings held for a human, and
`overrides.json` as read-only prompt context meaning "what a human decided".
Also added: the flag matrix from 2, the theme behavior from 3, the baseline
push from 4, the corrected init section, the `--agent` `mount` field, the new
`--json` fields, and the new options.

---

## Tests

New:

- `packages/vendo/src/cli/sync.coherence.test.ts` (**new file**, 16 tests) —
  the sync flag matrix (interactive asks twice / non-interactive never prompts
  / `--yes` and `--json` never prompt / `--json` emits one object / `--no-ai`
  wins in a TTY), the theme reconcile (rebrand lands, hand edit pinned,
  `--theme-refresh`, no-base path, silent bootstrap, `--json` shape, no
  `theme.json`), and the baseline push (record shape through the store door,
  prune + no re-push, keyless makes no request, a hiccup is a warning).
- `init.test.ts` — a new `the mount paste` suite (layout byte-identical,
  ambiguous layout, `--agent` `mount` parity, overlay-missing), a new
  `AI flag matrix` suite, and a new `sync hooks` suite (both hooks carry
  `--no-ai`, in-place idempotent upgrade, never clobbers a user's own call).
- `cli.test.ts` — `--ai`/`--no-ai` parse on both commands, legacy spellings,
  and both-at-once rejected.

**Test files touched, and why** (each is a semantic consequence of a decision,
never a weakened assertion):

- `packages/vendo/src/cli/init.test.ts` — deleted the `wireNextLayout` suite
  (the function is gone); the layout-fallback suite became the mount-paste
  suite; the main fixture now asserts 4 wired files and a layout containing no
  `VendoRoot`; hook assertions carry `--no-ai`; the `--agent` plan asserts
  `app/layout.tsx` is **not** in `codeChanges`; `aiPolish:` seams renamed to
  `ai:`; two skip-message strings updated.
- `packages/vendo/src/cli/sync.test.ts` — the judgment-integration tests now
  pass `ai: true` (they exercise the pass, which no longer runs by default) and
  the `--no-ai` test passes `ai: false`; the `--json` assertions gained
  `theme`/`baselines` and the new skipped-judgment note.
- `packages/vendo/src/cli/init-judgment.unattended.test.ts` — `consent: true`
  renamed to `ai: true`.
- `packages/vendo/src/cli/onboarding-safety.test.ts` — the pages-only host's
  two output strings follow the new block format; the assertion it protects
  (names `pages/_app.tsx`, never `app/layout.tsx`) is unchanged.
- `packages/vendo/src/cli.test.ts` — help + accepted-flag lists.

## Live proof — the built CLI against a scratch Next app

Fresh Next host, `app/layout.tsx` at `acecc0b7…`:

```console
$ node packages/vendo/dist/cli.js init . --yes
Wired (4 files):
  + vendo/registry.tsx
  + app/api/vendo/[...vendo]/route.ts
  + vendo/vendo-root.tsx
  ~ package.json
Learned: 1 tools · theme captured → .vendo/ (tools.json, theme.json, brief.md)
AI polish (descriptions, risk review, brief, theme): skipped — this run cannot ask (pass `--ai` to run it non-interactively, or run `vendo init` in a terminal).
Theme: accent #7c3bed · background #ffffff · ...

  File: app/layout.tsx

$ shasum app/layout.tsx
acecc0b70e4b08691b84d46a590c1bb28be83882   # byte-identical

$ node -p "require('./package.json').scripts"
{ dev: 'next dev', build: 'next build',
  predev: 'vendo sync --no-ai', prebuild: 'vendo sync --strict --no-ai' }
```

Doctor **before** the paste:

```console
$ node packages/vendo/dist/cli.js doctor .
OK     wiring/next-route — catch-all handler is wired
BROKEN wiring/next-root  — no app layout mounts <VendoRoot> — Vendo is wired but invisible. In app/layout.tsx, paste: import { VendoRoot } from "../vendo/vendo-root";  … then wrap: <VendoRoot>{children}</VendoRoot>. (The generated vendo/vendo-root.tsx wrapper's export is also named VendoRoot and mounts <VendoOverlay />; any layout that covers your pages works. `vendo init` never edits your source, so this paste is always yours.)
BROKEN wiring/surface    — no visible agent surface is mounted
```

Doctor **after** pasting exactly those two lines:

```console
OK wiring/next-route  — catch-all handler is wired
OK wiring/next-root   — <VendoRoot> wraps the app
OK wiring/surface     — a visible agent surface is mounted (<VendoOverlay /> or an equivalent)
OK wiring/dependency  — @vendoai/vendo dependency is declared
```

Theme re-scan, live (`--primary` changed in `globals.css`, `accent` hand-edited
in `theme.json`, `radius.small` hand-edited):

```console
$ vendo sync . --no-ai
theme: 1 slot re-read from your app (radius) → .vendo/theme.json
theme: your brand changed but these slots are pinned to your edits (accent → #1d4ed8) — `vendo sync --theme-refresh` takes the app's values

accent #ff5a1f (kept)  |  radius {"small":"2px","medium":"16px","large":"24px"}
#  medium+large followed the scan; the hand-edited small (2px) did not.

$ vendo sync . --no-ai --theme-refresh
theme: 1 slot re-read from your app (accent) → .vendo/theme.json
accent #1d4ed8  |  small 2px   # forced, and the base advanced
```

Baseline push, against a stand-in console store on `127.0.0.1:4599` seeded with
a stale `LegacyHeroCard`, with a real `<Remixable review>` wrapper in the host:

```console
$ vendo sync . --no-ai --key vnd_aaa… --api-url http://127.0.0.1:4599
pins: 0 captured, 0 drifted
baselines → Vendo Cloud: 1 pushed, 1 pruned (component source crosses the wire so the console can review forks)

# what the store saw:
POST /api/v1/store/records/vendo_pin_baselines/list    authorization=Bearer vnd_aaaaa…
POST /api/v1/store/records/vendo_pin_baselines/put     authorization=Bearer vnd_aaaaa…
  STORED NetWorthCard: {"slot":"NetWorthCard","source":"export function NetWorthCard({ total }: { total: number }) {\n  return <section className=\"net-worth\">Net worth: ${total}</section>;\n}\n","hash":"sha256:8b3f8085…","exportable":false,"capturedAt":"2026-08-03T01:59:54.364Z","review":true,"styles":[{"path":"app/globals.css","css":":root {…
POST /api/v1/store/records/vendo_pin_baselines/delete  authorization=Bearer vnd_aaaaa…
  DELETED LegacyHeroCard

# re-run: nothing changed, nothing re-sent (no baselines line at all)
```

Keyless, with a baseline on disk:

```console
$ vendo sync . --no-ai --json
baselines: null | theme: {"updated":[],"pinned":[]} | notes: []
# the injected fetch throws on any call — it was never called
```

---

## Notes for review

- `examples/demo-accounting/.vendo/theme.extracted.json` is committed: its
  `predev` hook ran the new sync, the scan agreed with the committed theme, and
  the base bootstrapped. Without it every build leaves an untracked file.
- **Finding, not fixed here:** `demo-bank` and `demo-template` have real theme
  drift the old init-only flow hid — their committed `theme.json` disagrees
  with their own CSS on `border` (`#ecebe8`) and `fontFamily`
  (`Inter, sans-serif`). Sync now says so on every build. Resolving it is a
  demo-design call (is the committed value deliberate?), so I left both files
  alone — which is the never-clobber law doing its job.
- Not done: a fully green `vendo doctor` on the scratch host (its live probes
  need an installed dep tree, a running dev server, and a model key). The four
  static wiring checks — the ones this PR moves — are green, before/after.

---
## Summary by cubic
Init no longer edits user files; sync now owns theme re-extraction and baseline uploads, with a single `--ai`/`--no-ai` flag pair across `init` and `sync`. This adds safety, keeps CI deterministic, and fixes theme/base churn, derived slot handling, and Cloud prune edge cases.

- **New Features**
  - `init`: drops layout auto-wire; prints an unmissable mount block with exact lines. `vendo doctor` `E-WIRE-004` now prints the same paste. `--agent` plans include `mount`.
  - Flags: unified `--ai` / `--no-ai` on `init` and `sync`. Interactive asks every run; non-interactive (CI, `--yes`, `--json`, package scripts) skips. Legacy `--ai-polish` and `--no-watermark` still work.
  - Theme: added `.vendo/theme.extracted.json` merge base without timestamps; `sync` updates the base only when slots change. Machine-owned slots update; unrecorded slots are always pinned. Derived slots move only when their source moves (`accentText` ← `accent`, `headingFamily` ← `fontFamily`). Output and `--json` list exactly the slots written under `updated` and show `pinned` with both values. Use `--theme-refresh` to accept app changes.
  - Cloud baselines: keyed `sync` pushes `.vendo/remixable/` to the `vendo_pin_baselines` store and prunes stale records by file presence; unreadable files are skipped with a warning. Keyless makes no network calls. Reported as `baselines` in `--json`.
  - Scripts: `init` writes `predev: vendo sync --no-ai` and `prebuild: vendo sync --strict --no-ai`, upgrading older entries in place.

- **Migration**
  - After `init`, paste the `<VendoRoot>` import and wrapper into your layout; the CLI prints the exact lines.
  - Commit `.vendo/theme.extracted.json`. If `sync` reports theme disagreements, run `vendo sync --theme-refresh`.
  - Expect `sync` to prompt in a terminal; pass `--ai` or `--no-ai` to avoid prompts. Package-script runs (e.g., `npm run dev`) are treated as non-interactive.
  - Re-run `vendo init` to upgrade existing script hooks.

<sup>Written for commit 2b1c1726242736e8d68a14ada94fa37518bf18b9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/742?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
## Review round 1 — all three blockers fixed, plus the IMPORTANTs

Pushed as `6c99158`. Gates green: `pnpm build && pnpm test && pnpm typecheck && pnpm lint` (53/53 tasks). Every fix has a test, and I mutation-checked the riskiest one (reverted the fix, watched the test fail, restored) so none of them are vacuous.

---

### BLOCKER 1 — `theme.extracted.json` churned on every sync ✅

You were right on both halves. `baseFrom()` stamped a fresh `at`, and `writeBase` wrote unconditionally.

- `at` is **gone** from the format — it carried no decision.
- `writeBase` now reads the current base first and writes only when the slots actually differ (`sameSlots`), returning whether it wrote.

Live: two consecutive no-op syncs, `shasum` identical. Test: `two consecutive no-op syncs leave theme.extracted.json byte-identical`, which also pins the shape to exactly `{ format, slots }` so a timestamp can't come back.

I regenerated `examples/demo-accounting/.vendo/theme.extracted.json` in the new format (the old committed one still had `at`, and `sameSlots` would have left it there forever).

### BLOCKER 2 — never-clobber failed on the upgrade path ✅

Fixed by deleting the guess entirely. `NEUTRAL_DEFAULTS` is gone; the law is now one line:

> A slot is machine-owned **only** when the base records it and `theme.json` still holds that value. Everything else is pinned.

That covers both the no-base case *and* "base exists, this slot was never recorded" — which had the same hazard and the same non-answer. Your framing went into the code comment, because the next reader needs to know why the obvious optimization is wrong:

```ts
// Machine-owned ONLY with recorded proof. No base entry means no evidence
// about who chose the value on disk, so it is the human's — never guessed
// from "it looks like our default".
```

**Corrected no-base behavior, re-stated:** with no base file, *nothing* is machine-owned, so sync touches nothing. It reports the full diff as pinned and repeats that warning on **every** sync until a human resolves it with `--theme-refresh` or edits `theme.json`. The base only advances on a run with zero pinned slots — an unambiguous run — so a stale value is never quietly adopted as the new truth. Code, PR body, and `cli.mdx` now all say exactly this.

Live probe of your exact scenario (CSS `#0f766e`, human accent `#2563eb`, no base):

```
theme: your brand changed but these slots are pinned to your edits (accent → #0f766e) — `vendo sync --theme-refresh` takes the app's values
accent on disk: #2563eb
base file: none (correct — run was ambiguous)
```

Tests: `a human value that happens to equal Vendo's neutral default is still pinned` and `an unrecorded slot is pinned even when the base exists for other slots`.

### BLOCKER 3 — a corrupt baseline deleted the remote row ✅

The comment described a skip; the behavior was a delete. Now the code matches the comment: `localBaselines` returns `present` (every slot with a **file**, parsed or not — the file stem *is* the slot, so presence is knowable without parsing) alongside `valid` (the push payloads) and `unreadable`. Prune keys off `present`. Malformed JSON is caught (N2 — it previously threw out of the whole push), and a file whose `data.slot` disagrees with its filename is also treated as unreadable, so a run can't push a row and then delete it.

Reproducing this live took a second try worth mentioning: on a normal sync, `capturePins` prunes the stale *files* before the push ever runs, so the bad state can't survive. The real path is a **degraded scan** — capture already prunes nothing when it hits a wrapper error, a missing compiler, or the file cap — which leaves a half-written file on disk for the push to read. So the fix makes the push honor the same invariant `capturePins` already defends. Probed with a deliberate wrapper error:

```
warning: unreadable baselines left untouched in Vendo Cloud: TruncatedCard — re-run sync to recapture .vendo/remixable/<slot>.json
baselines → Vendo Cloud: 1 pushed, 0 pruned
--- deletes: (none — correct)
```

Test: `a corrupt local baseline is skipped and warned — never a delete`.

---

### I1 — lifecycle runs are not interactive ✅ (done, documented)

`invokedByPackageScript()` in `shared.ts`, applied to **both** `init` and `sync`: a set `npm_lifecycle_event` means non-interactive, default off. Your ruling is in the code comment ("A run the human did not invoke never gets a question") and in `cli.mdx`, which now spells out that "non-interactive" includes package scripts and *why*.

Mutation-checked: the test forces a real TTY (`process.stdin.isTTY = true`) — without it the assertion is vacuous under vitest — and I confirmed it fails when the exemption is removed.

### I2 — one budget over the whole reconcile ✅

One `AbortController` with a 20s wall-clock budget for the entire reconcile, riding a wrapped `fetch` (`consoleSender` overrides per-request `signal`, so it can't ride a request option). Blowing it aborts the in-flight request and degrades to the warning, with a message naming the budget. Test asserts it returns in well under the per-request timeout.

### I3 — deferred, with a reason 🟡

I did **not** add a manifest row, and I want your call rather than a silent skip. A manifest is a second source of truth for a collection the console and the verification harness can both write directly (`cloud.ts` says so explicitly) — if a row is deleted or written out-of-band, a manifest-driven sync silently never re-pushes it. That trades a correctness bug for a latency win, right after a round of correctness bugs. A hash-only projection is the right fix, but the store adapter (01-core §12 `records.list`) has no projection, so it needs a contract amendment — bigger than this lane. I2's budget bounds the wall-clock damage in the meantime. Happy to do the manifest if you'd rather have the speed now.

### I5 — the stale auto-wire docs ✅

All six locations, and I found more rot in the shipped skill while I was in it:

- `packages/vendo/skills/vendo-setup/SKILL.md` — the auto-wire lines, **plus** the `aiPolish` delegation contract and `npx vendo extract --apply`, which describe a command that no longer exists. An agent following that hits a hard failure. Replaced with the real `sync --ai` path and an explicit "you are non-interactive: pass the flag you mean". Also added the mount step and the `--no-ai` CI line.
- `docs-site/quickstart.mdx:61`, `docs-site/connect/vendo-init.mdx:18,42`, `docs/quickstart.md:50,216` — all corrected, plus the stale `predev: vendo sync` hook example.

### I6 — the cli.mdx contradiction ✅

Resolved with B2's actual behavior; the "taken only while it holds the neutral default" bullet is gone.

### N1 ✅ — `sameValue()` normalizes colors before comparing. You called it: it *was* most of the demo noise — `demo-template`'s `border` warning disappeared. What remains in `demo-bank`/`demo-template` is genuine (`#E2E1DE` vs `#ecebe8`; committed `system-ui, …` vs CSS `Inter`), correctly pinned, and left alone per your scope call.

### N2 ✅ — `JSON.parse` is inside the try; malformed JSON is an unreadable file, not a thrown push.

### N3 ✅ — `pushPinBaselines` never throws. It returns `{ pushed, pruned, unreadable, error? }`, so `--json` reports rows that really landed even when the transport dies mid-loop. Test kills the second `put` and asserts `{ pushed: ["AaaCard"], pruned: [] }` plus the warning.

### N6 ✅ — `--report` now uses the same resolved `cloudKey` as the baseline push (the #567 dotenv fix applies to every keyed leg, not just the judgment pass).

---

### Left alone per your scope call

I4 (init regenerating `route.ts`/`vendo-actions.ts` — pre-existing, founder decision), I7 (consent asymmetry: own-model asks, Cloud upload doesn't — worth a founder ruling, since the Cloud upload is the one that ships source off the machine), and the demo theme values.

Ready for your live proof. Not merging.

yousefh409:
## Round 2 — fixed, and the label was hiding a worse bug

Pushed as `2b1c172`. Gates green (53/53). Thanks for the live proof — and for this one especially, because chasing the wrong line found a real visual defect.

### What I found reproducing it

My first repro printed the *correct* message, so I brute-forced a scenario matrix. The bad line only appears when a **derived** slot is involved, and the derivation is where the actual bug lives:

```
### hand accent #2563eb (pinned), css --primary #7c3bed -> #fde047
    theme: 1 slot re-read from your app (accentText) → .vendo/theme.json
    theme: your brand changed but these slots are pinned to your edits (accent → #fde047)
    ON DISK accent #2563eb accentText #000000
```

`accentText` is *derived* from the accent — but the merge treated it as an independent slot. So with the accent pinned, sync wrote the contrast color computed for the **app's** light-yellow accent (`#000000`) on top of the human's **dark blue** `#2563eb`. Black text on dark blue. The mislabelled summary line was the symptom; this was the disease, and it would have shipped a genuinely broken palette to anyone who pinned an accent while their app's accent moved across the light/dark contrast boundary.

### 1. Derived slots now follow their source

`accentText ← accent` and `headingFamily ← fontFamily` may only move when their source moves — the same law `radius.small`/`radius.large` already followed. Held silently: nothing is mis-written, and the pinned source is already named in the report.

```
holds a derived slot when the slot it derives from is pinned   → accentText stays #ffffff
a derived slot DOES follow its source when it is machine-owned → accentText becomes #000000 on #fde047
```

Mutation-checked: removing the two-line guard fails the first test with `expected '#000000' to be '#ffffff'`.

### 2. The count and the named slots now reflect what was written

`updated` was already only the written slots in the code path you'd expect — the "accent" in your output came from the derived-slot bug feeding a *second* slot into the list while the pinned line named accent separately. With the derivation fixed, the list is exactly what landed in `theme.json`, and the test asserts the negative too (`not.toContain("re-read from your app (accent, background)")`).

### 3. Pinned slots are reported, unambiguously

They *were* being reported — but I think the wording is what actually misled a reader, and it misled you the same way:

```
theme: your brand changed but these slots are pinned to your edits (accent → #b91c1c)
```

`accent → #b91c1c` reads as an assignment. Now both values are shown, so there's nothing to misread:

```
theme: 1 slot re-read from your app (accent) → .vendo/theme.json · 1 pinned by you, unchanged (border — yours #111111 vs your app's #e2e8f0) — `vendo sync --theme-refresh` takes your app's values
```

One line, both halves, "unchanged" said out loud.

### 4. `--json`

Same corrected `theme: { updated, pinned }` — both lists of slot names, `updated` = written, `pinned` = held. Asserted in the regression test: `{ updated: ["accent"], pinned: ["background"] }`.

### Tests

Three new, in `sync.coherence.test.ts`:

- `names exactly the slots written, and reports the pinned ones separately` — one written + one pinned in the same run; asserts the written list, the negative, the pinned detail with both values, the on-disk result for both slots, and the `--json` shape.
- `holds a derived slot when the slot it derives from is pinned` (mutation-checked).
- `a derived slot DOES follow its source when the source is machine-owned` — so the fix can't degrade into "derived slots never update".

Docs updated: `cli.mdx` gains a **Derived slots follow their source** paragraph explaining the black-on-dark-blue case, states that the "re-read" list is exactly what was written, and shows the real combined line. Changeset updated too.

Ready for your merge.

PR #747, opened 2026-08-03
feat(guard)!: risk grading stops guessing from names; ungraded asks; critical → confirmEach

yousefh409:
Implements [`docs/superpowers/specs/2026-08-02-risk-grading-redesign.md`](https://github.com/runvendo/vendo/blob/risk-grading-redesign/docs/superpowers/specs/2026-08-02-risk-grading-redesign.md) (committed here — it was not in the repo). Nothing beyond D1–D5; the spec's out-of-scope list is untouched.

## What changed

**D1 — all name-based inference dies.** `DESTRUCTIVE_WORDS` / `READ_WORDS` (actions `sync/common.ts`) and `DESTRUCTIVE_TOKENS` / `READ_VERBS` (`connectors/composio-risk.ts`) are deleted, along with the corpus harness's own `DESTRUCTIVE_NAME` regex. No code path concludes anything from a tool's name.

**D2 — only protocol facts grade at extraction.** HTTP `DELETE` → `destructive`; a declared GraphQL/tRPC `mutation` → `write`; Composio's `destructiveHint`/`readOnlyHint` → `destructive`/`read`. A `GET` is not a fact about reading, a `POST` is not a fact about writing, and a Next server action is neither. Everything else → `ungraded`.

**D3 — `ungraded` is a first-class `RiskLabel` that asks.** The treatment sits beside the existing `default → run` fallthrough in the guard pipeline, not in an init-written rule, so a hand-wired server with **no policy config at all** gets it too. A host loosens it consciously, in writing, with a `{ "match": { "risk": "ungraded" }, "action": "run" }` rule.

**D4 — not-knowing is felt.** `vendo doctor` reports the count (`E-TOOLS-003`, with its verify-page anchor); keyless `vendo init` / `vendo sync` say what the consequence is instead of implying the grades are real.

**D5 — `critical` → `confirmEach`, behavior identical.** Host-authored files read `critical:` as an alias indefinitely, via one preprocess at the four authored schemas; every writer emits `confirmEach`. The judge rubric is renamed, not rewritten.

## ⚠️ Breaking behavior

An unjudged catalog now asks on its mutations. That is the whole point — `payInvoice` classified `write` and ran un-gated — but it is a real behavior change for any install that never ran the AI judge. Three ways forward, each one a sentence, all in the changeset and the new `E-TOOLS-003` verify section.

Two consequences worth calling out:

- `descriptorHash` follows the field rename, so approvals/grants persisted before the upgrade no longer match their tool's new hash. They lapse into a re-ask — the fail-closed direction. Frozen vectors regenerated, plus a new `ungraded` vector.
- **The demo catalogs (`examples/demo-{bank,accounting,template}/.vendo/tools.json`) regenerated to all-`ungraded`** — they were relying on the word lists and have no `judgments.json`. Maple and Cadence will now ask on every host tool call until their catalogs are judged. That is honest, and it is what the E2E proof below shows, but it is a **product decision for Yousef**: judge the demo catalogs (`vendo sync` with a key, committed) or give them a `risk: "ungraded"` policy rule.

## Acceptance criteria — proof

| AC | Test file | Test name |
| --- | --- | --- |
| 1 grep-clean | `packages/actions/src/sync/protocol-facts.test.ts` | `carries no word list anywhere under packages/` |
| 2 ungraded by default | `packages/guard/test/ungraded-default.test.ts` | `asks for an ungraded tool on a guard with NO policy config at all` · `parks the call end to end rather than executing it` |
| 2 bare-GET ungraded | `packages/actions/src/sync/protocol-facts.test.ts` | `leaves a GET ungraded: GETs that mutate exist, so GET alone proves nothing` |
| 2 status reports it | `packages/vendo/src/cli/doctor-codes.test.ts` + live run below | registry snapshot incl. `E-TOOLS-003` |
| 3 payInvoice, un-judged | `packages/guard/test/ungraded-default.test.ts` | `un-judged: ungraded, so it asks instead of silently paying` |
| 3 payInvoice, judged | `packages/guard/test/ungraded-default.test.ts` | `judged write + confirmEach: asks EVERY call, and the standing grant is never consulted` |
| 4 GET alone ungraded | `packages/actions/src/sync/protocol-facts.test.ts` | `leaves a GET ungraded…` · `grades DELETE destructive — the one HTTP method that is a fact` |
| 5 legacy `critical:` loads | `packages/actions/src/critical-alias.test.ts` | 6 cases: overrides tool, overrides compound, judgments field + pending, tools.json entry, explicit-wins, typo-still-fails |
| 6 corpus re-baselined | `corpus/harness/src/layers/scored.test.ts`, `corpus/harness/src/ai/score.test.ts` + `pnpm corpus run` below | protocol-fact annotation check, identity-only inventory key, `gradings` bucket |
| 7 gates | below | |

The AC1 grep test is verified non-vacuous two ways: it asserts it swept >200 files, and dropping a `DESTRUCTIVE_WORDS` probe file into `packages/actions/src/` makes it fail (`expected [ Array(1) ] to deeply equal []`).

### AC6 — corpus, run for real

`pnpm corpus` refuses to run from a path containing a space (the harness says so itself: *"local-pack known issue: paths containing spaces break local Vendo package injection"*), and this worktree lives under `…/Desktop/Cool Code/…`. Run from a space-free export of this exact tree:

```
| express-host | Layer 1 structural | PASS | 7/7  |
| express-host | Layer 2 scored     | PASS | 10/10 |
```

with the catalog it extracted — exactly D2, nothing else graded:

```
  ungraded     POST   /api/tasks/{id}/complete  host_completeTask
  ungraded     POST   /api/tasks                host_createTask
  destructive  DELETE /api/tasks/{id}           host_deleteTask
  ungraded     GET    /api/tasks/{id}           host_getTask
  ungraded     GET    /api/tasks                host_listTasks
```

## E2E proof — real `vendo sync`, real doctor, real browser turn

**1. Keyless `vendo sync` on `examples/demo-bank` (judge skipped):**

```
tools: +0 -0 ~0
pins: 0 captured, 0 drifted
judgment: structural-only (no model credential — set ANTHROPIC_API_KEY / OPENAI_API_KEY (BYO) or VENDO_API_KEY (`vendo login`)) — 24 tools unjudged
```

**2. The catalog it wrote — every tool ungraded (no DELETE routes in Maple):**

```
grade counts: {'ungraded': 24}
  ungraded  POST  /maple/api/orders
  ungraded  GET   /maple/api/accounts/{id}
  ungraded  GET   /maple/api/insights/budgets
  …
```

**3. `vendo doctor` (the status surface — see deviations, there is no `vendo status` command):**

```
[ok]      tools/live-surface: 22 live host tools
[warning] tools/graded: catalog: 24/24 tools ungraded — each one asks on every call; run `vendo sync` with a model key to grade
          error_code=E-TOOLS-003  fix_ref=https://vendo.run/agents/verify?v=0.6.1#E-TOOLS-003
```

**4. A real agent turn against the running demo** (`pnpm --filter demo-bank dev`, real `ANTHROPIC_API_KEY`, real browser, signed in as Yousef Helal). Asked *"List my recent transactions."* → the agent called the **ungraded** `host_listTransactions` → it **parked with an approval that actually surfaces**: the card renders "NEEDS YOUR APPROVAL / List transactions", the risk chip reads **"Not reviewed"** (the new `ungraded` label), real inputs are shown, Approve/Deny are live, and the status line says *"List transactions — waiting for your approval"*. Not a refusal, not a 200 with nothing.

Maple: an ungraded tool parks with a real approval card, risk chip "Not reviewed"

Clicking **Approve** completes the loop with real host data, so the ask is a gate and not a dead end:

The same turn after Approve: real transaction data renders

The server log shows the same thing across the boot-time chip pre-generation:

```
[vendo] query "hostListTransactions" (tool "host_listTransactions") resolved no data for app app_2aea… — the call answered "pending-approval"; anything bound to it renders empty
```

## Gates

```
BUILD_EXIT=0        Tasks: 23 successful, 23 total
TYPECHECK_EXIT=0    Tasks: 41 successful, 41 total
LINT_EXIT=0         Tasks:  6 successful,  6 total
TEST_EXIT=1         Tasks: 49 successful, 53 total
```

Re-run at the round-4 tip (`8878a518a`): build, typecheck and lint are 0; the
test task's failures are the same known set below, plus `automations-e2e` under a
fully parallel `--continue` run (its fixture server answers 500 when starved —
35/35 pass with the package run alone).

The failing test tasks are **pre-existing failures, all reproduced at HEAD with this branch's work stashed** — plus, on a loaded machine, timing flakes that pass in isolation (`@vendoai/bench` `never exposes a partial file to a concurrent reader`, `@vendoai/apps` `renders islands in parallel and stays within budget`, and the `automations-e2e` fixtures, which boot a Next dev server each):

| Failure | Cause | Proof |
| --- | --- | --- |
| `packages/core/src/packaging.e2e.test.ts` (3) | The worktree path contains a space; Vite cannot load `file:///…/Cool%20Code/…` | Fails identically at `15f2e51f8` with a clean tree |
| `packages/store/src/{durability,split-brain}.drill.test.ts` (2) | Same: `Cannot find module '/Users/…/Cool%20Code/…/drill-writer.mjs'` | Fails identically at `15f2e51f8` with a clean tree |
| `packages/vendo/src/dev-creds/model.test.ts` (2) | `Cannot find module '@ai-sdk/anthropic'` in this install | Already failing in the pre-change baseline run of this branch |

None is reachable from this diff, and none exists in a space-free checkout (which is what CI and `~/orca/workspaces/flowlet/github` use — the same tree passes there, which is how these came back as turbo cache hits in the first baseline).

## Checker round 1 — six findings fixed (commit `836fdc5d5`)

| # | Finding | Fix | Test |
| --- | --- | --- | --- |
| 1 HIGH | App-venue dead approval loop: an ungraded app QUERY parked, the owner approved, and nothing landed — queries are never re-dispatched, and the refetch minted a fresh call id while the guard's approved replay pins it | Query call ids are derived from `(app, tool, args)` — the same triple the query resolver already keys its own state on — so the refetch **is** the call the user approved. Actions keep random ids: two identical mutations are two acts | `packages/vendo/src/ungraded-query-approve.e2e.test.ts` — `open → park → approve → reopening renders real data`. Verified to FAIL without the fix: `expected undefined to deeply equal { dining: 41200, … }` |
| 2 MED | `readonly` preset inversion: ungraded fell to the guard's ask-default, so the one preset that BLOCKS a known write offered an approve button for an ungraded tool | Every preset treats ungraded exactly as it treats destructive | `packages/guard/test/ungraded-default.test.ts` — `makes every preset treat ungraded exactly as it treats destructive` |
| 3 MED-LOW | Unclassifiable extraction still claimed `risk: "destructive"` | Nothing spoke → `ungraded` + `disabled`, in route-scan, server-actions, trpc, graphql | `packages/actions/src/sync/{route-scan,server-actions,trpc,graphql}.test.ts` — the four `emits … disabled with a note` cases |
| 4 LOW-MED | Write breaker counted only write\|destructive | Every non-read grade spends the per-run budget | `packages/guard/test/ungraded-default.test.ts` — `spends the per-run write budget — an ungraded call is not a free call` |
| 5 LOW-MED | Generation's law-2 checks skipped ungraded, so a generated surface could invoke an unknown-effect tool with fabricated args | `isMutatingRisk` now means "not provably a read" | `packages/apps/src/generation-ungraded-laws.test.ts` — incl. `flags fabricated args and a missing required input on an ungraded tool` |
| 6 LOW | Docs still taught `critical` | Renamed across README, `tools-and-safety`, `compound-tools`, `agents/tools`, `api-tools`, `reference/cli`, the `vendo-setup` skill; `tools-and-safety` now also names `ungraded` | — (prose) |

Also amended **the spec itself** (D4 + AC2), in place and marked as an accepted build-time deviation: there is no `vendo status` command, so doctor is the status surface.

## Checker round 2 — three items closed (commit `12f5f8d1d`)

| # | Item | Fix | Test |
| --- | --- | --- | --- |
| A MED-LOW | Deny was not durable: `#consumeApprovedCall` only read APPROVED rows, so a caller re-issuing a stable call id (the derived query ids from round 1) minted a fresh card on every retry — deny, reopen, new card, forever | A matching DENIED record now answers the re-issue as `action: "block", decidedBy: "denied"` instead of re-parking. Both lookups share one `sameParkedCall` matcher so a yes and a no can never mean different calls | `packages/guard/test/ungraded-default.test.ts` — `blocks the re-issue, attributes it to the person, and mints no second card`, `keeps standing until the QUESTION changes`, `does not leak across subjects`. E2E: `packages/vendo/src/ungraded-query-approve.e2e.test.ts` — `a denied query never reads the host, and the no is DURABLE across reopens` (3 reopens, zero pending) |
| B(i) LOW-MED | Zero-arg ungraded tools were unbindable: empty payload hit `missing-payload`, repair fell to `action-disclaim`, and a "Refresh" button vanished off every keyless catalog | `missing-payload` is exempt when the schema **declares** zero properties and does not open itself | `packages/apps/src/generation-ungraded-laws.test.ts` — `does not demand a payload from a tool that declares no inputs` (+ still demands one for declared inputs, an open schema, and an absent schema) |
| B(ii) LOW-MED | The substitution message called an ungraded tool "MUTATING" — the same guess this redesign deleted | It now names what is known: `UNGRADED tool "…"` + "nobody has graded this tool", so the operand rule reads as *unknown effect*, not *write* | `packages/apps/src/generation-ungraded-laws.test.ts` — `says UNGRADED, and says why that is worse rather than better` + `still says MUTATING for a tool that really is graded a write` |
| C | Design ruling, **no behavior change**: an approved replay crossing the same user's sessions is intended (approve on phone, laptop renders) | Deliberate-choice comment at the `#consumeApprovedCall` match | `packages/guard/test/ungraded-default.test.ts` — `lets a laptop session spend an approval the phone session parked — exactly once` + `never crosses to a DIFFERENT user, however identical the call` |

**Expiry semantics for A**, decided against the existing TTL machinery: a denial is *not* consumed. A yes is spent because it authorizes one act; a no is a standing answer about a question, and it stands until the question changes — different inputs, or a re-graded tool (whose descriptor rehashes) both miss the match and ask again. That is strictly more durable than `sweepExpiredApprovals`, which only ever touches PENDING rows and would have left the re-park loop running.

`decidedBy: "denied"` is new on the block union (`packages/core/src/guard.ts` + the audit enum) — nothing switches exhaustively on it; the activity ledger just prints it.

Corpus re-verified after round 1: `express-host` Layer 1 **7/7 PASS**, Layer 2 **10/10 PASS**.

## Checker round 3 — the deny fix's own blocker, closed (commit `d92203f56`)

Round 2's denial was **provenance-blind**. Four different things write a denied row — a real decision, the chat turn the user walked away from, a BYO embed timing out, the 60-minute TTL sweep — and it honored all four as the user's standing no. Worst case, verified: the apps runtime's secret-exposure and egress ceremonies re-issue a **stable** call id behind a **frozen** synthetic descriptor, so one hour of inattention permanently bricked the app's ability to ever ask again, with no undo.

| # | Item | Fix | Test |
| --- | --- | --- | --- |
| 1 HIGH | Every denial converged on the same provenance-less row | Rows carry `deniedBy: "human" \| "system"`. `approvals.decide` is human; `abandonApprovals` and `sweepExpiredApprovals` are system. Only a human denial stands; a system one reaps the pending row and the next issue asks again. Absent on pre-existing rows → reads as `system`, the fail-safe direction | `packages/guard/test/ungraded-default.test.ts` — `does NOT stand when housekeeping wrote it — the TTL sweep is not an answer`, `does NOT stand when the conversation walked away — abandonment is not an answer` |
| 2 MED | Parking never dedupes, so one call id could hold an older unconsumed APPROVED row and a newer DENIED one, and the replay lookup runs first — the old yes executed after the newer no | A human deny voids its unconsumed approved siblings | `the newest human decision on a call wins` — `approve THEN deny: the stale yes is voided, not replayed after the no` and `deny THEN approve: the approval was minted after the no, so it wins` |
| 3 MED-LOW | `#standingDenial` scanned all of a subject's historical denials on every ask draft | Rows carry a `call` ref (`call_id` column + composite index), so chat's random ids miss on the index; human-only provenance keeps the pile small | Covered by the whole denial suite through the routed store + the PGlite fixture, which both enforce ref-key routing |
| 4 LOW | `takesInput` only accepted `additionalProperties === true` | A schema-valued `additionalProperties` reads as OPEN (catalog-scan emits that shape) | `packages/apps/src/generation-ungraded-laws.test.ts` — `reads a SCHEMA-valued additionalProperties as open, not as zero-arg` |
| 5 LOW | Ledger printed the raw `denied` slug, which reads as a fresh refusal | `decidedByLabel` → "previously denied" (plus `confirmEach`, `default`) | `packages/ui/test/chrome/activity-semantics.test.ts` — `says what actually happened: an older no is still standing` |
| 6 | Un-deny | `approvals.revoke(id, principal)` + `DELETE /approvals/:id` — the exact mirror of `grants.revoke` / `DELETE /grants/:id`, so no new concept. A revoked denial stops answering; a revoked unconsumed approval can no longer replay. Owner-scoped, idempotent, pending conflicts | `takes a human no back through approvals.revoke…`, `guards revoke like every approval read: foreign not-found, pending conflicts`. E2E: `a misclicked no is recoverable: revoke the decision and the region asks again` |
| 7 | Known limit, comment only | `descriptorHash` excludes the binding, so a host re-pointing a route behind a byte-identical descriptor inherits the old denial — documented at `#standingDenial`, with `approvals.revoke` named as the way out | — |

**Item 6 needed no new concept**, so I built it rather than stopping: `DELETE /grants/:id → guard.grants.revoke` already existed, and this is the same six lines against the same collection. Rows are **voided, not deleted**, so the audit trail keeps both answers in order.

**Store change (unavoidable).** `vendo_approvals` is a typed table, so `deniedBy`/`voidedAt` would have round-tripped to nothing and `call` would not have been queryable. Added `denied_by`, `voided_at`, `call_id` as additive `ALTER TABLE … IF NOT EXISTS` migrations plus a `(subject, status, call_id)` index, and taught the three projections that mirror it: the routed store, core's memory adapter, and the guard's PGlite fixture.

**Per-package runs** (the gate stops at the first failing task, so each package was also run alone on an idle machine): `@vendoai/guard` 191 passed · `@vendoai/core` 704 passed + the 3 packaging · `@vendoai/store` 219 passed + the 2 drills · `@vendoai/apps` 811 passed · `@vendoai/ui` 699 passed · `@vendoai/actions` 600 passed · `@vendoai/corpus-harness` 401 passed · `automations-e2e` 35 passed.

## Checker round 4 — voiding and replaying were racing on nothing (commit `8878a518a`)

| # | Item | Fix | Test |
| --- | --- | --- | --- |
| 1 BLOCKER | `#consumeApprovedCall` read the row, claimed the `consumed:<id>` receipt, then put a STALE copy back; both void paths (`#supersedeApprovedSiblings`, `#revokeApproval`) wrote `voidedAt` with a plain put and claimed nothing. Interleaving: consumer reads un-voided → the deny completes and reports success → consumer wins the receipt → the tool runs and the void marker is gone | Voiding SPENDS the approval's one-time transition: every void claims the SAME `consumed:<id>` receipt, so exactly one of a replay and a void can happen. The receipt records which claimant won, so the loser reports honestly — a second-place replay refuses (the call parks and the standing no blocks it), a second-place take-back says `conflict` / writes a `supersedeTooLate` audit line rather than reading as success. The consume path re-reads after claiming, so it can never write a stale row | `packages/guard/test/ungraded-default.test.ts` — `refuses the replay when the no lands first: the voided yes is never spent` · `says the take-back came too late when the replay wins: the call ran, and the trail says so` · `cannot take back a yes the call already spent — revoke says so instead of reporting success` |
| 2 | A BYO expiry stamped `deniedBy: "human"` and planted a standing denial for a timed-out embed, because the `abandonApprovals`-less fallback fell through to plain `approvals.decide` | The fallback denies through the human-consent verb (the only one on the wire — it must not grow a provenance argument) and then immediately takes its own no back, so the row stays in the trail but stops standing. `abandonApprovals` still does it natively with system provenance | `packages/vendo/src/byo-approvals.e2e.test.ts` — `expires through the fallback path without turning the timeout into the user's no` (real store, real guard, `abandonApprovals` hidden behind a proxy) |
| 3 | `#supersedeApprovedSiblings` hand-rolled a LOOSER match than `sameParkedCall` (it omitted venue/presence/appId), so a no could void a yes it could not itself answer | One shared matcher, used by the replay, standing-denial and supersede lookups. `sameParkedCall`'s ctx parameter widened to `ApprovalRequest["ctx"]` so a stored decision matches a stored decision with no cast | Covered by the whole denial + replay suite (both orderings, the cross-session cases, and the app-venue reads suite) |
| 4 | Polish | A double revoke writes ONE audit event (the second is `already-void` and returns silently) · a `report()` failure after a landed void no longer surfaces as a revoke error · the routed store's `updatedAt` reflects `voidedAt` (the memory adapter already did) · the dead `approvalStore.decide` helper is deleted — zero consumers outside its own test, and 02-store §3 retired the typed-helper architecture · the ledger names the too-late outcome instead of spinning on "Running" forever | `takes a decision back exactly once — a second revoke adds no second audit line` · `packages/store/src/routing.test.ts` — `routes and updates approvals…` · `packages/ui/test/chrome/activity-semantics.test.ts` — `says a no that arrived mid-replay came too late — never a row still Running` |

All four race tests were verified non-vacuous by reverting `packages/guard/src/guard.ts` alone: 4 failed / 20 passed. The BYO test likewise fails (`expected 'blocked' to be 'pending-approval'`) with only `byo-approvals.ts` reverted.

**Per-package runs at this tip** (a fully parallel `turbo run test --continue` starves the automations fixture server, so each package was also run alone): `@vendoai/guard` 201 passed · `@vendoai/store` 219 passed + the 2 known drills · `@vendoai/ui` 700 passed · `@vendoai/core` 704 passed + the 3 known packaging cases · `@vendoai/vendo` 1595 passed + the 2 known dev-creds cases · `automations-e2e` 35 passed.

## Checker round 5 — the void protocol's boundaries (commit `6ebce49a1`)

| # | Item | Fix | Test |
| --- | --- | --- | --- |
| 1 BLOCKER | Erasure resurrection: the replay's post-claim re-read fell back to its listed copy when the row read back `null`, but subject erasure (`packages/store/src/erase.ts`) and anonymous-subject adoption (`helpers/subjects.ts`) DELETE approval rows — so a mid-flight replay re-created an erased subject's approval and ran the tool as them | `null` means GONE: the consume aborts, writes nothing, and the call parks. Same law applied at the other two re-read-then-write sites (the void path, the batch compensation) | `packages/guard/test/ungraded-default.test.ts` — `never resurrects an approval erased mid-replay — no row, no run` (deletes the row between the claim and the re-read; asserts nothing ran and nothing came back) |
| 2 | The automations engine was a third approval-row writer outside the receipt: `markConsumed` plain-put `consumedAt` without claiming `consumed:<id>`, and `approvalRowSchema` stripped `deniedBy`/`voidedAt` on the way back — a revoke and a grant mint could both win, and the winner erased the take-back marker | Core's `Guard` grows an optional `spendApproval(id, principal)` — the same feature-detected shape as `abandonApprovals` — because arming a standing grant SPENDS a yes without replaying its call (07 §3) and has to contend on the one transition. It answers `spent` / `already-spent` / `taken-back`; the engine arms the grant only on `spent`. The row schema passes unknown keys through, the pre-seam fallback refuses a take-back it can see, and the parked-run path arms nothing for a yes taken back inside the resume window | Guard half (real receipt, both orderings): `spends a yes for a grant-arming caller, and a take-back after it comes too late` · `refuses to spend a yes the person took back first — the grant is never armed`. Engine half: `packages/automations/src/engine.test.ts` — `arms nothing when the person took the yes back…` · `arms nothing when someone else already spent the yes` · `arms exactly one grant per won spend, and leaves the approval row to the guard` · `fallback for a Guard predating the seam…`. Real stack: `automations-e2e` 35/35 with every grant now minted through the seam |
| 3 | Receipt-durable-before-put recovery lie: a void whose `store.put` threw left the receipt claimed and the row un-voided, so the retry read "another void holds it" and reported success while a standing human denial kept blocking | The loser re-asserts `voidedAt` before answering, and reports it as the take-back that landed — so the trail gets exactly one line for it, not zero | `re-asserts a take-back whose row write failed instead of calling it already done` (breaks one approvals put, then retries, then proves the next issue ASKS) |
| 4 | Batch-deny compensation restored `member.prior` blindly, erasing a concurrent replay's `consumedAt` or a take-back's `voidedAt` and re-opening an ask whose single-use transition is already gone | A small fence fitted: skip the restore when the current row is consumed or voided, and never re-create a row that has been erased. Everything else still rolls back all-or-none, and the skipped member stays decided — which its own audit line already says | `does not re-open a batch member a concurrent spend already consumed`. Verified non-vacuous by deleting the fence alone: `expected … "status": "approved", "consumedAt" … received "status": "pending"` |
| 5 | Cosmetic | Done: the mixed-version deploy window is documented at the receipt read (a receipt written before claimants existed reads as `undefined`, which the void path treats as "spent" — fail-closed, and true, since the older build claimed it only from the replay path). **Not done:** collapsing the BYO expiry's `Denied` + `Decision taken back` ledger pair. It appears only on the legacy-guard fallback path, both lines are true, and suppressing one would mean either a `report`-suppressing argument on `approvals.revoke` or a provenance argument on the wire — the thing item 2 of round 4 exists to avoid | — |

Every round-5 test was verified non-vacuous by reverting the source alone: guard 5 failed / 24 passed, automations 4 failed / 35 passed.

**Per-package runs at this tip** (`turbo run test --continue` starves the two fixture-server suites, so each was also run alone): `@vendoai/guard` 206 passed · `@vendoai/automations` 51 passed · `@vendoai/core` 704 passed + the 3 known packaging cases · `@vendoai/store` 219 passed + the 2 known drills · `@vendoai/vendo` 1595 passed + the 2 known dev-creds cases · `@vendoai/ui` 700 passed · `automations-e2e` 35 passed · `redteam-e2e` 21 passed (4 of them time out under a fully parallel run; 17s alone vs 118s loaded, and `fixtures/host-app`'s `next build` loses its `.next` dir to a concurrently booting dev server the same way).

## Checker round 6 — the last automations path off the receipt (commit `e9a35f943`)

| # | Item | Fix | Test |
| --- | --- | --- | --- |
| 1 FIX | Parked-run resume armed a standing grant from a stale copy: it read the approval, checked THAT copy's `voidedAt`, and minted. A revoke landing in between returned success yet the grant armed anyway — the replay was refused, but `#matchingGrant` accepted the brand-new grant and the tool ran on the next fire | The grant is minted, the call replayed, and the grant **taken back** when the replay did not win (parked or blocked). The receipt stays the arbiter — the only thing that can tell "revoked mid-resume" from "we read the row a moment too early" — so a yes taken back leaves no standing authority behind, wherever in the window it landed | `packages/automations/src/engine.test.ts` — `arms nothing when a take-back lands between the resume read and the grant` (voids the row from inside the replay, then asserts: no grant, the run reports it is asking again on the new approval, the take-back marker intact). Non-vacuous: pre-fix `expected [ { …(6) } ] to have a length of +0` |
| 2 DOCUMENT | Two residual windows | Written down at the code that carries them plus the changeset's known-limits paragraph: (a) an approval **row** has no guarded write — the routed store exposes `RecordStore.atomic` for `vendo_threads` / `vendo_apps` / generic rows only, so every marker on an approval is `get` then `put`. Because the transition winner is settled *before* any row write, a lost race costs a stale marker, never an execution (an erase inside the window can be undone by a re-put, but the held receipt leaves the resurrected row unspendable). (b) A custom `Guard` without the optional `spendApproval` puts the automations mint back on read-then-write, where a revoke in the window can lose to it | — |
| 3 SKIP | Concurrent-retry double audit line | Skipped as directed (cosmetic; the sequential invariant is tested) | — |

**Deviation, with evidence.** The instruction was to route this path through `spendApproval` like the other two callers. It cannot be: the replay **is** the spend, on the same `consumed:<id>` transition, and one approval cannot be spent twice. Tried literally, the guard refuses its own replay and the step re-parks forever — `away run park and resume > resumes one approved critical step and finishes successfully` fails with *"did not reach ok; last status was pending-approval"*. Moving the mint after the call instead fails the other way, because an away dispatch acts as the user only through captured authority (05 §6): *"stopped at send: away execution requires a captured grant"*. Mint → replay → take back on a lost replay is the same law (`taken-back`/`already-spent` arm nothing) expressed with the one spend this path actually has.

**Gates at this tip:** build 0 · typecheck 0 · lint 0 · test = the 7 known pre-existing failures only (`core/packaging.e2e` ×3, `store` drills ×2, `vendo dev-creds/model` ×2 — all from the space in this worktree's path). `automations-e2e` 35/35 and `redteam-e2e` 21/22 pass, this time in the parallel run as well.

## Checker round 7 — the throw window, and one honest comment (commit `1661767d9`)

| # | Item | Fix | Test |
| --- | --- | --- | --- |
| 1 FIX | An `executeCall` throw (store hiccup, a `VendoError` out of the registry) stranded the grant minted moments earlier — the same outcome round 6 existed to close, through the one exit it did not cover | `armed` hoisted out of the `try`; the `catch` takes the grant back before landing the run on its terminal row. Best-effort delete, so a failing delete cannot replace the real failure the run reports | `packages/automations/src/engine.test.ts` — `takes the grant back when the resumed dispatch throws` (grant deleted, run `error` with the thrown message, park dropped). Non-vacuous: without the catch-side delete, `expected [ { …(6) } ] to have a length of +0` |
| 2 HONESTY | The comment claimed a take-back leaves no authority "wherever in the window it landed" — false for a hard kill between the mint and the delete | Named at the site and in the changeset's known limits: the grant survives, nothing sweeps it, and what the person has instead is that it is visible in `grants.list`, pinned to the tool's `descriptorHash`, app-bound, away-only, and revocable. Closing it needs both writes in one transaction, which the store's record-at-a-time seam does not offer | — |
| 3 (optional) | `grants.revoke` semantics for the take-back | **Left as a raw delete**, and written down so it is not "fixed" by accident. The engine holds core's `Guard`, which has no grants surface, so a tombstone means hand-writing `revokedAt` plus an audit event — and a revoked-grant row for authority the person never actually held would read as history that did not happen. The raw delete is symmetric with the mint, which writes no audit event either | — |

**Gates at this tip:** build 0 · typecheck 0 · lint 0 · test = the 7 known pre-existing failures only.

One note on the first attempt at this run, because it looked alarming: an orphaned `next-server` from an earlier local run was holding the fixture host's `.next` state, and every suite that boots a fixture server failed with *"Server is not running"* / `ECONNREFUSED`. Reaping it fixed all of them, each verified alone first (`automations-e2e` 35 · `integration` 35 · `redteam` 21 · `mcp-e2e` 19 · `demo-bank` 104 · `demo-accounting` 156) and then together in a clean full run. No code involved.

## Merged `main` (rebuild cutover #736) — commit `cfec662ea`

846 files moved under this branch. 14 textual conflicts; the work was semantic.

| File | What main had done | What this takes |
| --- | --- | --- |
| `core/guard.ts`, `core/audit.ts` | added `"org"` to the ask/block `decidedBy` unions (org strictness clamp) | union of both: `confirmEach` + `denied` + `org` |
| `core/descriptor-hash.ts` | added `title` to the hash preimage | kept, on top of the `critical` → `confirmEach` rename |
| `core/tools.ts` | added `ASK_USER_TOOL`, the `vendoAuthored` provenance symbol, and `bindingRisk` (its own protocol-fact carrier) | both sides kept; `bindingRisk` complements D2 rather than colliding with it |
| `core/grant-sets.ts` | landed `mechanicalRisk` / `resolvedRisk` — a SECOND, name-based vote at decision time | `RANK` gains `ungraded` **above** every real grade, so the vote escalates a label someone assigned but can never launder "nobody graded this" into a guessed `write` |
| `core/harness.ts` | new harness venue whose `ToolListing.risk` was a hardcoded `read\|write\|destructive` | widened to `RiskLabel` — the new venue's model is TOLD a tool is ungraded instead of handed a grade we invented |
| `guard/src/guard.ts` | §12's unattended gate, keyed on `resolvedRisk === "destructive"` | keyed on the new `withheldFromUnattended`, which covers `ungraded` too — **see the regression below** |
| `actions/sync/common.ts` | replaced its local word lists with imports of core's | both aliases die (D1); `words()` stays, for tool NAMING, which concludes nothing about risk |
| `apps/generation/validation/{actions,capability-substitution}.ts` | **deleted** by the conductor rewrite; nothing reads `isMutatingRisk` / `fabricatedOperands` / `SUBMIT_LABEL` any more | our patches die with the surface. The replacement lane gates on a read-only ALLOWLIST, so `ungraded` is withheld there by construction |
| `ui/chrome/adoption-card.tsx` | per-tool word table (`Reads` / `Changes`) | gains `ungraded` → "Not reviewed", the approval chip's wording |
| `examples/demo-bank/.vendo/tools.json` | added a route, textual merge | re-synced with the structural extractor instead of hand-merging JSON — which **caught one tool the merge left graded `write`** by the deleted word vote |
| `docs-site/{capabilities/compound-tools,reference/cli}.mdx` | new risk-override paragraph / reworded plan section | ours (`confirmEach`) plus main's new prose |

### ⚠️ A real regression the merge opened, found and closed

With extraction no longer guessing from names, Maple's `host_transferMoney` reads
`ungraded` — so §12's "destructive actions are never available without a person"
went **blind**, the enable-time standing grant authorized the call, and an
unattended automation **transferred money**. Proved, not theorised: the away
drill came back `status: "ok"`.

Closed by extending §12's own gate to the state that says "nobody knows":
`withheldFromUnattended` refuses `ungraded` exactly as it refuses `destructive`,
at both the projection filter and the call-time gate. The two laws agree — §12
keeps irreversible actions off an unattended run, D3 says an ungraded tool needs
a person, and an unattended venue has none to ask. Park-and-resume survives: an
UNGRANTED ungraded step still parks, and the approved replay stays exempt, so
what can no longer happen is a standing grant quietly running an unjudged tool
with nobody watching.

### Flagged for your call (not decided here)

1. **D1 vs the second mechanical vote.** D1: "no code path may conclude anything
   from a tool's *name*." Main's `mechanicalRisk` does — at decision time, only to
   escalate, and skipped for Vendo-authored tools. Both laws now coexist and
   neither is silently disarmed, but which survives is the spec author's call.
   `protocol-facts.test.ts` keeps its repo-wide sweep and NAMES that one
   exception, so if the vote goes the test tells whoever removes it.
2. **`isBundleEligible`** still bundles `ungraded` tools into one consent card
   (rendered "Not reviewed"). Excluding them would mean one card per tool on an
   unjudged catalog — a consent-UX tradeoff, so it is left as main wrote it.
3. **Unjudged catalogs and automations.** Because unknown is now withheld
   unattended, a catalog nobody has graded offers automations nothing. That is the
   same fail-closed direction as everywhere else in this PR, in the one venue where
   asking is impossible — but it makes grading the catalog load-bearing rather than
   advisory.

### Verification on the merged tree

- Gates: build 0 · typecheck 0 · lint 0 · test = the 7 known pre-existing failures only.
- Suites alone: `guard` 255 · `core` 968 (+3 known) · `store` 379 (+2 known) · `actions` 601 · `apps` 735 · `automations` 97 · `vendo` 2039 (+2 known) · `demo-bank` 110 · `demo-accounting` 156 · `integration` 35 · `redteam` 21 · `mcp-e2e` 19 · `automations-e2e` 35.
- Corpus, from a space-free export of the merged tree: `express-host` **Layer 1 structural PASS 7/7**, **Layer 2 scored PASS 10/10**.
- E2E 1 — keyless `vendo sync` on Maple: catalog all-`ungraded` (25/25); `vendo doctor` still reports it (`E-TOOLS-003`, now `1/25` because the committed `judgments.json` grades the rest at runtime).
- E2E 2 — through each demo's real `.vendo/` composition: `host_listAccounts` read → run · `host_listClients` (Cadence's re-enabled end-user tool) read → run · `host_transferMoney` destructive + confirmEach → **ask (confirmEach)** · `host_sendClientMessage` / `host_setDocumentStatus` → ask (confirmEach). Both away drills 4/4, with §12 refusing the money and the message unattended.

## Changed test files, and why

Every one of these asserted a word-list grade or the old field name. The spec says these legitimately change; nothing unrelated was weakened.

**Grades moved from word lists to protocol facts**
- `packages/actions/src/connectors/composio-risk.test.ts` — rewritten: slug-verb cases replaced with hint-only + ungraded.
- `packages/actions/src/connectors/connectors.test.ts` — untagged connector tools are ungraded, not read/write.
- `packages/actions/src/sync/route-scan.test.ts` — 24 `risk: "write"` → `"ungraded"`; DELETE stays destructive.
- `packages/actions/src/sync/trpc.test.ts`, `graphql.test.ts` — mutations write, queries ungraded, names irrelevant.
- `packages/actions/src/sync/server-actions.test.ts` — every enabled server action is ungraded.
- `packages/actions/src/sync/sync.fixture.test.ts` — the fixture host's grade map; only `host_deleteInvoice` stays destructive.
- `packages/vendo/src/cloud-tools.test.ts` — the Cloud gmail fixture carries no hint → ungraded.
- `packages/vendo/src/cli/try/extract.test.ts` — try's read-downgrade now keys off `ungraded`; the "never downgrade a real grade" case is built from a DELETE route instead of a delete-shaped name.
- `fixtures/integration/src/actions-sync.e2e.test.ts` — synced grades are ungraded; the fixture policy gains an `ungraded → run` rule.
- `fixtures/integration/.vendo/policy.json` — same rule, so the connections/e2e suites keep testing composition rather than grading.
- `tools/genui-bench/fixtures/fixtures.test.ts` — the determinism test picked a tool by `risk === "read"`; now picks by name, since these catalogs are extracted.
- `corpus/harness/src/layers/scored.test.ts` — fixture catalogs carry protocol-fact grades; the unsafe arms keep their `read`-on-a-mutation defect.

**`critical` → `confirmEach` rename**
- `packages/core/src/{schemas,conformance/index,contract-coverage.e2e,descriptor-hash,packaging.e2e,security/descriptor-hash-adversarial}.test.ts` — field, `decidedBy`, and the hash preimage's canonical strings.
- `packages/core/vectors/descriptor-hash.json` — regenerated (preimage key follows the field), plus a new `ungraded` vector.
- `packages/guard/test/{app-venue-reads,breakers-lifecycle.e2e,conformance,park-resume,pipeline-conformance,surface}.test.ts`, `test/fixtures/tools.ts`, `test/security/approval-replay.test.ts` — field + `decidedBy` + fixture names.
- `packages/guard/test/security/critical-unsuppressible.test.ts` → `confirm-each-unsuppressible.test.ts` (renamed, semantics unchanged).
- `packages/actions/src/{formats,judgments,runtime/compound,runtime/conformance,runtime/registry,sync/sync.synthetic}.test.ts`, `packages/apps/src/{conformance,secret-exposure,testing/fixtures}.test.ts`, `packages/automations/src/engine.test.ts`, `packages/ui/test/voice/stage.test.tsx`, `packages/vendo/src/{compound.e2e,refine,cli/judge/pass,cli/judge/prompts}.test.ts`, `fixtures/automations-e2e/src/park-resume.e2e.test.ts`, `corpus/harness/src/ai/{score,expectations,matrix}.test.ts`.
- `packages/vendo/src/cli/doctor-codes.test.ts` — append-only registry snapshot updated for `E-TOOLS-003`.

**Round-1 checker fixes**
- `packages/actions/src/sync/{route-scan,server-actions,trpc,graphql}.test.ts` — unclassifiable/subscription fixtures now expect `ungraded` + `disabled`; a real `DELETE` still asserts `destructive`.
- `packages/actions/src/sync/{sync.fixture,sync.synthetic}.test.ts` — the same expectation; the hand-authored `overrides.json` destructive case is untouched.

**Round-2 checker fixes**
- `packages/guard/test/park-resume.test.ts` — `denials notify false, do not resume…` asserted the OLD re-park-forever behavior; it now asserts the denial answers the identical re-issue (`status: "blocked"`, nothing executed). That behavior IS the bug, so the assertion legitimately flips.
- `packages/guard/test/ungraded-default.test.ts` — 7 cases added across both rounds (presets, write budget, 3 standing-denial, 2 cross-session replay).
- `packages/vendo/src/ungraded-query-approve.e2e.test.ts` — the denied case asserts durability across 3 reopens with an empty approval queue.

**Round-3 checker fixes**
- `packages/guard/test/ungraded-default.test.ts` — 6 cases added (TTL sweep, abandonment, revoke + its guards, both decision orderings).
- `packages/core/src/conformance/index.test.ts` — the reserved-projection assertions now include the derived `call` ref, plus a new `round-trips a denial's provenance and its taken-back marker` and a rejection case for a bogus `deniedBy`.
- `packages/guard/test/fixtures/pglite-store.ts` — the real-SQL fixture gains the three columns and the `call` ref mapping it mirrors from the routed store.
- `packages/apps/src/generation-ungraded-laws.test.ts` — schema-valued `additionalProperties` case.
- `packages/ui/test/chrome/activity-semantics.test.ts` — `decidedByLabel` + the revoke outcome label.
- `packages/vendo/src/ungraded-query-approve.e2e.test.ts` — the un-deny journey end to end.

**Round-4 checker fixes**
- `packages/guard/test/ungraded-default.test.ts` — 4 cases added (both void/replay interleavings through a store that interleaves a second operation mid-call, revoke-after-spend, double-revoke audit count).
- `packages/vendo/src/byo-approvals.e2e.test.ts` — one case added for the `abandonApprovals`-less expiry path; `harness()` gained an option that hides the optional method.
- `packages/store/src/helpers.test.ts` — `decides only pending approvals…` dropped its `decide` assertions and was renamed, because that helper is deleted; the pending-ordering and row-read coverage it also carried stays.
- `packages/store/src/routing.test.ts` — the approvals case now also asserts a voided row's `updatedAt`.
- `packages/ui/test/chrome/activity-semantics.test.ts` — one case for the new too-late outcome label.

**Round-5 checker fixes**
- `packages/guard/test/ungraded-default.test.ts` — 5 cases added (erasure mid-replay, failed-void retry, batch-compensation fence, the spend seam both ways); the interleaving store helper gained claim interception and a one-write-fails wrapper.
- `packages/automations/src/engine.test.ts` — 4 cases added; `GuardDouble` gained a scripted `spendApproval` (left unset by default, so every pre-existing case still exercises the fallback path).

**Round-6 checker fixes**
- `packages/automations/src/engine.test.ts` — 1 case added for the resume window; nothing else touched.

**Round-7 checker fixes**
- `packages/automations/src/engine.test.ts` — 1 case added for a throwing resume; nothing else touched.

**Merging `main` (rebuild cutover)**
- `fixtures/integration-browser/e2e/connections.spec.ts` — "connect-required → connect → **automatic retry executes**" asserted pre-redesign semantics. The fixture's Composio tool ships no upstream destructive/read hint, so it is `ungraded`, and nothing may guess a grade from its name (D1/D3): the retry now ASKS. Rewritten (and renamed) to walk the new truth — connect → the retry parks → the approval card carries the ungraded chip ("Not reviewed") and the real payload → Approve → "Sent the email." renders. This is a behavioral change per the spec, at the seam where the old behavior read worst: "connect, and it sends your email on its own." No deny sibling: the harness's scripted model is a FIFO that streams the post-tool text whatever the outcome (verified — a denied call still rendered its follow-up), and the Composio stub records no executions, so the browser has no observable separating sent from not-sent; the guard suite already pins deny → nothing runs.
- `packages/actions/src/sync/risk-vocabulary.test.ts` — DELETED. It asserted that extraction reads core's verb vocabulary, which is exactly the behavior D1 removes; main added it while this branch was open.
- `packages/actions/src/sync/protocol-facts.test.ts` — the repo-wide sweep now names the one remaining exception (core's decision-time vote) rather than pretending it is absent, plus a second sweep that keeps extraction and the connectors clean.
- `packages/apps/src/generation-ungraded-laws.test.ts` — DELETED with the two validation modules it imported, which main's conductor rewrite removed.
- `packages/guard/test/pipeline-conformance.test.ts` — stage list is the union (`confirmEach` … `org`).
- `packages/actions/src/sync/route-scan.test.ts` — kept this branch's `ungraded` expectations; main's comment described the word-list vote that no longer runs at extraction.
- `examples/demo-{bank,accounting}/src/vendo/away-drill.test.ts` — the drills now hand the registry `.vendo/overrides.json`, as the real app does. Reading the raw catalog was invisible while extraction guessed grades; with grades coming from a person, the judge, or a protocol fact, the file has to be loaded for the drill to see what the app sees.

**New across the checker rounds**
- `packages/vendo/src/ungraded-query-approve.e2e.test.ts` (real guard + apps + store, through `apps.open()`), `packages/apps/src/generation-ungraded-laws.test.ts`.

**New**
- `packages/guard/test/ungraded-default.test.ts`, `packages/actions/src/sync/protocol-facts.test.ts`, `packages/actions/src/critical-alias.test.ts`.

## Deviations

1. **There is no `vendo status` command** (spec now amended in place to say so). The spec and the brief both name one; the CLI has `try / init / login / doctor / sync / eject / knowledge / mcp / cloud / config`. I put the count on `vendo doctor` (a real check with a real error code and verify anchor) and left the consequence line on keyless `init`/`sync`, rather than inventing a command the spec's out-of-scope list would not have wanted.
2. **`descriptorHash` follows the rename** (the spec does not name the hash). Keeping `critical` in the preimage while the field is `confirmEach` would be a hidden inconsistency any reimplementer would get wrong. Cost is a one-time re-ask on in-flight approvals; called out in the changeset.
3. **Judge/override `risk` still accepts all four values** rather than a graded-only enum. `gradedRiskLabelSchema` is exported from core and used where it bites — the judge's own proposal schema, which may not answer "I don't know".
4. **`ungraded` ranks above the real grades** in the compound step-max and refine's ordering (a compound with an ungraded step does not know its own risk), and is handled explicitly in `classifyField` rather than by rank: grading a blank tool is the grading, not a loosening, so a judge grade lands; trading a real grade back for a blank still queues for a human.
5. **The corpus extraction-layer scoring was re-based**, not just re-baselined: `tools.precision`/`recall` key on surface identity alone (extraction no longer claims read vs write), and `annotations.match` now asserts the protocol fact. Risk accuracy against curator ground truth lives in the AI lane's `ai.risk.accuracy`, which is unchanged in kind.
6. **Demo catalogs left ungraded** — see the breaking-behavior note above. Your call.

---
## Summary by cubic
Risk grading now uses only protocol facts, never tool names. `ungraded` tools ask by default, `critical` is now `confirmEach`, and approvals/denials are durable and race-safe; merged the rebuild cutover so unattended runs also gate `ungraded` tools.

- **Bug Fixes**
  - Linearized replay vs. revoke with a single‑use receipt: either the replay runs or the take‑back wins; losers report conflict. Missing approval rows abort consumption.
  - Denials track provenance (`deniedBy: "human" | "system"`). Only human denials stand; you can revoke them; older unconsumed approvals are voided; the ledger shows “previously denied.”
  - Automations spend approvals via a `spendApproval` seam; parked resumes take grants back if the replay loses or if a resume throws; the fallback preserves `deniedBy`/`voidedAt` and arms nothing after a revoke.
  - Recovery fences: batch compensation skips consumed/voided/erased rows; receipts land before row writes; mixed‑version receipts read as spent; BYO expiry denies then immediately revokes so timeouts don’t plant a standing no.
  - Unattended runs withhold `ungraded` (and destructive) tools via `withheldFromUnattended`, closing the loophole where an ungraded step could run away.
  - Browser integration: connector retry now asks before sending when the tool is `ungraded`; the test verifies “Not reviewed” with real inputs, then Approve → send.

- **Refactors**
  - Merged the rebuild cutover: kept main’s `mechanicalRisk` vote but rank `ungraded` above all grades so it can’t be laundered; presets treat `ungraded` like destructive, harness/agent order puts `ungraded` last, extraction/connectors stay protocol‑fact only, UI shows “Not reviewed,” demo catalogs judged and drills now load `.vendo/overrides.json`.

<sup>Written for commit d53e3e53cd390b53f5044baae0cebc331824e085. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/747?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Decision recorded (Yousef, 2026-08-03): on the open question about the name-based escalate-only `mechanicalRisk` vote — **option (a), exempt human grades.** A human's `overrides.json` grade always wins; the vote survives only where no human has graded. Rationale: it matches the categorical ruling in this PR that no code path may conclude anything from a tool's name, and D2's human-always-wins precedence.

Not yet implemented. Context for whoever picks it up: the vote reaches consumer-facing copy on one path — `automations/adoption.ts:54` → AdoptionCard — found while landing #777 (the UI redesign), where the same class of name-based inference was deleted from the consent-card wording per this ruling.

PR #749, opened 2026-08-03
fix(cli): demo themes follow their CSS; init never regenerates existing files

yousefh409:
Two founder-decided fixes, both final.

## Fix 1 — demo themes: the CSS is the truth

Maple (`examples/demo-bank`) and `examples/demo-template` carried
`.vendo/theme.json` values that disagreed with their own CSS, so every build
warned once #742 made sync re-extract the theme. Ruling: the CSS is correct,
the committed theme was stale. I let the tool do it —
`vendo sync --no-ai --theme-refresh` in each app — and committed each app's
`.vendo/theme.extracted.json` merge base so a later sync is a no-op. No CSS was
touched.

```diff
--- a/examples/demo-bank/.vendo/theme.json
+++ b/examples/demo-bank/.vendo/theme.json
-    "border": "#E2E1DE"
+    "border": "#ecebe8"
-    "fontFamily": "system-ui, -apple-system, Segoe UI, Roboto, sans-serif",
+    "fontFamily": "Inter, sans-serif",

--- a/examples/demo-template/.vendo/theme.json
+++ b/examples/demo-template/.vendo/theme.json
-    "fontFamily": "system-ui, -apple-system, Segoe UI, Roboto, sans-serif",
+    "fontFamily": "Inter, sans-serif",
```

Only the stale slots moved. (demo-template's `border` already agreed with its
CSS case-insensitively, so it stayed put.)

### Proof — no warning, and the tree stays clean on a second run

```
$ cd examples/demo-bank && vendo sync --no-ai
tools: +0 -0 ~0
pins: 0 captured, 0 drifted

$ cd examples/demo-template && vendo sync --no-ai
tools: +0 -0 ~0
pins: 0 captured, 0 drifted
catalog.json: 0 discovered, 0 registered

$ cd examples/demo-accounting && vendo sync --no-ai
tools: +0 -0 ~0
pins: 0 captured, 0 drifted

$ git status --porcelain
(nothing but this branch's own changes — the syncs wrote no files)
```

A full `pnpm build` (which runs each demo's `prebuild: vendo sync --strict --no-ai`)
also leaves the tree clean.

## Fix 2 — init never regenerates files in your source tree

#742 stopped init rewriting `app/layout.tsx`. This extends the same law to the
last two cases:

- `app/api/vendo/[...vendo]/route.ts` — was being wired with `serverActions`
  when actions appeared after the route was generated
- `app/api/vendo/[...vendo]/vendo-actions.ts` — was being regenerated whenever
  the detected `"use server"` surface moved

Init still **creates** both. It never writes either again. When it finds a
change it will not make it prints it in the same framed block #742 introduced
for the mount, and carries it in `--agent` as `edits[]` of `{file, lines, why}`
alongside `mount`, plus `manualSteps` and the agent tail.

`package.json` script hooks are untouched — Vendo-owned config, out of scope.
`wireRouteServerActions`'s rewrite machinery is deleted; what survives is a
detector that returns the paste (`routeServerActionsEdit`).

### doctor: new E-WIRE-009

Doctor covered nothing here — a skipped paste left server-action tools failing
closed at execution time with every other check green. New
`wiring/server-actions` check: silent when the host has no server actions,
otherwise it fails when the map is missing entries init would have added or the
route never passes `serverActions`. Anchored on the verify page.

### Real run — scratch Next host, `vendo init` twice

Host: Next 16 + `app/actions/invoices.ts` (`"use server"`). Run 1 created
`route.ts` and `vendo-actions.ts` correctly. I then aged the host the way a
real one ages — renamed the exported action, and hand-removed the
`serverActions` wiring from `route.ts` (simulating a route that predates the
actions) — and ran init again.

Second run:

```
3 STEPS LEFT — paste these yourself (init never edits your files)

  File: app/layout.tsx

  File: app/api/vendo/[...vendo]/vendo-actions.ts
    --- a/app/api/vendo/[...vendo]/vendo-actions.ts
    +++ b/app/api/vendo/[...vendo]/vendo-actions.ts
    -import { createInvoice as action0 } from "../../../actions/invoices";
    -
    -export const serverActions = {
    -  "app/actions/invoices.ts#createInvoice": action0,
    -};
    +import { createDraftInvoice as action0 } from "../../../actions/invoices";
    +
    +export const serverActions = {
    +  "app/actions/invoices.ts#createDraftInvoice": action0,
    +};

  The "use server" surface moved since this map was generated — an action missing from it fails closed at execution time (no work performed).

  File: app/api/vendo/[...vendo]/route.ts
    import { serverActions } from "./vendo-actions";
    … then add inside createVendo({ … }): serverActions,

  createVendo dispatches server-action tools through that map — without it every one of them fails closed at execution time (no work performed).

Agent tail:
  edit app/layout.tsx — wrap the app in the <VendoRoot> lines above (…)
  edit app/api/vendo/[...vendo]/vendo-actions.ts — apply the change printed above yourself (it already exists, so init did not write it)
  edit app/api/vendo/[...vendo]/route.ts — apply the change printed above yourself (it already exists, so init did not write it)
```

(The diff header comment lines are elided above for length; the real output
carries the whole file.)

shasum before/after that second run:

```
$ diff before.txt after.txt
9c9
< b187736ba9b9df00730534c8098eceaf9d92ec42  ./.vendo/tools.json
> fdcb4af3b96e4247e3c07082d0b9deffcf2d900d  ./.vendo/tools.json
```

One file moved, and it is not in the user's source tree: `.vendo/tools.json` is
the machine catalog, and it correctly tracked the renamed action. Every file
under `app/` and `vendo/` is byte-identical:

```
b187736…  ./.vendo/tools.json            (machine catalog — tracked the rename)
c340ee29…  ./app/actions/invoices.ts      unchanged
bd121a15…  ./app/api/vendo/[...vendo]/route.ts            unchanged
f7e29376…  ./app/api/vendo/[...vendo]/vendo-actions.ts    unchanged
9a183ead…  ./app/layout.tsx               unchanged
aa706db1…  ./app/page.tsx                 unchanged
4160cec1…  ./package.json                 unchanged
82baffe6…  ./vendo/registry.tsx           unchanged
c44da05e…  ./vendo/vendo-root.tsx         unchanged
```

A third run is byte-identical to the second across every file, `.vendo/`
included — fully idempotent.

Doctor on that host, before applying the pastes:

```
ok       wiring/next-route — catch-all handler is wired
broken   wiring/server-actions — server actions fail closed — app/api/vendo/[...vendo]/vendo-actions.ts is missing app/actions/invoices.ts#createDraftInvoice; app/api/vendo/[...vendo]/route.ts does not pass serverActions to createVendo. Re-run `npx vendo init`: it prints the exact paste for each (it never rewrites a file you already have).
```

After applying both pastes by hand:

```
ok — 1 server action is registered and wired
$ vendo init --yes --byo --no-ai
Already wired — nothing to change.
  File: app/layout.tsx
```

## Tests

`packages/vendo/src/cli/init.test.ts`
- `generates the server-action registration map and the wired route on a fresh install (ENG-248)` — fresh install still creates both correctly
- `never regenerates an existing route.ts or vendo-actions.ts — it prints the paste instead` — both files byte-identical across re-runs; the paste is emitted for each; `3 STEPS LEFT` framing
- `carries the pastes it will not write into the --agent plan` — `edits[]` shape, and the same lines in `manualSteps`

`packages/vendo/src/cli/doctor.test.ts`
- `fails E-WIRE-009 when detected server actions are neither registered nor wired`
- `passes wiring/server-actions once the map registers the action and the route passes it`
- `says nothing about server actions in a host that has none`

`packages/vendo/src/cli/doctor-codes.test.ts` — append-only registry snapshot
extended with E-WIRE-009 (the snapshot is designed to be extended by a new
code); `doctor-codes.docs.test.ts` keeps the verify-page anchor 1:1.

## Gates

`pnpm build && pnpm test && pnpm typecheck && pnpm lint` all green from the
repo root (1,628 tests passing in `@vendoai/vendo`, 53/53 turbo tasks).

Branched from `cli/init-sync-coherence` while #742 was open, rebased onto `main`
after it merged.

Docs updated: `agents/verify.mdx` (E-WIRE-009 section), `connect/api-tools.mdx`,
`connect/vendo-init.mdx`. Two changesets.

No files owned by the `cli/host-component-capture` lane (`packages/actions/src/sync/pins.ts`,
the catalog) were touched.

---
## Summary by cubic
Aligns demo theme configs with their CSS to remove build warnings, and makes `vendo init` create-only while adding a precise check for server-action wiring that prints pastes instead of rewriting files.

- **Bug Fixes**
  - Demo themes: updated `.vendo/theme.json` (`border`, `fontFamily`) in `examples/demo-bank` and `examples/demo-template`, and committed `.vendo/theme.extracted.json` so `vendo sync` is a no-op.
  - Init behavior: `app/api/vendo/[...vendo]/route.ts` and `vendo-actions.ts` are only created once; existing files are never rewritten. When wiring is missing, init prints the exact lines and carries them in `--agent` as `edits[]`.
  - Server actions: init and doctor share one wiring model. An import line alone is not wiring; the check looks for `serverActions` inside `createVendo({ … })`. Existing maps are compared by registered keys (not bytes); missing entries are printed only. Disabled tools in `.vendo/overrides.json` (or judgments) are ignored. No generated map is created or required when a route composes its own.

- **Migration**
  - If your server actions changed after the first init: run `npx vendo init`, apply the printed changes to `route.ts` and/or `vendo-actions.ts`, then run `npx vendo doctor`.

<sup>Written for commit 575d29f6b719721b90895101fb043d03b122282b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/749?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Review fixes pushed in `575d29f6b`. All four reproduced first, then fixed, then re-proven live against fresh scratch Next hosts. Gates green from the repo root.

## BLOCKER 1 — one shared predicate, scoped to the call

You were right, and right about why it matters: the printed paste's second line is prose, so a half-applied paste lands exactly there. Doctor grepped the whole file; init sliced from `createVendo(`.

There is now one exported helper in `init-scaffolds.ts` returning a tri-state, and both sides read it:

```ts
export type ServerActionsWiring = "wired" | "unwired" | "unknown";

export function serverActionsWiring(source: string): ServerActionsWiring {
  const call = source.match(/createVendo\(\s*\{/);
  if (call === null) return "unknown";
  return /(^|[\s{,])serverActions\b/.test(source.slice(source.indexOf(call[0]))) ? "wired" : "unwired";
}
```

The third state matters for agreement: init declines to name a paste for an unrecognized composition, so doctor must decline to grade it. Init raises the wiring paste on `"unwired"` and doctor fails on `"unwired"` — the same word, the same function.

Live, route with the import and no `serverActions,` in the call:

```
route.ts now:
2:import { serverActions } from "./vendo-actions";
5:const vendo = createVendo({
--- doctor:
broken — server actions fail closed — app/api/vendo/[...vendo]/route.ts does not pass serverActions inside createVendo({ … }). Re-run `npx vendo init`: …
```

It also stops accusing the map, which is complete.

## BLOCKER 2 — the map is compared by keys, never bytes

Fixed as you specified: an edit is raised only when a required `"<module>#<export>"` key is absent, and only the missing entries print.

Live, a host carrying the **previous release's** header plus two hand edits (an added comment, and the alias renamed `action0` → `handler`), surface unchanged:

```
--- init (surface unchanged):
  File: app/layout.tsx
--- map unchanged: YES
--- doctor:
ok — 1 server action is registered and wired
```

No map paste at all. The `connect/api-tools.mdx` promise now holds, and the "yours from here" header is true — your formatting, comments, aliases and extra entries all survive.

This also closes the whole-file-diff nit. A genuinely missing action prints three lines, with the alias continuing the file's own numbering:

```
  File: app/api/vendo/[...vendo]/vendo-actions.ts
    import { createDraftInvoice as action1 } from "../../../actions/invoices";
    … then add inside the serverActions map:
      "app/actions/invoices.ts#createDraftInvoice": action1,

  1 action the host exposes is not registered here — it fails closed at execution time (no work performed). The rest of the file is yours; nothing else needs to change.
```

## IMPORTANT 3 — a route with its own map

Doctor is silent when the route is `"wired"` and does not import `./vendo-actions` — it has no generated map to grade and must not guess.

**One thing beyond what you flagged.** Fixing the doctor half exposed that *init* was still dropping an orphan `vendo-actions.ts` beside such a route — a file that composition will never import. That is the same disagreement from the other side, and the registry branch immediately above already has the rule, so I mirrored it: the map is created only for a route that will consume it (being created this run, already importing `./vendo-actions`, or `"unwired"` and about to be handed the import paste). Flagging it since it is scope I added.

```
--- init:
  ~ package.json
  File: app/layout.tsx
--- route unchanged: YES
--- orphan vendo-actions.ts created? NO — correct
--- doctor:
(no wiring/server-actions check emitted)
```

## IMPORTANT 4 — overrides honored, on both sides

Done via a single shared resolver, `requiredServerActions(root)`, used by init and doctor. It applies the same three-layer stack the runtime resolves — skeleton ⊕ judgments ⊕ overrides — so judgments are covered too.

**One deliberate departure from your instruction**, worth your call. You said to source bindings from the already-synced `.vendo/tools.json`. I source the *surface* from live extraction and apply the *disable* layers from `.vendo/overrides.json` + `judgments.json` on top. Reason: `buildPlan` runs **before** this run's sync writes `tools.json`, so reading it would put init a full run behind — add an action, run init, hear nothing; run init again, hear about it. The disable layers are what carry the human decision, and they are on disk already. Net effect is what you asked for (nothing disabled is ever demanded, both sides agree), without the lag. Say the word if you want tools.json anyway.

Live, second action added then disabled:

```
--- BEFORE disabling — init + doctor:
  File: app/api/vendo/[...vendo]/vendo-actions.ts
    import { internalSweep as action1 } from "../../../actions/internal";
      "app/actions/internal.ts#internalSweep": action1,
broken — server actions fail closed — app/api/vendo/[...vendo]/vendo-actions.ts does not register app/actions/internal.ts#internalSweep. …
--- disabling tool(s): host_internal_sweep
--- AFTER disabling — init:
(init asks for nothing in the app dir)
--- map unchanged: YES
--- doctor:
ok — 1 server action is registered and wired
```

## Nits

- **`.env.example` carve-out** — named in the header law, with the fact that it appends and never rewrites a line.
- **Whole-file diff** — gone, as above.
- **Hex case in `demo-bank/theme.json`** — left as the extractor wrote it (`#ecebe8` is what the CSS says). Uppercasing would be stable (`sameValue` is case-insensitive, so sync stays a no-op) but it re-introduces a hand edit to a machine-owned slot for cosmetics, against the "let the tool do it" ruling. Happy to flip it if you'd rather have the file internally consistent.

## CI flake

Confirmed a flake. `pnpm test` from the repo root: **53/53 tasks, 1,633 passing**. The suite on its own:

```
$ pnpm --filter @vendoai/actions exec vitest run src/runtime/fixture.e2e.test.ts
 Test Files  1 passed (1)
      Tests  7 passed (7)
   Duration  3.71s
```

3.7s against a 120s hook timeout — nowhere near the boundary under normal load.

## Regression tests

`init.test.ts`
- `says nothing about a map whose surface is unchanged, however far its text has drifted` — previous-release header + a comment + a renamed alias → no paste, file byte-identical
- `does not demand registration of an action disabled in overrides.json` — second action added and disabled → silence, map untouched
- `never regenerates an existing route.ts or vendo-actions.ts` — updated to assert the targeted entries and `not.toContain("--- a/")`

`doctor.test.ts`
- `fails E-WIRE-009 when the route imports the map but never passes it to createVendo` — also asserts the complete map is *not* accused
- `stays silent when the route passes a map it composes itself`
- `stays silent about an action disabled in .vendo/overrides.json`

## End-to-end idempotency, re-proven on the final build

Aged host (action renamed, wiring hand-removed), `vendo init` three times:

```
--- shasum diff (run 2):
9c9
< b187736…  ./.vendo/tools.json
> fdcb4af…  ./.vendo/tools.json
--- run 3:
run2 == run3: fully idempotent, every file
```

One file moves, and it is the machine catalog tracking the rename — every file under `app/` and `vendo/` is byte-identical. Run 3 is identical to run 2 across everything, `.vendo/` included.

## Gates

`pnpm build && pnpm test && pnpm typecheck && pnpm lint` green from the repo root. Fix 1 untouched; the three demos still sync clean with an empty tree.

Not merged — yours to proof.

yousefh409:
## Live proof — all four review scenarios pass

Run against fresh scratch Next hosts with the CLI built from `575d29f6b`.

| Scenario | Required | Result |
|---|---|---|
| Route imports `serverActions` but never passes it into `createVendo` (the half-applied paste) | doctor red | **broken E-WIRE-009** ✅ |
| Previous-release header + unchanged action surface | init prints nothing | **0 paste-related lines** ✅ |
| Host passes its own map, no `vendo-actions.ts` | doctor silent | **check absent** ✅ |
| Action disabled in `.vendo/overrides.json` (`host_do_thing`) | doctor silent | **check absent** ✅ |

Note on the fourth: my first attempt reported a false failure because I hand-wrote an `overrides.json` without its `format` field, so schema parsing rejected it and no override applied — a defect in my test, not the code. With a valid file and the real tool name, doctor is correctly silent.

Confirming the judgment call raised in the fix comment: **sourcing the surface from live extraction and layering `overrides.json` + `judgments.json` on top is right.** Reading `tools.json` would put init a full run behind its own sync, which is a worse failure than the one it avoids.

Hex case in demo-bank: leave it as the extractor wrote it. Uppercasing for looks would re-introduce a hand edit to a machine-owned slot, against the ruling that the tool owns those values.

Merging.

PR #750, opened 2026-08-03
feat(cli): capture registered host components so the console can render them

yousefh409:
Rebased onto `main` after the Rebuild cutover (`abc7a44a1`); the diff is only
this lane's work. Verified end to end in a real browser against a live Cloud
project: all three Cadence host components mount and draw in the console — 3
iframes, 0 console errors, 0 CSP refusals, 0 in-page mounts. That needed a
matching console-side re-vendor of `@vendoai/ui` (vendo-web #254), because a
capture can only render on a jail runtime that carries the packages it declares
in `requires`.

The console's Apps gallery renders each app live in a jail. Agent-generated
components render for real; the host's own registered components rendered as
grey labeled blocks, because Vendo stored only their **name**. Now sync stores
the component.

Gates green: `pnpm build && pnpm test && pnpm typecheck && pnpm lint`
(53/53 test tasks, 41/41 typecheck), plus a real run of the built CLI against
`examples/demo-bank` and a real push against a local store, both below.

---

## 1 — Capture source for every registered component

`.vendo/components/<Name>.json`, one per component in
`<VendoRoot components={…}>`, written by the same walk pins already use
(`captureClosure`, extracted from `captureSubSources` — pins call the new one,
nothing forked). Same never-clobber and degraded-scan rules: a degraded catalog
scan (no compiler, no tsconfig, a walk that hit its file cap) prunes nothing.

**Why that location.** It mirrors `.vendo/remixable/<slot>.json` — one file per
unit, stem = the id a consumer looks up — so the prune rule, the "presence on
disk, not parseability" rule, and the commit story are all the ones already in
the repo. It sits beside `catalog.json` rather than inside it because
`catalog.json` is a hand-editable generation input that a human curates copy in;
this is a machine artifact regenerated wholesale.

The record holds **no source**. That is load-bearing, not tidiness — see part 4.

```jsonc
// .vendo/components/MapleSpendingDonut.json  (real, from demo-bank)
{
  "name": "MapleSpendingDonut",
  "hash": "sha256:89e91a41…",            // content hash of the whole capture
  "capturedAt": "2026-08-03T03:49:34.896Z",
  "module": "src/vendo/registry.tsx",
  "export": "MapleSpendingDonut",        // the binding to render; absent = the module default
  "entry": "fbab51e4…",                  // content address of the entry module
  "modules": {                           // captured module id -> content address
    "src/components/charts/donut.tsx": "499fc969…",
    "src/lib/money.ts": "89b51d53…"
  },
  "styles": [{ "path": "src/app/globals.css", "ref": "9283964e…" }],
  "bytes": 20093
}
```

`export` exists because Maple registers `MapleSpendingDonut`, a function its
module **never exports**. Capture records the binding; `hostComponentEntrySource`
(exported from `@vendoai/actions`) appends `export { X as default }`. Capture
guarantees the append is legal — a module that already default-exports something
else is skipped with a warning rather than shipping code that will not compile.

## 2 — Byte budget replaces the depth cap

`MAX_SUB_SOURCE_DEPTH = 2` and the `beyond capture depth 2` warning are gone.
The walk now goes to closure and stops at two honest lines:

- **Package boundary.** Anything resolving into `node_modules` is never captured
  and never warned about — it isn't the host's code. A bare specifier with no
  tsconfig alias mapping it is a package (`isPackageSpecifier`), which is what
  separates "not yours" from "your import is broken"; the latter still warns.
- **Byte budget**, 256 KB of JS/TS per component. Over it, the capture is
  skipped entirely and the warning names the culprit:

  > `host component Heavy was not captured: its import closure is 402 KB, over the 256 KB per-component budget (largest: src/data/fixtures.ts) — split the component away from that module, or import it lazily, so the console can render it`

  The record still lands, carrying `skipped: { reason: "too-large", bytes,
  budgetBytes, largest }`, so the console shows an honest "too large to preview"
  tile instead of silently falling back to the old placeholder. The previously
  captured baseline is left byte-identical.

Shared by pins and components — one implementation, `sync/capture.ts`. The CSS
warning behaviour is unchanged. **Styles are outside the budget**: they are
app-root, identical for every component, and stored once, so charging each
component for `globals.css` would police the wrong thing.

**Test-file change, called out:** the existing pins test asserted the depth-2
cut. I updated it to assert the four-deep chain is now captured in full — the
behaviour change is the point of this part.

## 3 — Content-addressed dedupe

`.vendo/components/modules/<sha256hex>.json` = `{ source, imports? }`, keyed by
the sha-256 of its own canonical JSON. Two components that reach the same file
produce identical JSON, so they land on one key. Stylesheets use the same shape
and the same store.

Pruning reads the reference set back **off disk**, including records this run
never touched, so a module survives as long as any record still names it. Same
rule in Cloud, and blobs are only pruned after the record side fully reconciled.

`.vendo/remixable/` keeps its self-contained fat rows. `pinBaselineSchema` is
what the console and the OSS apps runtime read today, and re-pointing it at refs
is a separate contract change with no payoff here — the pin corpus is bounded by
`<Remixable>` slots, and the registry is where the N-copies problem actually
bites. Stated so it is a decision, not an oversight.

## 4 — Hash-manifest sync, and **no contract amendment was needed**

The deferred I3 asked for a hash projection on `records.list`. I did not add
one, because the frozen contract already has a keys-only door — and using it is
strictly better than a projection.

| half | where | manifest |
| --- | --- | --- |
| heavy (source bytes) | blob namespace `vendo_host_components`, key = content hash | `blobs.list()` → **keys only**, already in 01-core §12 and already served by the console |
| index (refs, ~400 B) | records collection `vendo_host_components`, id = name | listing the collection *is* the id→hash manifest |

Content addressing means "the key exists" and "the content is current" are the
same fact, so one keys-only call answers the whole question for the bytes. No
manifest row exists, so an out-of-band deletion is still seen and re-pushed —
the reason the previous lane rejected the row. And the index is small **by
construction**: making the collection thin is what makes deriving the hash list
from the collection itself affordable.

The same reshape is available for `vendo_pin_baselines` later; its fat-row list
survives only because those rows are the console's current read contract.

## What crosses the wire

The source of every registered component, the source of every module in its
import closure, and the app-root stylesheets. **Never**: package code, anything
outside the project root, env vars, data.

Because that widens the blast radius from "components you wrapped" to "your
whole registry", sync **asks once per project** before the first upload and
commits the answer to `.vendo/cloud.json` (`{ "pushComponents": true }`); every
later run is silent. This is project config, not user consent state — the
"never persist" rule from #742 was about the AI pass. `--push-components` /
`--no-push-components` answer it in CI and do **not** rewrite the file; a
non-interactive run with neither says so and pushes nothing. Keyless/BYO never
asks and never makes a request — there is a tripwire test whose fetch throws.

Documented in `docs-site/connect/host-components.mdx` ("What sync captures",
"What crosses the wire"), `docs-site/reference/cli.mdx`, and
`docs-site/reference/dot-vendo.mdx`.

## The wire shape the console must consume

Read from the project store, then inflate — this is the whole console-side
change, and it is mechanical:

```ts
const record = capturedHostComponentSchema.parse(row.data);   // @vendoai/actions
if (record.skipped) return <TooLargeTile … />;

const blob = (ref) => capturedModuleSchema.parse(JSON.parse(await blobs.get(ref)));
const entry = await blob(record.entry);

// exactly JailFurnishing (packages/ui/src/tree/jail/JailedComponent.tsx)
const furnishing = {
  sourceImports: entry.imports ?? {},
  subSources: mapValues(record.modules ?? {}, async (ref) => {
    const m = await blob(ref);
    return { source: m.source, imports: m.imports ?? {} };
  }),
  styles: (record.styles ?? []).map(async (s) => ({ path: s.path, css: (await blob(s.ref)).source })),
};
const source = hostComponentEntrySource(entry.source, record.export);
```

`AppDocPreview` (`apps/console/components/console/preview/app-doc-preview.tsx`)
builds `hostComponents[name] = hostStub(name)` for every `source === "host"`
node. It swaps `hostStub` for a `JailedComponent` fed the pair above; the stub
stays as the fallback for a component the corpus has no record for. Collection
and namespace are both `vendo_host_components`, read the same way `cloud.ts`
already reads `vendo_pin_baselines`.

**Verified by loading every committed capture through a replica of the real
jail loader** (sucrase `["typescript","jsx","imports"]` → `jailRequire`, per
`runtime-entry.tsx`). Review round 1 caught that my first check measured the
wrong thing and that 4 of 6 captures threw; see the review comment for the
before/after. Current state, both demos:

```
loads: 2   honest skips: 4   THROWS IN THE JAIL: 0
```

A capture either loads, or carries a `skipped` reason the console can show.
Nothing ships that would error-box.

⚠️ **Only 2 of 6 registered components in our own demos can be previewed
today** — the other four import `zod`, `recharts`, or `clsx`/`tailwind-merge`,
which the jail cannot resolve. This lane makes that honest and named; making it
*rare* needs the separate founder call on blessing packages in the jail.

## Real end-to-end — `examples/demo-bank` (Maple)

```console
$ node packages/vendo/dist/cli.js sync --no-ai --no-push-components .
tools: +0 -0 ~0
pins: 0 captured, 0 drifted
components: 3 captured, 0 updated

$ node packages/vendo/dist/cli.js sync --no-ai --no-push-components .   # again
components: 0 captured, 0 updated                                       # byte-stable
```

```
MapleNetWorthCard  -> src/components/home/net-worth-view.tsx | export: default            | modules: 0 | 10,697 B
MapleSparkline     -> src/vendo/registry.tsx                 | export: MapleSparkline     | modules: 6 | 20,093 B
MapleSpendingDonut -> src/vendo/registry.tsx                 | export: MapleSpendingDonut | modules: 6 | 20,093 B
```

**Corpus size.** Post-review, demo-bank stores 1 capturable component in 28 KB
over 2 shared blobs (the other two records are skip reasons of a few hundred
bytes); demo-accounting is 20 KB. The dedupe ratio was measured before the
loadability fix shrank the set: 3 components over 8 distinct modules stored
**25,677 B** vs **61,229 B** if each inlined its own copy — **58% saved**, and
it improves with registry size because every shared module and the app-root CSS
are stored once.

**Depth proof.** `MapleSpendingDonut`'s closure reaches `src/lib/cn.ts`,
`src/lib/money.ts`, `src/components/charts/colors.ts` at depth 2 through three
depth-1 modules — a fourth level in the pins fixture is captured too.

**Round trips**, real built CLI against a local store implementing the console's
`/api/v1/store` routes. First sync:

```
GET  /api/v1/store/blobs/vendo_host_components          ← keys-only manifest
PUT  /api/v1/store/blobs/vendo_host_components/<hex>    × 8   (22 KB total)
POST /api/v1/store/records/vendo_host_components/list
POST /api/v1/store/records/vendo_host_components/put    × 3
```

Second sync, nothing changed — **two calls, zero uploads, zero source
downloaded**:

```
GET  /api/v1/store/blobs/vendo_host_components
POST /api/v1/store/records/vendo_host_components/list
```

## Durable data in a project blob namespace

This lane is the first thing to put durable, non-user-scoped data in a project
BLOB namespace. The console's `erase.bySubject` cascade should be checked
against that before this reaches real projects: if a subject erase sweeps
project blobs, the corpus would be stripped silently — the records survive and
their hashes still match, so no later sync would notice or repair it. Named
here, not fixed here (pre-existing console behavior).

## Tests

`packages/actions/src/sync/components.test.ts` (7)
- captures a registered, non-remixable component, including an unexported local
- deep-but-small chain captured in full — depth is no longer the limit
- stops at the package boundary silently; warns only on a broken host import
- byte-budget cut with the honest warning, plus the `too-large` record
- dedupe: one copy for two importers, one entry blob, one helper blob
- prune safety: a shared module survives its other importer's deletion
- idempotent — an unchanged project rewrites nothing, `capturedAt` holds
- degraded scan prunes nothing

`packages/actions/src/sync/pins.test.ts` (2 changed/new)
- the four-deep chain is captured (replaces the depth-2 assertion)
- an over-budget slot is skipped without clobbering its existing baseline

`packages/vendo/src/cli/cloud/host-components.test.ts` (6)
- shared modules upload once; only missing records are put
- a second sync compares over the keys-only manifest and uploads nothing
  (asserts the exact two-call trace)
- stale record deleted, then the module nothing references — the still-shared
  one survives
- consent asked once, saved, silent after
- `--no-push-components` never asks, never saves, never calls
- keyless = zero network (tripwire fetch that throws)

## Contract amendments

**None.** `RecordQuery`, `RecordStore`, and `BlobStore` are untouched; the push
uses `blobs.list()` and `records.list()` exactly as 01-core §12 defines them,
and the console already serves both routes. `SyncReport` gains a `components`
field (additive) and `@vendoai/actions` exports `capturedHostComponentSchema`,
`capturedModuleSchema`, and `hostComponentEntrySource`.

## Not in this lane

The vendo-web PR: swap `hostStub` for a jailed render of the corpus above, and a
"too large to preview" tile for `skipped`. Nothing else moves.

## Note for a separate PR: the portability gate reads prose as code

`packages/core/src/packaging.e2e.test.ts` ("ships a platform-clean dist") greps
the packed dist for `/\brequire\s*\(/`. It matched **doc comments** in
`island-ambient.ts` and `jail-modules.ts` that named the mechanism literally —
there was no CJS require in the emitted code, and the dist was portable the
whole time.

Fixed here by rewording the comments. The gate is deliberately **untouched**: it
guards a real edge/Bun portability guarantee, and loosening a safety test to let
my own code through would be the wrong trade.

Worth fixing properly in its own PR: the check should ignore comments and string
literals before matching, so it flags CJS requires rather than any file that
talks about them. Two things make that more than cosmetic — CI reported only the
first offending file (`island-ambient.js`), so the third occurrence in
`jail-modules.ts` would have failed the *next* run and cost another round trip;
and a gate that fires on prose trains people to reword comments under merge
pressure, which is exactly when a real finding is most likely to be waved
through. Not changed here because it is a change to a safety test and belongs on
its own, not smuggled into a feature PR.

---
## Summary by cubic
Capture the host app’s registered components and their import closures so the console renders them live instead of grey placeholders. Adds content‑addressed storage, a 256 KB per‑component budget, deterministic preview seeds, a slim jail runtime with bundled `clsx`/`tailwind-merge` and a `zod`‑shaped shim, optional Cloud push, and per‑capture runtime requirements.

- **New Features**
  - Capture every `<VendoRoot components={…}>` entry to `.vendo/components/<Name>.json` (refs only) and module/style sources to `.vendo/components/modules/<sha>.json` with content‑addressed dedupe and prune by references.
  - Walk to full closure; stop at package boundaries; enforce a 256 KB per‑component budget (styles excluded); clear skip records when over budget.
  - Preview seeds: prefer declared `examples`, else generate deterministic values from the `props:` schema; record `sampleOrigin`.
  - Jail bundles `clsx` and `tailwind-merge` and ships a `zod`‑shaped shim; each capture records `requires` so consumers can detect stale runtimes.
  - Cloud push for `vendo_host_components`: `records.list()` is the id→hash manifest; keys‑only `blobs.list()` decides module uploads. One‑time consent in `.vendo/cloud.json`; flags `--push-components` / `--no-push-components`; keyless/BYO never uploads. Exports from `@vendoai/actions`: `capturedHostComponentSchema`, `capturedModuleSchema`, `hostComponentEntrySource`.

- **Bug Fixes**
  - Skip captures the jail cannot load; record `unsupported-imports` with specifiers (e.g., `recharts`, `clsx`, `tailwind-merge`, `zod`).
  - Safer prune: derive blob keep‑set from surviving Cloud records; abort on unreadable kept rows; run prune against discovered components; block in‑project symlinks into `node_modules`; paginate index listing.
  - Mark `.vendo/components/**` as generated; docs clarify the package boundary; reworded jail‑module comments in `@vendoai/core` to satisfy the portability gate (no runtime change).

<sup>Written for commit 5d34523a43e749f65e3d85a181c72a8b145d7830. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/750?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
## Review round 1 — all three defects fixed, pushed as `55848f0`

Gates green (`build` 23/23, `typecheck` 41/41, `test` 53/53, `lint`). Two of
the three were reproduced first, then fixed, then re-proven.

### The proof harness

You were right that my original verification measured the wrong thing —
"every `sourceImport` resolves" is vacuous, because `sourceImports` only ever
contains what the walk captured, so a dropped specifier is definitionally not
in the checked set. I rebuilt it as a replica of the real loader (sucrase
`["typescript","jsx","imports"]` per module → evaluate behind a require bound
to the captured table, falling back to `jailRequire`), matching
`runtime-entry.tsx` `compile`/`load`/`localRequire`.

**Before** — reproduced your finding exactly:

```
examples/demo-bank
  ✓ MapleNetWorthCard      LOADS (NetWorthView)
  ✗ MapleSparkline         THROWS — module "zod" is not available in the Vendo jail
  ✗ MapleSpendingDonut     THROWS — module "zod" is not available in the Vendo jail
examples/demo-accounting
  ✗ CadenceDocProgress     THROWS — module "zod" is not available in the Vendo jail
  ✓ CadenceMissingDocsHero LOADS (MissingDocsHero)
  ✗ CadenceStatusBadge     THROWS — module "zod" is not available in the Vendo jail

loads: 2   honest skips: 0   THROWS IN THE JAIL: 4
```

**After:**

```
  ✓ MapleNetWorthCard      LOADS (NetWorthView)
  ~ MapleSparkline         SKIPPED (unsupported-imports) — It imports "clsx", "recharts", "tailwind-merge", "zod", which the sandboxed preview cannot load.
  ~ MapleSpendingDonut     SKIPPED (unsupported-imports) — …
  ~ CadenceDocProgress     SKIPPED (unsupported-imports) — It imports "clsx", "tailwind-merge", "zod", …
  ✓ CadenceMissingDocsHero LOADS (MissingDocsHero)
  ~ CadenceStatusBadge     SKIPPED (unsupported-imports) — …

loads: 2   honest skips: 4   THROWS IN THE JAIL: 0
```

### B1 — loadability is now a capture precondition ✅

`captureClosure` returns `unsupported`: every specifier it did not capture and
`isStrippedIslandSpecifier` does not resolve. I widened it past your ask —
component-local **stylesheet** imports and unresolvable host paths throw in the
jail for exactly the same reason, so the rule is one sentence: *a specifier the
jail will request and cannot answer*. `BLESSED_JAIL_MODULES` (a hand-copy of
the react list) is deleted in favour of the shared core allowlist, so this
cannot drift from `JAIL_MODULES`.

`importSpecifiers` also now drops `import { type A, type B } from "x"` — an
all-inline-type clause erases like `import type` and must not count as a
missing import. I did **not** replicate sucrase's usage-based elision (an
import bound but only used in type position); that biases toward a false skip,
never a false capture, and I would rather not carry a shadow copy of sucrase's
scope analysis.

Not done, as instructed: nothing is blessed in the jail.

**`components.test.ts` fixed** — the test that asserted `recharts → captured`
now asserts the honest skip with `specifiers: ["recharts"]`, plus a companion
proving a react-only-plus-type-imports component still captures.

### B2 — the blob keep-set comes from Cloud, not from what parsed locally ✅

Reproduced both of yours. The keep-set is now derived from the records that
**survive in Cloud** after the reconcile (local record preferred, since that is
what the `put` just wrote), and a kept row this CLI cannot parse aborts the
prune rather than guessing. Three regressions added: corrupt `Donut.json` keeps
its row *and* both blobs; a deleted `modules/` directory deletes nothing; an
unreadable kept row on both sides prunes nothing.

Finding your fixtures were too weak was half the fix: my fake cloud seeded rows
without refs, so an empty keep-set looked correct. The fake now stores real
rows, which is what made the bug fail loudly.

### I2 — prune runs against discovery ✅

`seen.add()` moved to the top of the loop, and the transient "could not be
read" path now writes **nothing** (an existing good capture survives an unlucky
read) instead of recording a skip. Test asserts the record is byte-identical
and unpruned.

Fixing this surfaced a bug of my own: `module` was being derived from the
pre-realpath path, so on a symlinked project dir (`/var` → `/private/var`) it
came out as a `../../..` escape. realpath now happens first. Caught by the
existing test.

### The cheap ones ✅

- **I3** — six reasons now, each with a `detail` sentence a surface can render
  verbatim: `unsupported-imports`, `too-large`, `no-default-export`,
  `default-export-conflict`, `in-package`, `no-named-declaration`. The
  no-named-declaration case moved out of `catalog-scan` so there is exactly one
  explanation per component, on disk, where the console can read it.
- **Symlink nit** — `resolvedCandidate` now tests `realCandidate` for
  `node_modules`, closing the in-project-symlink hole.
- **Page size** — index list asks for 1000.
- **`.gitattributes`** — `examples/*/.vendo/components/** linguist-generated=true`.
- **Docs** — the package boundary is now stated first as a **functional**
  limit under a `<Warning>` ("today, a component that imports any package other
  than React cannot be previewed", naming zod/recharts/clsx/tailwind-merge and
  the shadcn `lib/cn.ts` default), with the skip-reason table beside it.
  `dot-vendo.mdx` gains the size-expectation line and says why gitignoring the
  corpus is unsafe.

### Numbers moved

demo-bank 56 KB → **28 KB**, 8 modules → 2; demo-accounting 36 KB → 20 KB.
Round trips unchanged: fresh push = 2 blob PUTs + 3 record PUTs; **second sync
= 2 calls, 0 uploads, 0 deletes**.

That drop is the honest headline: **only 2 of 6 real registered components in
our own demos can be previewed today.** The lane makes the other four say why
instead of crashing, but the feature only becomes broadly useful once the
package question is answered.

---

## Report-only

**Blessing packages in the jail — the founder call this now blocks.** Our own
two demo hosts hit it on 4 of 6 components, via three separate routes: `zod` in
the registry module (Vendo's own documented `ComponentRegistry` pattern),
`recharts` in a chart, and `clsx` + `tailwind-merge` in `lib/cn.ts` (the shadcn
default, present in most Next hosts). Any realistic registry will look like
this. Worth noting `zod` is the cheapest and highest-leverage of the three: it
is reached only because the registry module carries `props:` schemas next to
the component, so a narrower fix — capture the component's declaring module
with the registration object stripped, or steer hosts to declare components
outside the registry file — might recover a large share without blessing
anything.

**`erase.bySubject` and the project blob namespace.** This lane is the first to
put durable, non-user-scoped data in a project blob namespace, so the
interaction with the console's erase cascade is now live surface. I did not
touch it (pre-existing console behavior, and out of this lane) — but it should
be checked before this ships to real projects: a subject erase that sweeps
project blobs would silently strip the corpus, and the next sync would not
notice, because the records still exist and their hashes still match. If that
is real, the durable fix is for the push to reconcile blob keys against the
record keep-set on every run rather than trusting existence. I have added a
sentence naming this to the PR body.

**The console-side change is small, but it is not mechanical.** Correcting my
earlier claim — the obstacles are real:

1. `AppDocPreview` is a client component building `hostComponents` in a
   synchronous `useMemo`; inflating a record needs N async blob reads, so the
   corpus has to be loaded by a server component above it and passed in.
2. Blob reads need the project's blob namespace; `cloud.ts` wires only
   `records` today, and bodies come back as bytes needing decode + parse.
3. The same `globals.css` and shared modules are referenced by every component,
   so the loader needs a per-request blob cache or it refetches the corpus once
   per tile.
4. Each host node becomes its own `JailedComponent`, which needs
   `onAction`/`onStateSet` — the preview currently stubs those once at the
   `AppFrame` level — and means N nested jail iframes per gallery card.
5. `skipped` needs its own tile, with six reasons to render.
6. The console consumes vendored `vendoai-*-next` packages, so
   `capturedHostComponentSchema` / `capturedModuleSchema` /
   `hostComponentEntrySource` need a re-vendor at a new flowlet commit.

Not merging — over to you for the live browser proof.

PR #755, opened 2026-08-03
self-serve R+P: try unlisted, runnable JS scaffolds, doctor version-skew, CLI honesty fixes

yousefh409:
Seven fixes from the 2026-08-02 self-serve audit (Appendices B/C), Waves R+P. One changeset, `@vendoai/vendo` **minor** (drives 0.7.0).

## Items

**1 — `vendo try` is unlisted.** Removed from the HELP command list and scrubbed from the `--engine` and `--no-ai` flag descriptions; the try-only `--port` and `--no-open` lines are dropped from HELP (the flags still work). The command, its dispatch branch, and every option keep working when invoked. The two retired-command stubs that pointed users AT try — `vendo playground` and `vendo refine` — no longer advertise it (playground now points at `vendo init` + the quickstart). Docs-site untouched (another lane owns it).

**2 — JS-emitted scaffolds are valid JavaScript (audit B2, blocker).** `anonymousPrincipalLines()` emitted `kind: "user" as const` and `expressServerSource()` emitted `(source.headers as Headers & { getSetCookie?: () => string[] })` unconditionally, so every plain-JS host (Express, bare Node, `--framework custom`) died with `SyntaxError: Unexpected identifier 'as'` on its first `node server.js`. Both now branch on the `typescript` flag, mirroring the `signatures` map beside them.

**3 — `vendo doctor` names a stale install (audit F1).** New `npmLatestVersion()` reads npm's dist-tag document with a 2s timeout and fails soft to `null` on any problem. When the installed CLI is behind, doctor prints:

```
warning: installed @vendoai/vendo 0.6.1 is behind latest 0.7.0 — npm install @vendoai/vendo@latest (release-cooldown npm configs like min-release-age resolve old versions silently)
```

A hint, not a registry check: no `error_code`/`fix_ref`, no exit-code change, nothing in `--json`. That keeps `agents/verify.mdx`'s code table (which IS test-enforced) correct without touching docs-site.

**4 — `vendo mcp server-json` no longer hangs a script (audit B6, blocker).** Missing `--domain`/`--url` on a non-TTY stdin used to fall into `readline.question` forever. Now exit 1:

```
vendo mcp server-json: --domain and --url are required non-interactively (example: vendo mcp server-json --domain example.com --url https://example.com/api/vendo/mcp)
```

TTY behavior is unchanged.

**5 — `vendo sync --report` without a key exits 1 (audit B6, blocker).** It used to print a complaint and exit 0, so a CI reporting lane stayed green while never reporting. New message: ``vendo sync: --report needs a Vendo Cloud key — run `vendo login`, set VENDO_API_KEY, or pass --key.`` The `--strict` codes (2 breaking, 3 impacted) are more specific and keep their meaning; 1 applies when the run would otherwise be green.

**6 — Hosted-store notice is once per process (audit F7).** A Next dev server recomposes on nearly every request, so "once per composition" printed the paragraph 29 times in one short session. Module-level latch, same posture as `hostedSessionOps`' door warn.

**7 — Init's ending and the Cloud pitch (audit F5, F8).**
- The run's **last** line is now the outstanding paste, on interactive and non-interactive runs alike. The frame itself stays up-screen and the closer is a one-line echo: `→ Don't forget the paste in app/layout.tsx (frame above)`. This is the brief's stated escape hatch — the agent tail's own pointers say "the `<VendoRoot>` lines **above**", so moving the frame below the tail would make those lines false.
- "Then start your dev server — the agent is live in your app" is **withheld while a paste is pending** — it contradicted the frame two lines above it, which had just said Vendo stays invisible until the paste lands.
- The keyless Cloud pitch is 3 lines (what a key unlocks / `vendo login` / `--byo`) instead of 5; the device-flow walkthrough lives in `vendo login`.

**Bonus (comment-only, no behavior change):** two stale `server.ts` JSDoc comments corrected — the pipeline-flags comment now states real defaults (`structuredRepair` and `smokeRender` ON by default; `exemplarContract`/`regionParallel`/`endPass`/`rebind` opt-in), and `profileDir`'s cross-reference no longer names a nonexistent top-level "actions block".

### Item 2 — real Express host, built CLI, `node --check`

`npm init -y` + `npm install express@5` in a temp dir, then this branch's built CLI:

```
$ node <worktree>/packages/vendo/dist/cli.js init . --yes --byo --no-ai --framework express
Wired (3 files):
  + vendo/registry.mjs
  + vendo/server.mjs
  ~ package.json

$ node --check vendo/server.mjs
PARSE OK: vendo/server.mjs
$ node --check vendo/registry.mjs
PARSE OK: vendo/registry.mjs

$ grep -n 'kind:\|getSetCookie =' vendo/server.mjs
21:  principal: async () => ({ kind: "user", subject: "demo-user" }),
48:  const getSetCookie = source.headers.getSetCookie;
```

Same file with the two pre-fix expressions restored (lines 21 and 48 — exactly the lines the audit reported):

```
$ node --check before.mjs
before.mjs:21
  principal: async () => ({ kind: "user" as const, subject: "demo-user" }),
                                         ^^
SyntaxError: Unexpected identifier 'as'
```

Unit test `init-scaffolds.test.ts` runs `node --check` over both JS scaffolds, so Node's own parser is the gate — a substring assertion would miss the next annotation that sneaks in.

### Item 7 — init transcript tail, Next host, built CLI

**Before** (order on main): frame → `Then start your dev server — the agent is live once you add a model key.` → `Verify everything: …` → `Agent tail:` … ending on `gate: run vendo doctor --json`. Interactive runs ended on the star ask.

**After** (real run, this branch):

```

  File: app/layout.tsx

Verify everything: `npx vendo doctor` (it can start the server and run a live turn).

Agent tail:
  auth: none wired — sessions stay anonymous until a preset is added
  edit vendo/registry.tsx — register the components the agent may render (generated empty)
  edit app/api/vendo/[...vendo]/route.ts — add the auth preset named in the advisory above when the host has auth
  edit app/layout.tsx — wrap the app in the <VendoRoot> lines above (it mounts <VendoOverlay />, the visible surface; without it users see nothing)
  edit .vendo/brief.md — replace the placeholder with what this product does and for whom
  cloud key: none — for Vendo Cloud, fetch https://vendo.run/auth.md and run `vendo login` …
  gate: run `vendo doctor --json` — done when every check reports green

→ Don't forget the paste in app/layout.tsx (frame above)
```

Note the "start your dev server" line is gone (paste pending) and the paste echo is the last word.

Keyless Cloud pitch, same CLI, 5 lines → 3:

```
Vendo Cloud (optional): not configured. A key unlocks …
Vendo Cloud key (agent path): run `vendo login` — it prints a code your human approves in the browser, and the minted VENDO_API_KEY lands in .env.local (never printed).
Then re-run `vendo init` (it picks the key up) or pass --cloud-key <key>. Protocol: https://vendo.run/auth.md
No Cloud account wanted? Re-run with --byo and set a provider key (ANTHROPIC_API_KEY / OPENAI_API_KEY / GOOGLE_GENERATIVE_AI_API_KEY).
```

### Item 6 — the once-only test is real

With the latch removed the new test fails `expected [ …(3) ] to have a length of 1 but got 3`; with it, 1. Three compositions in one process, one notice.

## Tests

New: `init-scaffolds.test.ts` (node --check both JS scaffolds + the TS variants keep their annotations), `doctor-version-skew.test.ts` (stale / current / offline / never-backwards, plus the lookup's fail-soft paths), `hosted-store-notice.test.ts` (once per process). Extended: `server-json.test.ts` (non-TTY exit, half-flags, both-flags-no-TTY), `sync.test.ts` (keyless `--report` exits 1).

Updated because the change IS the change: the help/stub assertions in `cli.test.ts` / `cli.try.test.ts` / `cli.playground.test.ts`; the two `sync.test.ts` cases that encoded exit 0 and the old wording; the init-ending order in `init.test.ts` and the closing-line cases in `onboarding-safety.test.ts` (those now run against a host with the surface already mounted, so the model-key truth-telling invariant is tested exactly as before, and a new case asserts the line stays silent while a paste is pending). `doctor.test.ts` gains an `npmLatest: async () => null` stub so the suite makes no registry round-trip.

## Gate

`pnpm build`, `pnpm typecheck`, `pnpm lint` green. `pnpm test` green.

Pre-existing, unrelated, reproduced on a clean tree at this branch point: `packages/vendo/src/dev-creds/model.test.ts` fails 2/29 locally (`Cannot find module '@ai-sdk/anthropic'` under `withoutGlobalPaths` — a local resolution issue, not a code change). `examples/demo-bank/src/vendo/away-drill.test.ts` flaked once ("Maple exited early") and passed on re-run.

---
## Summary by cubic
Improves CLI reliability and UX: JS scaffolds run in plain JS hosts, non-interactive flows fail fast with clear errors (blank flags included), and `vendo doctor` warns when the installed `@vendoai/vendo` lags npm (skips under `--json`). `vendo try` is hidden from `--help`, logs are quieter, and init ends with a clear paste reminder.

- **Bug Fixes**
  - JS scaffolds no longer emit TypeScript in JS output; Node parses cleanly (branches on the `typescript` flag).
  - Non-TTY flows fail fast: `mcp server-json` requires `--domain` and `--url` (blank values count as missing; TTY prompts again, non-TTY exits 1); `sync --report` without a key exits 1.
  - Hosted-store notice prints once per process to avoid dev-server spam.

- **New Features**
  - Doctor version hint: warns when installed `@vendoai/vendo` is behind npm `latest` (2s timeout, fail-soft, no exit-code change, skipped under `--json`).
  - Init UX: the final line is the paste reminder; “agent is live” is withheld until the paste lands; Cloud pitch trimmed to three lines via `vendo login`.
  - CLI help: `vendo try` is unlisted and no longer promoted by retired stubs; the command still works when invoked.

<sup>Written for commit 67b348f76a58b1777afd32099b057ccbffaa3ed6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/755?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/self-serve-cli-honesty.md
+++ b/.changeset/self-serve-cli-honesty.md
@@ -1,0 +1,42 @@
+---
+"@vendoai/vendo": minor
+---
+
+Seven self-serve fixes across the CLI: the install path stops lying, and the JS
+scaffolds run.
+
+**Plain-JavaScript hosts boot again.** The generated `vendo/server.mjs` carried
+two pieces of TypeScript — `kind: "user" as const` in the principal line and a
+`as Headers & { … }` cast around `getSetCookie` — so every Express, bare-Node
+and `--framework custom` host on a JS codebase died with `SyntaxError:
+Unexpected identifier 'as'` on its first `node server.js`. Both expressions now
+follow the host's language, and Node's own parser gates them in CI.
+
+**`vendo doctor` names a stale install.** npm release-cooldown configs
+(`min-release-age`) silently resolve an old `@vendoai/vendo`, and nothing ever
+said so. Doctor now checks npm's `latest` and prints `warning: installed
+@vendoai/vendo X is behind latest Y` with the upgrade command. Fail-soft: an
+offline, blocked or slow registry says nothing at all and never changes the
+exit code.
+
+**Two silent CI failures are loud.** `vendo mcp server-json` with missing flags
+used to fall into a readline prompt even on a piped stdin — a script or agent
+hung forever; it now exits 1 naming `--domain` and `--url`. `vendo sync
+--report` without a Cloud key used to complain and exit 0, so a reporting lane
+stayed green while never reporting; it now exits 1.
+
+**`vendo try` is unlisted.** The command still runs for anyone invoking it, but
+help no longer advertises it (nor do the retired `playground`/`refine`
+notices): the pre-install `npx vendo try` pitch it fronted resolves no npm
+package.
+
+**Init's ending puts the paste last.** The run's final line is the outstanding
+paste, on interactive and non-interactive runs alike, instead of the star ask
+or the agent tail; the "start your dev server — the agent is live in your app"
+line is withheld while a paste is still pending (it contradicted the frame
+right above it); and the keyless Cloud pitch is three lines, since `vendo
+login` narrates its own ceremony.
+
+**Quieter dev-server logs.** The hosted-store automations notice is latched per
+process — a Next dev server recomposes on nearly every request, and the
+paragraph was landing in the host's log dozens of times per session.
--- a/packages/vendo/src/cli.playground.test.ts
+++ b/packages/vendo/src/cli.playground.test.ts
@@ -2,27 +2,30 @@ import { describe, expect, it, vi } from "vitest";
 import { main } from "./cli.js";
 
 describe("vendo playground retirement", () => {
-  it("fails with a one-liner pointing at vendo try", async () => {
+  it("fails with a one-liner pointing at the install path", async () => {
     const error = vi.spyOn(console, "error").mockImplementation(() => {});
     expect(await main(["playground"])).toBe(1);
     const printed = error.mock.calls.flat().join("\n");
-    expect(printed).toContain("vendo try");
+    expect(printed).toContain("vendo init");
     expect(printed).toContain("retired");
+    // `try` is unlisted (self-serve audit B1) — a retirement notice must not
+    // send the next stranger at a command help does not name.
+    expect(printed).not.toContain("vendo try");
     error.mockRestore();
   });
 
-  it("points at vendo try even when the old flags ride along", async () => {
+  it("says the same thing when the old flags ride along", async () => {
     const error = vi.spyOn(console, "error").mockImplementation(() => {});
     expect(await main(["playground", "--port", "4123", "--no-open"])).toBe(1);
-    expect(error.mock.calls.flat().join("\n")).toContain("vendo try");
+    expect(error.mock.calls.flat().join("\n")).toContain("vendo init");
     error.mockRestore();
   });
 
-  it("--help lists try, not playground", async () => {
+  it("--help lists neither playground nor the unlisted try", async () => {
     const log = vi.spyOn(console, "log").mockImplementation(() => {});
     expect(await main(["--help"])).toBe(0);
     const help = log.mock.calls.flat().join("\n");
-    expect(help).toContain("try");
+    expect(help).not.toMatch(/\btry\b/);
     expect(help).not.toContain("playground");
     log.mockRestore();
   });
--- a/packages/vendo/src/cli.test.ts
+++ b/packages/vendo/src/cli.test.ts
@@ -43,13 +43,12 @@ describe("vendo CLI commands", () => {
     log.mockRestore();
   });
 
-  it("vendo refine is no longer a command: retirement notice pointing at sync + try", async () => {
+  it("vendo refine is no longer a command: retirement notice pointing at sync", async () => {
     const error = vi.spyOn(console, "error").mockImplementation(() => {});
     expect(await main(["refine"])).toBe(1);
     const output = error.mock.calls.flat().join("\n");
     expect(output).toContain("vendo refine was retired");
     expect(output).toContain("vendo sync");
-    expect(output).toContain("vendo try");
     error.mockRestore();
   });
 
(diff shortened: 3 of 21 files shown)


yousefh409:
Greptile P2 on `server-json.ts:70` (empty flags bypass the required check) — **real, confirmed, fixed in 67b348f**.

Reproduced against the built CLI on the previous head:

```
$ vendo mcp server-json . --domain= --url= < /dev/null
Could not generate server.json: Invalid registry domain:
```

`option()` in `cli/mcp/index.ts` returns `""` for both spellings — `--domain=` via `.slice(name.length + 1)`, and `--domain ""` via `args[exact + 1]`. My guard tested `=== undefined`, so a blank value read as supplied: the non-TTY guard was skipped and the run reached the generator with garbage. Second hole from the same root cause: `options.domain ?? await prompt(...)` only falls through on null/undefined, so on a **TTY** a blank flag silently used `""` instead of asking.

Fix treats blank as missing on both paths:

```ts
const supplied = (value: string | undefined): string | undefined =>
  value === undefined || value.trim() === "" ? undefined : value;
```

After, same invocations:

```
$ vendo mcp server-json . --domain= --url= < /dev/null
vendo mcp server-json: --domain and --url are required non-interactively (example: …)
$ vendo mcp server-json . --domain "" --url "" < /dev/null
vendo mcp server-json: --domain and --url are required non-interactively (example: …)
$ vendo mcp server-json . --domain example.com --url https://example.com/api/vendo/mcp
Wrote server.json for com.example/maple
```

Two tests added: blank flags non-interactively hit the guard (and never `Invalid registry domain`), and a blank flag on a TTY asks the question instead of accepting the blank. `server-json.test.ts` 10/10.

PR #768, opened 2026-08-03
feat(automations): Automation Rehearsal — replay a schedule before enabling

Ayush2k02:
## What
Replay a disabled automation's recent schedule firings before enabling it — real reads, simulated writes — as a per-firing timeline with a 7d/30d window toggle (default 30). Each firing reads as-of its own date, and Rehearse is offered only where it would actually show something.

## Why
Enabling a standing automation is otherwise a leap of faith: no way to see what it would do before it runs unattended. Rehearsal previews the real outcome — a derived savings transfer, a client-chase, a document-review pass — and never advertises a preview it can't deliver.

## How
`rehearse()` replays firings through a dedicated guard venue: reads run for real, writes resolve to a simulated card carrying the same block/ask verdict a live run would hit. Windowed reads are pinned per firing (projected as-of that date where a host isn't natively time-series), so amounts derived from them differ per firing. A per-row outlook reuses the exact predicates `rehearse()` applies, so the panel disables or drops the control where a rehearsal would be unsupported or empty. A `forEach` matching nothing is reported as skipped, not dropped.

## Docs and Vid
https://derive.to/artifacts/vendo-automation-rehearsal-v1-pr-1-wozbol6c

https://github.com/user-attachments/assets/d58aabe1-4d69-491a-801a-e804787cd524

## Suggested reading order
1. `packages/automations/src/engine.ts`
2. `packages/guard/src/guard.ts`
3. `packages/automations/src/index.ts`
4. `packages/vendo/src/wire/automations.ts`
5. `packages/ui/src/chrome/automations-panel.tsx`
6. `apps/demo-bank/src/demo-script/automations.ts`
7. `apps/demo-accounting/src/demo-script/automations.ts`
8. `apps/demo-accounting/src/server/asof.ts`


Ayush2k02:
[yousefh409](https://github.com/yousefh409)

let me know if this feature makes sense, will resolve the conflicts and clean up code a bit. needed to seed some data n some small fixes which I included in this PR

Ayush2k02:
@cubic-dev-ai review

PR #770, opened 2026-08-03
claudeCode() harness redesign — capability-first (D1–D8)

yousefh409:
Makes `claudeCode()` capability-first: the box is the security boundary, the door is the permission system, and everything else gets out of the model's way. Spec: `docs/superpowers/specs/2026-08-03-claude-code-harness-redesign-design.md` (approved in conversation before any code).

## The eight decisions

| | | |
|---|---|---|
| **D1** | `permissionMode: "bypassPermissions"`; `canUseTool` + `allowedTools` deleted | 13-name deny-list is the only local tool law |
| **D2** | Claude-code door lists everything the ctx projects | new `Harness.toolSurface`; `vendo()`'s loadout untouched |
| **D3** | `search_connectors` + `list_connections` | present only when connectors are configured |
| **D4** | Files-first app gen; `vendo_apps_create`/`_edit` withheld | the model writes `plan.vendo`/`app.vendo` with its own hands |
| **D5** | `outputSchema` carried descriptor → listing → MCP wire | the model learns query fields without calling |
| **D6** | `/host/components/*.md` + `references/format.md` (623 lines) | feeds Claude Code's real strength: reading files |
| **D7** | `building-apps` rewritten | validate law, `Task` staffing, verbatim ask, data-honesty |
| **D8** | Connectors prompt section, app-default line, 6-line brief | one assembler, harness-conditional sections |

The safety projection (design §12 — unattended runs never see destructive tools) is untouched and runs *before* the listing. `withhold` binds at both listing and call, and the door serves both off live `turn.tools`, so "not projected" means not reachable.

## What the verification loop found

Six adversarial rounds. **Rounds 2, 4, 6 and 8 each caught bugs created by the previous round's own fixes**, and five tests were caught asserting a defect *as* an invariant — which is why green suites repeatedly meant nothing here.

Two findings never shipped and are the reason this took the night:

- **Cross-tenant `fn:` exfiltration** (`8f22f417e`) — the new files-first persist path read the existing app row *unscoped*, so one user's harness could write `/user/apps/<another user's appId>/app.vendo` carrying `<Query tool="fn:dump"/>`. The write was refused; the query still ran on the other user's sandbox and returned the answer. Found by a security assessment of the new path, proven with a failing test, fixed in one logical line.
- **Two silent data-loss paths** (rounds 6→9) — a refused save left an orphaned history version so the next `undo()` destroyed the concurrent edit it had just preserved; and a files-first save let `pins.rebase` overwrite a user's remix with the pristine baseline while returning `status: "rebased"`. The second survived its own first fix and took three attempts.

Six more defects were **latent or pre-existing**, exposed rather than caused by this work: the render seam's data fill had never had a caller (so every harness-built app rendered empty), the connect card had no consumer on the harness path, `outputSchema` could collapse an entire `tools/list` via a dangling `$ref`, connector expansion leaked process-wide (35 → 301 tools for every later session), `strictMcpConfig` was unset, and four `ToolRegistry` wrappers silently dropped an optional method.

## Proven live, in a real browser

Boxed on a template hash-verified byte-identical to this worktree's `dist/claude-turn.js` (13/13 deny-list names baked, `canUseTool` absent), against `demo-bank` + a real e2b sandbox + a real Composio account:

- **Files-first app renders real data** — skeleton at +171s, real values at +222s (`$4,405.05`, `$2,850.00`, 18 transactions), card settles to `ready`. Store row proven by an `app-lifecycle` audit row; `vendo_apps_open` works; the app lists; a fresh turn re-renders it from the store alone.
- **Validate loop** recovered from a genuine error twice, organically.
- **Formerly-denied SDK tools** (`Skill`, `Agent`) ran with zero denials.
- **Tool surface** 14/14: `create`/`edit`/`find_tools` absent, `validate` + both connector tools + lifecycle present, `outputSchema` on the wire.
- **Audit rows**: 42 host-tool calls, **0 with `venue=mcp`** — the parity gate holds.
- **Connect card** rendered live; **expansion scoping** fixed (fresh conversation back to 35 tools).

## Known limits, deliberate

- **In-session connector expansion does not work on Claude Code.** Our door delivers `notifications/tools/list_changed` correctly and the expanded tool *is* callable — but agent-sdk 0.3.215 registers no list-changed handler for an HTTP MCP server, so the model never re-lists. Measured: exactly one `tools/list` per session. The connectors prompt still implies immediate availability; corrected with the dispatcher follow-up (Yousef's call).
- **`outputSchema` is silently absent on Cloudflare Workers** — each schema is compiled with the MCP client's own Ajv validator and omitted on throw; Workers forbids code generation, so every compile throws. Warned once. Safe direction, feature-less on that target.
- **Connector search is keyword-based.** Published measurements put keyword/BM25 retrieval over thousands of tools at 34–64% on *straightforward* requests vs ~94% for embeddings. Follow-up.
- `machine: "local"` unchanged this wave. `HISTORY_LIMIT` left at 50 deliberately. The deny-list is correct but not provably complete (its audit is keyed on PascalCase schema names; snake_case-registered tools are invisible to it) — accepted.

## Gates

`pnpm build && pnpm test && pnpm typecheck && pnpm lint` green at every commit, uncached, 55/55 test tasks. Note for CI: at default turbo concurrency the real-server fixture e2e suites flake with uniform ~2.5s timeouts and the failing suite *moves* between runs — all pass standalone. A concurrency cap is separate infra work.

Contract amendments recorded in `docs/archive/contracts/01-core.md`, `10-mcp.md`, and `2026-07-30-build-contract.md` §1/§1.1/§1.6.

---
## Summary by cubic
Redesigns the `claudeCode()` harness around capability-first access with an uncurated tool surface, connector discovery, end‑to‑end `outputSchema`, and files‑first apps that persist, list, and render real data. Hardens the MCP door, registry scoping, history, and the render seam for correctness and safety.

- **New Features**
  - Capability-first surface: `bypassPermissions` with a 13‑name deny‑list; new `toolSurface` contract (uncurated for `claude-code`, `vendo()` loadout unchanged); `vendo_apps_create`/`_edit` withheld for files‑first generation.
  - Connector discovery: `search_connectors` and `list_connections` projected as ordinary tools when connectors are configured; lazy expansions are session‑scoped via a new `listingScope`, with `releaseListingScope` plumbed through `ToolRegistry`, guard.bind, `createConnectGate`, `vendo_delegate`, and BYO approvals.
  - Result shapes: `outputSchema` moves onto `ToolDescriptor` and travels descriptor → turn listing → MCP door; sanitized and object‑typed on the wire, omitted on Workers; structured content guaranteed; truncation envelope keys moved to `vendo_*`.
  - Files‑first apps: `AppsRuntime.authored` upserts via the engine writer and resolves queries through the same guard‑bound caller; the render seam streams a plan skeleton (deferred if an app render occurs) and always settles with the final view; `open()` and the UI show “Data didn’t load” when any query fails.
  - Authoring/reference: `/host/components/*.md` projection and skill companion files; `building-apps` rewritten (conditional `Task`, validate‑after‑save, data‑honesty); prompt adds a connectors section per harness rail and a 6‑line embedding brief; the thread renders a connect card from `data-vendo-connect`.

- **Bug Fixes**
  - Security: cross‑tenant `fn:` exfil blocked — `authored()` reads a previous row only when `mayWrite`.
  - Registry scope: fixes process‑wide connector expansion leaks; search/list carry the caller’s `ctx`; listings narrow by session; `releaseListingScope` now forwarded through all decorators and called on session end.
  - MCP door: single‑SSE handling repaired; reliable `tools/list_changed` after successful expansion; strict MCP config; `outputSchema` sanitized and compiled with the client’s Ajv (cached) to avoid list‑killers.
  - History/pins: transactional append with discard on failure; prune only after a write lands; `PinIntent` adds `kind: "fork" | "edit" | "touch"` and `pins.rebase` now requires a recorded fork and replayable edits.
  - Render seam/UX: plan emit no longer overwrites app data; final paints always settle; truncation envelope no longer collides with host schemas (`vendo_*`); UI dedupes connect card against the engine path and shows a truthful partial‑data notice.

<sup>Written for commit 66f57c92771bff8d474eea3b8861fed10711c1e0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/770?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
## Rebased onto main (`abc7a44a1` → `d0c3cc936`), and one interaction worth reading before merge

Rebased across **#747** (`feat(guard)!: risk grading stops guessing from names` — 156 files) and #750. 14 commits, all four gates green with real exit codes at the rebased head: build 24/24, typecheck 43/43, test **55/55 uncached**, lint 6/6.

### The semantic audit came back clean, for a structural reason
This wave adds **zero risk-derived branching** in production code — every `risk` reference it introduces is a pass-through (descriptor → listing → wire → row) or a constant on a Vendo-authored descriptor. So nothing here can pass for the wrong reason on that axis, and `ungraded` needs no answer we failed to give it. Verified individually: nothing of ours reads `critical`; turn-tools already maps `pending-approval`, so an ungraded call parks correctly; `search_connectors` passes `match.risk` straight through, so an ungraded Composio tool honestly reports `ungraded`; and `dataUnavailable` counts `pending-approval`, so an ungraded query that parks renders the honest notice rather than a silent `—`.

The live proofs survive: demo-bank's `tools.json` is now 25/25 `ungraded`, but `judgments.json` grades 24 of them and the registry merges judgments over the extracted label, so the $4,405.05 proof's queries are still `read` and run silently.

### ⚠️ One consequence of #747 and D2 *meeting*, which is nobody's bug
#747 withholds `ungraded` from unattended runs alongside `destructive`, and re-grades every extracted GET to `ungraded`. D2 deleted the loadout, so "everything the context projects" is now the whole story. Together: **an unattended claude-code run against a catalog nobody has judged sees zero host tools**, where before it saw a loadout of nominally-`read` ones.

That is #747's deliberate fail-closed posture, not a defect in either change — but the practical effect is that `vendo judge` (or a hand-written `judgments.json`) becomes a prerequisite for unattended automation on this harness. The D2 spec section said "the safety projection is untouched"; that sentence is now corrected in `dbbb0bbf0` rather than left standing.

### Two archived contracts #747 left behind, corrected here
Both sit in blocks this wave amends, so a reviewer would land directly on the contradiction: the build contract still declared `ToolListing.risk` as the three-value union, and `01-core.md` still documented `critical` / `decidedBy: "critical"`.

### Three inherited issues deliberately NOT fixed (they are #747's semantics)
1. **`door.ts` emits `destructiveHint: false` for an `ungraded` tool.** MCP's spec default for that field is `true`, so the door actively tells clients "this is not destructive" about a tool nobody graded — the exact class of claim #747 exists to delete. Our new `turnTools` listing is a second call site inheriting it. **Worth a follow-up.**
2. `pack.ts` invents `risk: "write"` in the delegate registry's fallback descriptor (pre-existing, display-only).
3. #747's own "no code path concludes anything from a tool's name" is overstated — `mechanicalRisk` in `grant-sets.ts` still does verb-shape inference, kept deliberately. (Good for us: a `connector-discovery.ts` comment depends on it existing.)

Also untracked two per-PID scratch fixtures (`packages/vendo/.wire-parity.*`) that a killed test run left behind and an earlier commit here swept in.

PR #771, opened 2026-08-03
feat(jail): preview resolves npm packages from a pinned CDN

yousefh409:
The console renders a host's own registered components in the jail (#750). A
component that imports anything the jail does not bundle is still refused —
Maple's two chart components read `Can't preview — imports recharts`. This closes
that gap: the **preview** jail resolves npm packages from ONE pinned CDN, at the
exact version the host has installed.

Verified in a real browser against the real captured bytes: Maple's
`MapleSpendingDonut` draws an actual recharts donut inside the jail — 2 pie
sectors, its `$628.60` centre label, 8 CDN requests all on one origin, 0 console
errors, 0 CSP violations, no `Referer`, no cookies, `Origin: null`.

---

## 1 — The CSP, before and after

The jail is two nested `about:srcdoc` frames; both carry the same policy.

**Before — every jail:**

```
default-src 'none'; script-src 'nonce-<N>' 'unsafe-eval'; style-src 'unsafe-inline';
img-src data:; font-src data:; connect-src 'none'
```

**After — a jail with NO package pins.** Every production render, and every
preview of a component that needs no packages. Character-for-character the line
above; the only change to that document is one new `<meta name="referrer">` (§5).

**After — a jail WITH package pins (console preview only):**

```
default-src 'none'; script-src 'nonce-<N>' 'unsafe-eval' https://esm.sh data:;
style-src 'unsafe-inline'; img-src data:; font-src data:; connect-src 'none'
```

One origin, `script-src` only. `default-src 'none'` and `connect-src 'none'`
unchanged; no `allow-same-origin`; `blob:` still banned; the sandbox is still
exactly `allow-scripts`.

**`data:` is the one thing beyond the single origin, and it is not optional.**
The CDN serves the package with React left external, so the package imports bare
`react`; an import map redirects that to the jail's OWN React instance, and the
only transport for "a module this opaque-origin realm owns" is a `data:` URL
(`blob:` is the alternative, and is deliberately banned). Without it the CDN
serves a second React and the render dies — browser-measured, not assumed:

```
recharts + the CDN's own React → 0 svg, 0 paths,
                                 "Cannot read properties of null (reading 'useContext')"
recharts + the jail's React    → 1 svg, 3 paths, no errors
```

`data:` scripts are inline code and open no network channel; behind
`'unsafe-eval'` they grant the realm nothing it did not already have.

### A pre-existing finding, flagged not folded in

The comment in `JailedComponent` says `script-src` is a "second wall" stopping a
jailed component from reaching the module loader. **It is not.** A nonce in
`script-src` is inherited by dynamically-imported module scripts, `new Function`
included — so today's jail, before this PR, already loads modules from any
origin:

```
policy: script-src 'nonce-N' 'unsafe-eval'          (today, unchanged)
probe:  new Function("return import('https://cdn.jsdelivr.net/npm/clsx@2.1.1/+esm')")()
result: LOADED
```

The harness probe asserting `import: FAILURE (CSP)` is mislabelled: what blocks
it is sucrase's `imports` transform rewriting `import()` into the jail's own
`require`, not the policy. So naming `https://esm.sh` here grants the jail no
capability it lacked — the declaration is what makes CDN loading survive the real
fix (dropping the nonce in favour of `'unsafe-inline'`, which is its own change
with its own proof).

## 2 — The venue gate, and how it is enforced

`JailFurnishing.packages` (import specifier → `name@version[/subpath]`) is the
whole gate. No pins → no CDN origin in the policy → no requests. The production
path is not "configured off"; it is structurally the jail that has always
shipped.

Three independent walls, because the jail is ALSO the production render path for
a remix fork inside a customer's own page:

1. **The runtime cannot produce one.** `attachPinFurnishings` copies a fixed
   field list (`sourceImports`, `subSources`, `sampleProps`, `styles`). Tested
   with a pin baseline that *does* carry `packages` — it is not forwarded.
2. **A document cannot claim one.** `stripServerAuthoritativeFields` now deletes
   `packages` off every furnishing, so a model-written tree or an imported
   `.vendoapp` cannot make a customer's page fetch third-party scripts. It runs
   at persist AND at open — the same treatment `inClient` and `pinDrift` get.
3. **Capture reports facts; the venue decides.** `captureClosure` returns
   `packages` ALONGSIDE `unsupported` rather than removing them from it. Pin
   baselines read `unsupported` exactly as before, so their behaviour is
   byte-identical; only the host-component path subtracts the CDN-loadable ones
   (`previewBlockingSpecifiers`).

Browser-proven, not asserted: `/tree-jail` renders a real pin furnishing (the
fork shape) and every mounted `script-src` comes out
`script-src 'nonce-…' 'unsafe-eval'`, with zero requests to the CDN. The same
captured Maple component with only `packages` removed also makes zero.

## 3 — The CDN: esm.sh

Ranked on the requirement that actually decides it — **can the fetched package
use the jail's React?**

| | one origin for the whole graph | externalises React | exact-version immutable |
| --- | --- | --- | --- |
| **esm.sh** | yes | yes (`?external=`), and it propagates through transitive deps | yes |
| jsdelivr `+esm` | yes | **no** — `?external`/`?deps` silently ignored, and its graph pulls three different Reacts | yes |
| unpkg `?module` | yes, but a 301 + fetch per file (~500 for recharts) | no | ranges appear in URLs |
| ga.jspm.io | yes | yes | yes, but no entry resolution — needs a build-time generated import map |
| skypack | — | — | unmaintained; resolves recharts against React 17 |

Pinned in exactly one place: `JAIL_PACKAGE_CDN_ORIGIN` in
`packages/core/src/jail-modules.ts`. The template is
`{origin}/{name}@{version}{subpath}?target=es2022&external=react,react-dom&standalone`,
and every parameter is load-bearing:

- `external=react,react-dom` — the React-instance fix above.
- `standalone` — collapses the dependency graph into one module. Measured:
  recharts costs **113 requests without it, 8 with it**, and a gallery mounts
  several jails at once.
- `target` — esm.sh otherwise answers `Vary: User-Agent`, which would make the
  bytes a preview gets depend on the browser string.

**Versions come from the package the host actually has INSTALLED** —
`node_modules/<name>/package.json` — not from the range in their `package.json`,
and not from a lockfile. That file is what Node reads to run their app, so a
preview draws the version their product draws, and it needs no
pnpm/npm/yarn/bun lockfile matrix. A version that cannot be resolved exactly is
never guessed at.

Known limits, stated rather than discovered later: the pin freezes the
*package*, not the dependency graph esm.sh bundles into it; esm.sh is one
maintainer and one origin server (self-hosting is the escape hatch — it is a Go
binary, and "one pinned origin" can become ours); and one browser test now
depends on esm.sh being reachable.

## 4 — What a failure looks like

| what | what the user sees |
| --- | --- |
| package is private, or a workspace/`link:` package | capture-time skip: `Can't preview — imports @acme/design-system`, and the record's detail says *"@acme/design-system is marked private, so it is not published"* |
| package is not installed | capture-time skip: *"recharts is not installed, so its exact version cannot be resolved"* |
| CDN unreachable, 404, version unpublished | render-time notice: **`MapleSpendingDonut: could not load recharts@3.9.2`** — the muted contained note, not the red error box |
| a pin that is not an exact version | the same notice; no request is made at all |

The render-time notice deliberately ignores `streaming`. A preview sets
`streaming` permanently, and the existing rule ("mid-stream, a crash is not a
verdict") would swallow this into a shimmer skeleton that never resolves — the
exact bug that cost a browser-investigation cycle last lane. A package that will
not load is terminal and factual, so it is always said out loud. Proven by
aborting every request to the CDN origin and asserting the notice appears with no
skeleton left on the page.

## 5 — One leak the browser caught

An opaque origin is not a private one. Under the default referrer policy the
jail's subresource requests carried **`Referer: <the embedder's URL>`** — which in
the console is the page naming the project. Fixed with
`<meta name="referrer" content="no-referrer">` in the jail head, and now
asserted: no `Referer`, no `Cookie`, no `Authorization`, `Origin: null`.

## 6 — Proof

**Real browser** — `packages/ui/e2e/cdn-packages.spec.ts`, 4 tests, real
Chromium, real esm.sh. The spec reads Maple's committed capture off disk (record,
module blobs, sample seed, package pins), inflates it exactly the way the
console's `inflateHostComponent` does, and hands it to the shipped renderer. It
supplies the jail **nothing the record did not declare** — the specific failure
mode of the previous lane's harness, which reported "4 of 6 draw" while the
browser drew 1 because it provided packages unconditionally. If the record were
skipped or carried no pins, the spec fails rather than passing vacuously.

1. `MapleSpendingDonut` draws: 2 `.recharts-pie-sector` paths, its `Total`
   label, 8 CDN requests all on `https://esm.sh`, 0 console errors, 0 CSP
   violations, no referrer/cookie/auth headers.
2. Production venue (`/tree-jail`, a real pin furnishing): every `script-src` is
   `'nonce-…' 'unsafe-eval'`; 0 CDN requests.
3. The same captured component with `packages` removed: 0 CDN requests.
4. CDN blocked at the network layer: the honest notice, no skeleton.

**Real CLI** — `vendo sync` in `examples/demo-bank` moves `MapleSpendingDonut`
and `MapleSparkline` from `skipped: unsupported-imports` to captured, with
`requires: ["clsx","recharts","tailwind-merge","zod"]` and
`packages: {"recharts": "recharts@3.9.2"}` — recharts' installed version,
resolved from a `^3.8.1` range. Both records are in this diff.

**Unit** — pin validation, including `recharts@3.9.2/../../etc/passwd` (which the
first draft of the regex accepted) and every floating form; capture recording
exact versions, subpaths, private packages and workspace links; the two venue
walls; and the CSP asserted from the rendered `srcdoc` rather than from source.

## 7 — Backward compatibility, which matters across repos

Package names go into the EXISTING `requires` field; the CDN pins go into a new
`packages` sibling. A console that predates CDN loading checks `requires` against
what its vendored runtime can supply, finds `recharts` unsatisfied, and shows its
honest "preview needs a newer Vendo runtime" tile. Had the pins gone into
`requires` alone, today's deployed console would inflate these newly-captured
components, `require("recharts")` would throw, `streaming` would swallow it, and
Maple's charts would shimmer forever.

**The console needs a follow-up PR to use this**: read `record.packages` into the
furnishing in `lib/host-components/corpus.ts`, subtract those keys in
`unsatisfiedRequires`, and re-vendor `@vendoai/ui`. Until then it degrades
honestly, as above. The console's own page CSP is `frame-ancestors 'none'` only,
so no inherited policy blocks the jail there.

## 8 — Gates

`pnpm build` ✅ · `pnpm typecheck` ✅ (43/43) · `pnpm lint` ✅

`pnpm test`: **54 of 55 tasks green.** One failure, attributed:

| run | failing task | isolated re-run |
| --- | --- | --- |
| A | `demo-accounting` — Cadence away drill (120s timeout, real Next server) | **passes** — alone AND with the whole Cadence suite (156 passed) |
| B | `@vendoai/vendo` — `audit-superset.e2e` (`Test timed out in 30000ms`) | **passes** (5 passed, 19s) |

The failing task MOVES between runs and every failure is a uniform timeout in a
real-server / live-model e2e suite — the flake #747's PR already documented ("at
default turbo concurrency the real-server fixture e2e suites flake with uniform
timeouts and the failing suite *moves* between runs"). I also checked the
direction that matters: the Cadence drill passes on `origin/main` in isolation
AND with this branch in isolation, and nothing in this diff touches automations,
guard, the store, or the wire.

**Touched packages, in isolation, all green:** core 975 · actions 634 ·
apps 738 · ui 770 (each +3/+4 for this lane's new tests).

**`pnpm --filter @vendoai/ui test:browser`: 102 passed, 20 failed — all 20
pre-existing.** Measured, not assumed: the same six spec files on `origin/main`
with this branch stashed fail **17 of 38**, and the 17 in those files with this
branch are the identical 17 (the other 3 are in `screenshots` /
`verification-eng225`, the same thread/voice/activity family). They are
thread, overlay, palette, keyboard, voice, activity and share-picker
assertions — no jail among them. Every jail-touching spec passes:
`jail-and-tree`, `exfil-probe`, `byo-embeds`, `height-chain`, `mcp-shim`'s
jail-theme test, `containment`'s jail tests, and this lane's `cdn-packages`.

---
## Summary by cubic
Preview jail can now load npm packages from one pinned CDN (`https://esm.sh`) at the exact versions the host has installed, so components like `recharts` render in previews. Production remains network‑denied; pins are preview‑only and stripped from stored/imported trees.

- **New Features**
  - `vendo sync` records `packages` (import specifier -> `name@version[/subpath]`) from installed `node_modules`; reasons are recorded for private/workspace/not‑installed packages. `requires` still lists all.
  - UI jail fetches pinned modules via an import map from `https://esm.sh` and maps `react`/`react-dom` to the jail’s own copy (CDN React is never used). CSP adds only `https://esm.sh` and `data:` to `script-src` when `packages` are present; `default-src 'none'`, `connect-src 'none'`, and no `allow-same-origin` remain unchanged.
  - Venue gate: previews send `packages`; production never produces or accepts them (`attachPinFurnishings` excludes them; `stripServerAuthoritativeFields` removes any claimed ones). Exact pins only via `jailPackageUrl`; floating/risky pins are refused.
  - Added browser/unit tests for real chart rendering, CSP, and venue gating; docs updated. Example `MapleSpendingDonut`/`MapleSparkline` are now captured with `packages: { "recharts": "recharts@3.9.2" }`.

- **Bug Fixes**
  - Stopped leaking the embedder URL via `Referer` in jailed requests with `<meta name="referrer" content="no-referrer">` (no `Referer`, no cookies/auth, `Origin: null`).
  - Previews now show “could not load <pkg@ver>” when a package fails to load instead of hanging on a loading skeleton.

<sup>Written for commit de5bf2651231fb16802f79577a615d447407692b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/771?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
## Rebased onto #775 — the two lanes compose, proven on this branch

`main` now carries the `script-src` fix (#775): the jail's nonce is gone, because
CSP blanks a nonce's content attribute but not its IDL property, so the untrusted
code could read the jail's own nonce off a script element and stamp it on a
`<script src>` of its own. Rebased onto `4515c7f74`. Two conflicts, both resolved
so **both** properties hold.

### The composed policy, read off live frames

**Production / a remix fork in a customer's page** (no packages declared) — verbatim:

```
default-src 'none'; script-src 'unsafe-inline' 'unsafe-eval'; style-src 'unsafe-inline'; img-src data:; font-src data:; connect-src 'none'
```

No nonce, and no network source at all.

**Console preview with pins:**

```
script-src 'unsafe-inline' 'unsafe-eval' https://esm.sh data:
```

The venue gate is untouched: `loadsPackages` is still the only thing that puts a
source in the directive, and it is still false everywhere but a preview.

### The two conflicts

1. **`packages/ui/src/tree/jail/JailedComponent.tsx`** — both sides, as designed:
   `'unsafe-inline'` in place of the nonce, and #771's `${scriptSources}` suffix
   with the `loadsPackages` parameter. The doc comment keeps both explanations.
2. **`packages/apps/src/open.ts`** — `stripServerAuthoritativeFields` gained
   `dataUnavailable` on main and `stripFurnishingPackages` here. Additive on both
   sides; kept both.

### Three test changes, called out because they are deliberate removals

These assert a nonce that no longer exists. Not weakened gates — two of them got
**stronger**, because a deterministic policy can be asserted exactly:

- `test/tree/jail-cdn-packages.test.tsx` L36 — `/^script-src 'nonce-[a-z0-9]+' 'unsafe-eval'$/u`
  → `.toBe("script-src 'unsafe-inline' 'unsafe-eval'")`.
- `test/tree/jail-cdn-packages.test.tsx` L50 — same, plus ` https://esm.sh data:`.
  Its nonce-normalisation helper (L42–43) is deleted: with no nonce the two
  policies are byte-identical, so the comparison is now direct.
- `e2e/cdn-packages.spec.ts` L162 — **this one was a real hole.** It read
  `` .toBe(`script-src ${directive.split(" ")[1]} 'unsafe-eval'`) `` — an
  expectation built from the string under test, to tolerate the per-mount nonce.
  It would have passed for `script-src https://evil 'unsafe-eval'`. It is now the
  literal policy. That is the test whose whole job is "the production venue names
  no network source", so it mattered.

### `scriptNonce` removed

`runtime-entry.tsx` read `document.currentScript.nonce` to stamp the injected
import map. Genuinely dead: it resolves to `""`, and `'unsafe-inline'` permits the
import map anyway (proven — the recharts draw below goes through that import map).
Left in place it would tell the next reader the jail still nonces its scripts, so
it is gone, with a comment saying why there is nothing to carry.

### Proof — both lanes' evidence, one branch, after the rebase

`packages/ui/e2e/{cdn-packages,exfil-probe,jail-and-tree}.spec.ts` — **11/11**:

- **#771, 4/4** — `MapleSpendingDonut` draws a real recharts donut from the pinned
  CDN (2 pie sectors, `Total` label, every request on `https://esm.sh`, no
  referrer/cookie/auth, `Origin: null`, 0 CSP violations); the production venue
  makes 0 CDN requests and carries the network-denied policy above; the same
  capture with `packages` removed makes 0; a blocked CDN shows the honest tile.
- **#775, 2/2** — the exfiltration probes still blocked: stolen nonce
  (`FAILURE (CSP) nonce=none` — there is no nonce left to lift), runtime-composed
  `import()`, and a remote script element, with nothing on the wire.
- **jail-and-tree, 5/5** — the pin fork with captured sub-components, root CSS and
  sample props; the raw-form intercept; the height chain; containment.

### Gates

`pnpm build` ✅ · `pnpm typecheck` ✅ 43/43 · `pnpm lint` ✅ 6/6 forced.

**Touched packages in isolation, all green:** ui 779 · apps 802 (+21 skipped) ·
core 996 · actions 647 (+4 skipped).

**`pnpm test`: 55/55 on runs 2 and 3.** Run 1 was 54/55 — `@vendoai/knowledge`
failed. Attributed, not waved off: it passes 114/114 in isolation on this branch,
it passed on both later full runs, clean `origin/main` full-suite on this same
machine is 55/55, and nothing in this diff touches knowledge. A load flake, not
this branch. Flagging it since #772 supposedly fixed the suite — if it recurs it
is worth a look, but it is not this PR.

**`test:browser`: 103 passed, 20 failed — all 20 pre-existing.** Byte-identical to
the list I measured against a clean `origin/main` tree while building #775
(accessibility ×2, chrome-behavior ×5, containment pin-mount, eng-222, keyboard
×7, screenshots ×2, verification-eng225, wave3-consumer-voice). No jail spec among
them; every jail-touching spec passes.

yousefh409:
## Greptile's P1 on `cdnPackages` is real — verified, not this rebase, and here is the patch

It flagged the same thing three times, so to save the next reader: **one finding,
and it holds.** I traced it rather than take the bot's word.

`resolvePackages` filters on the specifier alone:

```ts
const wanted = Object.entries(pins).filter(([specifier]) => !cdnPackages.has(specifier));
```

so once a realm has loaded `recharts@3.9.2`, a later render pinning
`recharts@4.0.0` is filtered out as "already loaded" and `jailRequire("recharts")`
keeps returning 3.9.2.

**Reachable**, because the realm survives the change:

- `srcDoc` is `useMemo(() => buildJailSrcdoc(loadsPackages), [loadsPackages])` — a
  **boolean** dep. A version-only change leaves the srcdoc byte-identical.
- `renderer.tsx` mounts `<JailedComponent>` with no `key`, so React keeps the same
  iframe element, the document is never reloaded, and `cdnPackages` (module-level
  in that realm) persists.
- Every render message carries `packages`, so `resolvePackages` does run again —
  and skips.

The path: a preview is open, `vendo sync` re-runs and bumps the pin, the console
re-inflates. The jail silently keeps drawing the old version. No error, no notice —
which is worse than a failure, and it is the exact failure mode this PR's
`packages-unavailable` tile exists to prevent elsewhere.

**Not caused by the rebase** — it is in the original commit, untouched by the CSP
resolution. Flagging rather than fixing: it is this lane's logic, not mine, and a
change here costs another ~30-minute CI cycle on a green branch.

**My recommendation: merge as-is, fix in a follow-up.** Preview-venue only, no
security impact, and it needs an in-place version change to bite.

Whenever it is picked up, it is four lines — key the cache by the pin, not the
specifier:

```ts
const cdnPackages = new Map<string, { pin: string; module: unknown }>();

// resolvePackages
const wanted = Object.entries(pins).filter(([specifier, pin]) => cdnPackages.get(specifier)?.pin !== pin);
cdnPackages.set(specifier, { pin, module: { __esModule: true, ...namespace, default: namespace.default ?? namespace } });

// jailRequire
const fromCdn = cdnPackages.get(specifier);
if (fromCdn !== undefined) return fromCdn.module;
```

PR #785, opened 2026-08-05
feat(store): Db.transaction() primitive with beforeWork hook

yousefh409:
## What

Piece 2 of the phase-1 store stack. Adds a `Db.transaction()` primitive with a
`beforeWork` hook, implemented for both the Postgres and PGlite drivers.

## Why

The local `StoreOps` backend (#T4) has to make each verb atomic — the F4
cascade (deleting a thread takes its messages and harness state with it) is
several statements that must not half-apply. There was no transaction boundary
to hang that on, so the primitive comes first, on its own, with its own tests.

`beforeWork` exists so a caller can claim a lock or set a session variable
inside the same transaction before the body runs.

## Verification

Gates were run and verified at descendant SHA `aa88d773` (this commit is an
ancestor): `@vendoai/store` 431 pass / 2 skip, `@vendoai/core` 1055 pass,
build 24/24, typecheck 43/43, lint clean. CI re-runs them against current
`main` on this PR.

## Status

Proven: gates above, plus the 6 dedicated `db.transaction.test.ts` cases.
Still running: the live cut-down acceptance and AI reviewer triage. Does not
merge until gates are green, the acceptance passes, and every reviewer comment
is triaged.

---
## Summary by cubic
Add a `Db.transaction()` primitive with an optional `beforeWork` hook for Postgres and PGlite. It makes verbs atomic and supports session-scoped setup (meets Linear T4 and safe F4 cascade deletes), and it now rejects on a closed store.

- **New Features**
  - New `Db.transaction(work, { beforeWork? })` with commit/rollback; returns the work result.
  - Postgres: dedicated client per transaction; no shared-pool tx statements; client released.
  - PGlite: uses its serialized `transaction()` wrapper.
  - `beforeWork` runs after BEGIN and before work; use for `SET LOCAL` (RLS), visible to work.

- **Bug Fixes**
  - Postgres: `transaction()` rejects after `close()`; tests cover both engines and a pg seam pins statement order, release, and the no-acquire-on-close case.

<sup>Written for commit 51b171f739661f1b60530ba85b0e8a70e03fb98c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/785?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Closed both P2 findings (PGlite-only transaction coverage) in fe4dfa54:

- `db.transaction.test.ts` now parameterizes the whole suite over both engines: the PGlite `memory://` leg always runs, and a `POSTGRES_URL`-gated leg runs the identical assertions against real Postgres in CI (same gating pattern as `postgres-entry.test.ts`). A new test proves `beforeWork` runs after BEGIN by asserting `SET LOCAL` visibility inside the work callback — the RLS tenant-context pattern the hook exists for.
- New `db-postgres.transaction.test.ts` pins the pg engine's transaction lifecycle without a server via a mocked `pg` module: one dedicated pool client per transaction, statement order BEGIN → beforeWork → work → COMMIT, ROLLBACK on work failure and on beforeWork failure (work never runs), `release()` called exactly once on every path, and no transaction statements riding the shared pool.

Gates: `@vendoai/store` tests 390 passed / 2 skipped, typecheck and lint clean. The postgres leg is skipped locally (no POSTGRES_URL here) and will run in CI where it is provisioned.

yousefh409:
Fixed the closed-handle blocker: `transaction()` on the pg engine now throws `[vendo] store is closed` before acquiring a client, mirroring `query()`'s check (1b2d82af). PGlite already rejected via its `open()` gate. Added a rejects-after-close test on both engines in `db.transaction.test.ts` plus a pg-seam test asserting no client is acquired and work never runs. Also merged latest main. Store suite 392 passed, typecheck and lint clean.

PR #786, opened 2026-08-05
feat(core): the StoreOps conformance suite and its memory reference

yousefh409:
## What

Piece 3 of the phase-1 store stack. Adds `storeOpsConformance` — one shared,
backend-agnostic test kit (40 cases) that any `StoreOps` implementation must
pass — plus an in-memory reference implementation that passes it, and wires the
kit into contract coverage.

## Base

This PR targets `store/phase1-base`, which is the merge of the two independent
parents (`store/wire-interface` + `store/db-transaction`), each reviewed in its
own PR. That keeps this diff **exactly** this piece rather than replaying the
two below it. Retarget to `main` once those merge.

## Why

Three separate backends land this phase — in-memory, local SQL, and the HTTPS
cloud client — and a fourth counterparty (the console doors) sits behind the
wire. If each one brings its own tests, they can disagree forever without any
suite going red. One kit run against every backend is what makes "the local and
hosted stores behave identically" a checked claim instead of a hope.

The memory reference exists so the kit itself is proven executable against a
trivially-correct implementation before it is pointed at real infrastructure.

## Verification

Gates were run and verified at descendant SHA `aa88d773`: `@vendoai/core`
1055 pass (includes the memory reference against this kit), `@vendoai/store`
431 pass / 2 skip, build 24/24, typecheck 43/43, lint clean. CI re-runs against
the base on this PR.

## Two amendments to the kit, stated because they touch test strictness

Both were made by the local-backend piece (#T4) and audited hunk by hunk
against this tip. **No assertion was weakened.**

1. **Fixtures made shape-valid** — thread ids `t1` → `thr_t1`, and the
   `lifecycle.promote` fixture now seeds a shape-valid `vendo_apps` record. The
   old fixtures asked every backend to accept input that core's *own*
   `threadIdSchema` declares illegal; a typed backend correctly refuses them.
   This is a tightening, and it also closes the `thr_*`-vs-`t1` grammar
   mismatch the console doors flagged from the other end of the wire.
2. **`assertDeepEqual` → canonical JSON compare** — relaxes object **key order
   only**, because Postgres `jsonb` does not preserve it and a byte-exact
   compare is unimplementable over it. Missing keys, extra keys, types, and
   **array order** all remain significant, so the pagination and list-filter
   cases keep their teeth.

## Status

Proven: gates above; the kit runs green against the memory reference and (in
#T4) against the local SQL backend. Still running: the kit through the **real**
cloud client into the **real** console doors (the seam test, vendo-web
`store/t7-seam`) — that is the proof that matters most and it is in flight.
Also running: live acceptance and AI reviewer triage. Does not merge until all
three are done.

---
## Summary by cubic
Add a backend-agnostic StoreOps conformance suite, the `store-wire` protocol with a named `StoreOps` interface, and an in-memory reference that passes the suite. This locks the 32‑op store contract and verifies pagination, idempotency, cascades, and undo semantics across backends.

- New Features
  - Introduced `storeOpsConformance` (41 cases) covering all 32 ops across 7 families, including deterministic cursors and the 100 default / 1000 max page caps, idempotent `workspace.commit`, and F4 cascade on `transcripts.deleteThread`.
  - Added `store-wire` protocol and a named `StoreOps` interface; kept the erase-target contract as exactly-one-of `subject` or `appId`.
  - Added `memoryStoreOps`, a reference implementation that passes the suite; exported `storeOpsConformance` and `memoryStoreOps` from `@vendoai/core/conformance` and included them in contract coverage.

- Bug Fixes
  - Fixtures use `thr_`-prefixed thread ids and a schema-valid `vendo_apps` seed; equality uses canonical JSON so key order is ignored; tests ensure the suite passes a schema-validating and a key-order-scrambling backend.
  - `workspace.undo` now preserves later writes when undoing an older commit; a new conformance case enforces this ordering.

<sup>Written for commit 51dd3fba3dd3ae3fdab0b33768b84e7092467b5a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/786?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Closed all three P1 findings in 2d4bb7aa:

- **Invalid transcript fixture IDs**: every transcript fixture (round-trip, subject filter, pagination, delete-cascade, putMessage, recordAnswer, and the erase/adopt lifecycle threads) now uses `thr_`-prefixed IDs that pass `threadIdSchema`.
- **Invalid reserved app fixture**: the promotion case now seeds `vendo_apps` with a full schema-valid payload — `{ subject, enabled, doc: { format: "vendo/app@1", id: "app_promote", name: "Promoted" } }` — so `parseAppData` accepts it and `lifecycle.promote` is actually exercised.
- **Key-order false failures**: `assertDeepEqual` in `store-ops.ts` now compares `canonicalJson` output (with the same undefined sentinel as the canonical helper in `conformance/index.ts`) instead of raw `JSON.stringify`, so object key insertion order no longer matters while array order still does.

Two new kit tests pin the contract and reproduce your findings: the suite must pass (1) a schema-validating backend that enforces `threadIdSchema` and app-payload validation on reserved writes, and (2) a backend whose reads return records with reversed key insertion order. Both fail on the pre-fix suite (verified by stashing the fix: 2 failed / 43 passed) and pass with it.

Gates: `@vendoai/core` 1057 tests passed, typecheck and lint clean.

yousefh409:
Closed the P1 data-loss finding in 51dd3fba: **workspace undo no longer destroys later writes**.

**Fix** — `memoryStoreOps().workspace.undo` restored every path of the undone commit from its preimage, so undoing an older creation commit deleted a path a later commit had since rewritten. Undo now drops the target commit and rebuilds workspace state by replaying the retained commits (the reviewer's suggested rebuild option), so it reverts only the undone commit's own effect and a later write to the same path always wins. The per-commit `before` preimage map is gone with it.

**Contract pinned** — new StoreOps conformance case "workspace.undo of an older commit preserves later writes": commit `p=first`, commit `p=second`, undo the first — `p` must still read `second`, and a path only the undone commit wrote must revert. Verified red on the pre-fix reference (4 failed / 42 passed, failing with `undefined !== {"v":"second"}` — the exact reproduced loss) and green with the fix. No conflict variant: `undo` returns `Promise<void>`, so the pinned semantics is skip/later-write-wins rather than a reported conflict.

Gates: @vendoai/store 386 passed, @vendoai/core 1059 passed, typecheck and lint clean.

PR #784, opened 2026-08-05
feat(core): store-wire protocol and StoreOps named-op interface

yousefh409:
## What

Piece 1 of the phase-1 store stack. Adds the `vendo/store-wire@1` protocol
(envelope, error codes, status shape, operation manifest) and the `StoreOps`
named-op interface — 32 operations across 7 families — in `@vendoai/core`.

No behaviour changes: this is the contract every later piece implements or
consumes.

## Why

Phase 1 replaces ad-hoc store access with one named-op surface so a Cloud-key
host can persist without holding database credentials. The interface has to
exist and be agreed before the local backend (#T4), the cloud client (#T6),
the console doors (vendo-web) and the conformance kit (#T3) can be written
against it.

## Verification

- [x] `pnpm build && pnpm test && pnpm typecheck && pnpm lint` green
- [ ] Screenshots attached (if UI-affecting) — n/a, no UI

Gates were run and verified at descendant SHA `aa88d773` (this commit is an
ancestor): `@vendoai/core` 1055 pass, `@vendoai/store` 431 pass / 2 skip,
build 24/24, typecheck 43/43, lint clean. CI re-runs them against current
`main` on this PR.

## Known gap, disclosed not buried

`STORE_WIRE_PATHS` in this commit **disagrees with the shipped console server
for 4 of the 7 families** — records is `/records/{collection}/{op}` not
`/records/get`; blobs are REST raw-byte routes, not POST+base64; erase takes
its target flat; erase/adopt/sessions are not under `/lifecycle/*` at all.

The constant has **zero production consumers** — only its own test asserts it,
which is exactly the producer-and-consumer-mock trap this repo has a law
about. It is made truthful *and* actually consumed by the client (so it cannot
drift again) on `store/phase1-integrate`, which is in flight. Reviewers should
read that constant as not-yet-load-bearing.

## Status

Proven: gates above. Still running: the live cut-down acceptance and AI
reviewer triage. Per the merge bar, this does not merge until gates are green,
the live acceptance passes, and every reviewer comment is triaged against the
code.

---
## Summary by cubic
Adds the `vendo/store-wire@1` HTTP protocol and the `StoreOps` named-operation interface to `@vendoai/core` for a 32-op store contract. Also enforces that `lifecycle.erase` names exactly one scope (`subject` or `appId`) at both type and schema levels. No runtime behavior changes.

- **New Features**
[...]

PR #787, opened 2026-08-05
feat(store): the StoreOps local backend, with transactions at verb boundaries

yousefh409:
## What

Piece 4 of the phase-1 store stack. Implements `StoreOps` over the existing
`@vendoai/store` helpers — all 32 operations — with `Db.transaction()` at every
verb boundary, and runs the shared conformance kit against it.

Targets `store/conformance-suite` (#T3), so this diff is exactly this piece.

## Why

This is the no-key path. A host that brings its own Postgres gets the same 32
named operations the Cloud client exposes, executed locally, with each verb
atomic. Without it the named-op surface would only exist over HTTPS, and the
hard BYO rule says every single-player capability keeps a no-key path.

## Verification

Verified by the orchestrator, not taken from the worker's word:
`@vendoai/store` 431 pass / 2 skip — including `ops.conformance.test.ts`, the
full 40-case `storeOpsConformance` green against this backend — `@vendoai/core`
1055 pass, build 24/24, typecheck 43/43, lint 0 errors. Worktree clean at
`aa88d773`.

## A real bug this piece found and fixed

An earlier draft of this work shipped `harness.set` as
`DELETE FROM vendo_state WHERE app_id = $1` — which wipes **every subject's**
row under an app, not just the one being written. The key is `(appId, subject)`
per three independent sources in this repo plus core's own reference
implementation. The draft's own (untracked, never-committed) test had enshrined
the bug; it was replaced. No committed test file was touched.

## Deferred, each with a reason — none are in phase-1 scope

- `getThread` ignores its `{cursor, limit}` opts. Core's reference ignores it
  identically, and so does the console door — so this is a gap in the
  **contract's** coverage, not a divergence between implementations. Confirmed
  independently from the cloud-client end too.
- `lifecycle.adopt` is 3 statements, not 1 transaction; the underlying helper
  linearizes on the session claim.
- `workspace.index` uses a direct keyset query because the existing helper
  returns everything unpaginated.

## Gap, stated not buried

`POSTGRES_URL` is unset on the machine that ran the final gates, so only the
**PGlite** leg of the store suite executed at this SHA. (An earlier worker
reported the kit 40/40 on real Postgres via docker at an earlier SHA, but that
is not reproven here.) Real Postgres is exercised end-to-end through the
console doors by the seam test in vendo-web `store/t7-seam`, and CI runs both
legs on this PR.

## Status

Proven: gates above. Still running: the live cut-down acceptance and AI
reviewer triage. Does not merge until gates are green, the acceptance passes,
and every reviewer comment is triaged against the code.

---
## Summary by cubic
Adds a local `StoreOps` backend that runs all 32 ops directly on Postgres with per-verb transactions, closing the F4 orphaned-messages gap and matching the cloud client. Aligns with core’s new `store-wire` protocol and the erase-target contract; targets `store/conformance-suite` (#T3) and runs the shared `@vendoai/core` conformance suite.

- **New Features**
  - Introduces `createStoreOps`: serves all 32 ops locally via `@vendoai/store` helpers, with `Db.transaction()` for multi-statement verbs.
  - Workspace commits are idempotent and undoable: writes a `vendo_workspace_commits` ledger, replays by key, and undoes in one transaction.
  - Blobs touched inside a verb use a tx-scoped files adapter (`storeFilesForDb`) so blob ops ride the same transaction.
  - Conformance updates: canonical JSON equality for `jsonb`, thread IDs `thr_` prefix, and `vendo_apps` promote seeds a shape-valid record.
  - Adopts core’s `store-wire` interface and enforces the erase-target rule: exactly one of `subject` or `appId`.

- **Bug Fixes**
  - `transcripts.deleteThread` now atomically deletes the thread, its `vendo_thread_messages`, and its harness state (closes F4).
  - `harness.set` writes via shared `putStateRow`, correctly keying on `(appId, subject)`.
  - Prevents PGlite deadlocks by issuing blob operations through the verb’s tx-scoped query.
  - Fixes `workspace.commit` idempotency race by claiming the ledger row before any workspace writes.

<sup>Written for commit a013ddacaf96fc6f55c528a47f352811eaa5402c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/787?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Fixed the P2 idempotency race in fe03a61.

**What changed**
`workspace.commit` now claims the ledger row FIRST, inside the same transaction as the mutation: an `insertIfAbsent` on the commit ledger (unique `(collection, id)` key, where the id derives from the idempotency key) is the serialization point. Only the successful claimant runs `txRows.land(...)`; the loser's insert conflicts, and it then reads the recorded row — matching request body → returns the recorded (replay) result, different body → `conflict`. The mutation can never run before the claim succeeds; a failed land rolls the claim back with the transaction.

**Proof**
- New test in `ops.test.ts` fires two same-key commits with different bodies concurrently and asserts exactly one lands, the other gets `conflict`, exactly one body stays recorded, and the workspace file matches the winner.
- Verified the test actually catches the bug: with the fix reverted, it fails on the real-Postgres backend leg with BOTH commits fulfilled (PGlite's serialization masks the race, exactly as the finding said); with the fix, both the pglite and postgres legs pass.
- `@vendoai/store` suite 432/432 green, repo typecheck and lint green.

yousefh409:
Merged latest main and resolved the conflicts, keeping main's erase-target contract from #784: `store-wire.ts` (the `{subject}|{appId}` union and refined schema) is byte-identical to main, and this branch's `lifecycle.erase` dispatches on that union. The ledger-claim-before-write fix (fe03a61c) is untouched. Store suite 432 passed, typecheck and lint clean. Pushed a013ddac.

PR #788, opened 2026-08-05
feat(vendo): the full 32-op cloud store client over vendo/store-wire@1

yousefh409:
## What

Piece 5 of the phase-1 store stack. `hosted-store.ts` implements the full 32-op
`StoreOps` over HTTPS against the console's store doors: retry with the **same**
`Idempotency-Key`, cursor passthrough, envelope → `VendoError`, `501` → a clean
`unsupported_operation`, and never partial execution.

Targets `store/conformance-suite` (#T3), so this diff is exactly this piece —
2 files, `hosted-store.ts` +347 and `hosted-store.test.ts` +584.

## Why

This is the Cloud path, and the point of the phase: an agent workspace that
persists with **zero database credentials host-side**. The client is the only
component that turns the named-op interface into wire traffic, so it is also
where version skew has to be handled gracefully rather than crashing a host.

## Verification

Verified by the orchestrator: commit touches exactly the 2 files above (the
shared conformance kit deliberately untouched — #T4 owns it). `@vendoai/vendo`
2136 pass / 23 skip, build 24/24, typecheck 43/43, lint 6/6.

## The honest limit of these tests

**Both test surfaces mock the counterparty** — an injected fake `fetch` and a
`fakeConsole`. Per this repo's law, that proves the client talks to *itself* and
nothing more. The real proof is the seam test on vendo-web `store/t7-seam`,
which runs the shared conformance kit through this client into the **real**
console handlers, real Postgres and the real ledger, with no stub on either
side. That is in flight; **this PR does not merge before it is green.**

## Findings reported rather than silently fixed — these are decisions

1. **`STORE_WIRE_PATHS` (#T1) is wrong for 4 of 7 families.** The client
   followed the *real shipped server* instead, which was the right call. The
   constant has zero production consumers today. It is made truthful **and**
   actually consumed by this client — so it cannot drift again — on
   `store/phase1-integrate`, in flight.
2. **The idempotency ledger covered only 3 of 7 families.** `withIdempotency`
   was wired into transcripts/harness/workspace/`lifecycle.promote` but **not**
   records, blobs, `/erase`, or `/sessions/*`. This client sends an
   `Idempotency-Key` on all 20 mutations; those doors ignored it. `put` and
   `blobs.put` are naturally idempotent, so the real exposure was
   `claim`/CAS/`sessions.claim`: a timeout retry can report `claimed:false`
   after the first attempt actually won. Spec amendment 2 requires idempotency
   on **every** mutation, so this was a shipped gap against the approved spec —
   Yousef approved closing it, and it is closed on vendo-web
   `store/t8-idempotency`.
3. **`transcripts.getThread` drops its `{cursor, limit}` server-side** — the
   same contract gap #T4 found from the local end. Confirmed from both ends.
4. **A `401` degraded to not-implemented.** `"unauthorized"` is not a
   `VendoErrorCode` and `STATUS_TO_CODE` had no `401` entry, so a revoked key
   reported as an unsupported operation. Bug; mapped correctly on
   `store/phase1-integrate`.

## Status

Proven: gates above, against mocked counterparties only. Still running: the
seam test (the real proof), the live cut-down acceptance, and AI reviewer
triage. Does not merge until all three are done.

---
## Summary by cubic
Implements the full 32-operation cloud client `hostedStoreOps` over `vendo/store-wire@1` and adds `.ops` to `hostedStore`, using the same mount and API key. Enables HTTPS StoreOps with idempotent retries, cursor passthrough, and clear error mapping; supports #T3 conformance target.

- **Bug Fixes**
  - Records and blobs now use `STORE_WIRE_PATHS` and the v1 request schemas (collection/namespace/key in JSON; blob bytes base64) instead of per-collection routes/REST bytes.
  - `/erase` validation enforces exactly one of `subject` or `appId`, matching the v1 contract.

<sup>Written for commit 271ea71feeebdf404070204908c57cce86f00974. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/788?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
P1 closed in 271ea71f: the client now speaks the protocol it declares.

**Fix** — `hostedStoreOps` records and blobs no longer use the console's legacy doors (per-collection `/records/{collection}/{method}` routes, REST blob bytes). All 11 ops now POST to their `STORE_WIRE_PATHS` route with bodies matching the exported `storeWire*RequestSchema`s: `collection`/`namespace`/`key` ride the JSON body, blob bytes are base64 on the wire. Both mismatches from the T-Rex capture are gone — `records.get` sends `POST /records/get` with `{collection,id}`; `blobs.put` sends `POST /blobs/put` with `{namespace,key,bytes:base64,contentType?}`. No schema was touched; the wire contract stayed the source of truth.

**Seam test** — a new contract test drives every records and blobs op and parses each outgoing request body with its corresponding store-wire zod schema, so the client is checked against the exported contract rather than against itself. This class of drift can no longer pass a green suite.

**Gates** — `@vendoai/vendo` tests, typecheck, and lint all green (2130+ passed; 14 initial failures in unrelated LLM-driven e2e files were an exhausted API key — all 170 tests in those 7 files pass on a fresh key).

PR #789, opened 2026-08-05
feat(guard): freeze flag — the emergency stop

yousefh409:
## One row, checked first

The emergency stop is a single store row: `freeze` in the guard's own
`guard:controls` collection, holding `{ frozen, by, at }`. `#checkWithMetadata`
reads it at the very top, before risk resolution, before the pipeline, before
any breaker is spent.

## Blocks everything while set

While the flag is set, every check comes back
`{ action: "block", decidedBy: "frozen" }` — a declared read under no policy at
all, a call a standing grant authorizes, an approved replay. Nothing is spent on
the way out: no breaker slot, and no parked approval left behind for someone to
answer later. A freeze refuses; it does not queue.

## The console flips it

It is a row and not a config field on purpose: the moment you need a kill switch
is the moment you cannot redeploy to get one. The console writes the same row
straight through the store, and a guard in another process obeys it on its very
next check. `guard.freeze(by)` / `guard.unfreeze(by)` / `guard.frozen()` are the
in-process seam onto the identical row.

## Audited both directions

Freezing and unfreezing each write a `policy-decision` audit event naming who
flipped the switch (`detail.reason` = `frozen` / `unfrozen`), and every call the
freeze refused is audited exactly as any other block is. `@vendoai/core`'s
`GuardDecision` block arm and `AuditEvent` gain the `"frozen"` provenance,
schemas included.

## Tests

`packages/guard/test/freeze.test.ts` — real PGlite store, no mocks, and it
covers the seam in both directions: the row `freeze()` writes is read back
through raw SQL as the console will read it, and a row written straight through
the store freezes the guard. Proven able to fail: with the gate in
`#checkWithMetadata` disabled, all three tests go red (run/default instead of
block, 2 audit events instead of 3, `ok` instead of `blocked`).

Retargets to `main` once `yousefh409/agents-runtime` merges.

---
## Summary by cubic
Adds an emergency freeze flag to `@vendoai/guard` that stops all checks and executions immediately. Also chips `@vendoai/core` audit events with the effective risk for descriptor-backed rows; frozen-path audits are best-effort and never break the freeze.

- New Features
  - Guard reads `guard:controls/freeze` first; while set, blocks all calls with decidedBy "frozen" and spends nothing (no risk, breaker, or parking).
  - New `VendoGuard` APIs: `freeze(by)`, `unfreeze(by)`, `frozen()`.
  - Audit: add "frozen" to `GuardDecision`/`AuditEvent`; chip `risk` on descriptor-backed rows (tool-call, approval park, policy-decision blocks, grant-invalidation, org-policy-unavailable) using the effective grade. Control flip rows carry none.

- Bug Fixes
  - Execute re-reads the freeze before dispatch to close the check-to-execute race; frozen-path audit writes are best-effort if `vendo_audit` is down.
  - Frozen block rows omit `risk`. Absent control row is unfrozen, but unreadable/corrupt row or a read error fails closed (treated as frozen) and is audited.

<sup>Written for commit c41025dec6c8b5212942a0ca270c1d0e0b8d3776. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/789?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
**Triage (round 2, SHA 363f72879):**
- **Freeze bypass during policy eval (P1)** — REAL, fixed. `bind().execute` re-reads the freeze row immediately before touching the registry; a freeze landing mid-pipeline or after the run verdict now blocks. Test proves freeze between a "run" check and execute → tool never runs.
- **Frozen audit write breaks block response (P1)** — fixed. Both frozen paths audit best-effort now; the block is always returned even if `vendo_audit` is down. Test: audit.put throws → still blocks, no exception.
- **Frozen records declared vs effective risk (P1 security)** — already fixed in an earlier commit: the frozen row omits `risk` (asserted `not.toHaveProperty("risk")`), since the fast path short-circuits before risk resolution.
- **Required freeze members break typed mocks (P1)** — not actually breaking the gate (agents excludes \*.test.ts from typecheck), but no-op stubs added to both mocks so they honestly satisfy their VendoGuard annotation.

PR #791, opened 2026-08-05
feat(agents): @vendoai/agents — a governed agent in any Node backend

yousefh409:
Standalone runtime for spawning a governed, harness-grade agent, plus the seams it needed in the existing packages.

## What's on this branch
- **`packages/agents`** — the new standalone runtime: `agent()` / `session()` / `stream()`, approval events, `tool()` and `api()` sources, MCP sources, folder skills loader, agent-level egress with an audit row per box boot, and the store/sandbox resolution ladders.
- **`door: { baseUrl }`** — optional config key, falling back to `VENDO_BASE_URL`; loopback only for `machine: "local"`. A sandboxed agent with no reachable origin **refuses at boot** with an actionable error instead of silently having no tools.
- **guard** — `GuardLike` contract extracted, `onApprovalRequested` added, dead exports removed.
- **core** — `RunContext` carries the host's user profile, context bag and the turn's transcript; a model seat is required only where a harness reads it; per-turn `model`/`effort` overrides removed (they bind at construction).
- **claude-code** — `template` option, `VENDO_API_KEY` inference rung, hard-coded Vendo-voice briefing removed.
- **knowledge** — `lexicalKnowledge` → `vendoKnowledge`; the prompt note moved into the knowledge package.
- **delete-sweep** — verified zero-caller code removed (incl. the orphaned workspace-bash wrapper).
- **`createVendo({ agent })`** — the graduation seam, and the sandbox ladder moved into one shared module used by both packages.

## Proven / not proven
- Package and scoped gates green locally (`@vendoai/agents` 52/52); full suite run on the build machine.
- **Live BYO acceptance run is PENDING** — real box, real approval parked and resumed, files persisting across two sessions.
- **Cloud acceptance leg is BLOCKED** on the separate store project (the hosted store cannot serve a workspace until its typed doors land).

Do not merge until CI is green and reviewer findings are triaged.

---
## Summary by cubic
Introduces `@vendoai/agents`: a standalone runtime to spawn a governed, harness-grade agent in any Node backend, with a first-class MCP tool door and a shared sandbox ladder. Vendo’s embed can now adopt a whole agent via `createVendo({ agent })` for the same brain, store, and sandbox across chat and server; approval events are now scoped to the originating session and subscriptions are cleaned up when handlers detach.

- **New Features**
  - `@vendoai/agents`: `agent()`/`session()`/`stream()`, `tool()`/`api()` sources, external MCP via `mcp: [...]`, skills folder loader, egress allowlist with one audit row per box, prompt assembly, and guard binding.
  - Tool door: `door: { baseUrl }` (falls back to `VENDO_BASE_URL`; for `machine: "local"` a loopback door is auto-served). Exposes a fetch handler at `DOOR_PATH` for hosts to mount.
  - Shared sandbox ladder (`@vendoai/apps/sandbox-ladder`): explicit adapter > `E2B_API_KEY` > `VENDO_API_KEY` > none. Both `@vendoai/agents` and `@vendoai/vendo` use the same ladder.
  - `createVendo({ agent })`: adopts a full agent from `@vendoai/agents`; conflicting slots (`harness`, `store`, `files`, `sandbox`) error at boot; guard and host tools remain the deployment’s.
  - Guard: extracted `GuardLike` seam; `onApprovalRequested` event for parked approvals.
  - Core: `RunContext` now carries `user`, `context`, a `messages()` accessor, and preserves `turnId`; model seats are relaxed (only seats read by a harness are required).
  - Harnesses: `claudeCode` adds `template`, infers `VENDO_API_KEY` as a third inference rung, and keeps only `maxTurns` as a per-turn option.
  - Knowledge: `lexicalKnowledge` renamed to `vendoKnowledge`; prompt note lives in `@vendoai/knowledge`.
  - MCP: turn-credential registry moved to `@vendoai/mcp` and used by the internal door.
  - Corpus: local pack and inject fixtures include `@vendoai/agents` to avoid broken installs.

- **Migration**
  - Replace `lexicalKnowledge()` with `vendoKnowledge()`.
  - `claudeCode` per-turn `model`/`effort` removed; set them at construction. Only `maxTurns` is per-turn.
  - Risk grading: the dev’s declared `risk` is final; the second mechanical vote is removed. Label destructive actions (e.g. messaging humans) as `destructive`.
  - `@vendoai/store`: removed `workspaceBash()` and its types (breaking). No in-repo callers.
  - `createVendo({ agent })`: accepts `AgentOptions | ComposedAgent`. Do not pass `harness`/`store`/`files`/`sandbox` alongside an agent.
  - Boxed harnesses need a reachable origin for the tool door. Set `door.baseUrl` or `VENDO_BASE_URL`. Local `machine: "local"` auto-binds loopback.

<sup>Written for commit 61cf3f9bb03b7f56f87925af51dfad26efc385a6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/791?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Final gates on the shipping tree (branch tip `4ec2c07a`, with `origin/main` `e0962e45` merged in via `40351b87`):

- **build**: exit 0
- **typecheck**: exit 0
- **lint**: exit 0
- **test**: 58/58 turbo tasks green — **8756 passed | 89 skipped | 0 failed** (VITEST forks/threads capped at 2, per main's `edc2ab89`)

Two fixes landed during gating, both from the same root cause: the umbrella now depends on the new `@vendoai/agents` package, but the corpus harness's pack closure didn't know it. `fcc9d69a` adds it to `LOCAL_VENDO_PACKAGE_NAMES` (caught by `local-pack.test.ts`), and `4ec2c07a` mirrors it into the `inject.test.ts` fixture monorepo, which discovery now rightly rejects without it.

Full-suite note for honesty: everything ran fresh at `fcc9d69a` (57/58 green; corpus-harness red on the stale fixture), corpus-harness then re-executed fresh and green at `4ec2c07a` (401 passed | 2 skipped); the final `pnpm test` at the tip is 58/58.

yousefh409:
The P1 approval-isolation failure is closed at `b0de28ef`.

**What was wrong.** Every `session()` subscribed to the shared guard's `onApprovalRequested` without scoping, so a guarded action parked in one conversation surfaced in every other session's `on("approval")` handler — another user's pending action, preview included, with live approve/deny closures. The subscription was also never released, so a dead session's callback outlived it on the guard.

**The fix.** The guard has always recorded the parking conversation in `ApprovalRecordData.sessionId` (from `RunContext.sessionId`); that identity now rides the emitted request too, as `ApprovalRequest.ctx.sessionId` — optional only because rows persisted before it existed cannot carry it. A session delivers to its handlers only when the request names its own thread, so an **ownerless request matches nothing and delivery fails closed**. The guard subscription is taken on the first `on()` handler and released with the last. Deciding was and remains owner-scoped: a foreign principal's `decide` is `not-found`.

**The regression proof** is `session.test.ts` → "delivers an approval ONLY to the conversation that parked it, and a foreign session cannot decide it": three sessions on a real `createGuard` and a real agent — two of them *the same user on different threads*, to pin that scoping is per **thread**, not per subject — asserting the foreign handlers never fire and cannot decide. No stub on either side. The pre-existing case in that file also changed, deliberately: delivery now requires an owner, so its hand-built request must name its own thread to be surfaced at all. That is stated in the commit.

### Gates at `b0de28ef`

- **build**: exit 0 (25/25)
- **typecheck**: exit 0 (45/45)
- **lint**: exit 0 (6/6, 1 pre-existing `demo-bank` warning)
- **test**: **56/58 turbo tasks green — 8751 passed | 93 skipped**, in 28m40s

Everything in the change's blast radius is green in that full run: **core 947**, **guard 257 | 3**, **vendo 2082 | 23**, **agents 62**, apps 809 | 21, harnesses 309 | 10.

**Being straight about the two red tasks — neither is a regression, and I did not take that on faith.** `@vendoai/redteam-e2e` and `demo-bank` both failed on wall-clock, not on an assertion about changed behaviour, and both pass standalone on the same commit on an idle machine:

- `demo-bank` → `away-drill.test.ts` died in `beforeAll` with `Hook timed out in 240000ms` while booting its dev server. Alone: **26 files, 111 passed, exit 0**, with the three away-drill cases taking 6–10s each. A 240s boot budget versus 10s of real work is the gap.
- `@vendoai/redteam-e2e` → two cases reported a run reaching `error` instead of `ok`. Alone: **24 passed | 1 skipped, exit 0**.

Both were `Cached` in the previous tip's run, so this is the first time either actually executed against the merged tree — the cache was hiding them, and what it was hiding is a capacity problem on this 8-core box, not a defect. Worth noting for its own change: that 240s hook budget is exactly the "wall-clock budget tighter than the machine can honour" hazard the repo's testing rules call out, and it will flake under parallelism on any similarly-sized runner. CI is the arbiter on that, not this box.

### Acceptance

- **BYO leg: PASSED live**, 9/9 — real e2b box on a built template, real Postgres in Docker, real cloudflared tunnel as clearly-labelled test scaffolding, guarded tool call parked and resumed through a real approval, and a nonce written in one session read back verbatim by a separate session.
- **Cloud leg: BLOCKED**, unchanged — the P4 hold. `hostedStore` is HTTP-only and fails `storeServesHarnessTurns`, so `agent()` keeps failing loudly and naming `store: postgres(url)`. One line in `server.ts` fills it when P4 unblocks.

**Not merging.** This needs a maintainer's read on the two capacity-flaky tasks above.

yousefh409:
@greptile-apps review

Re-review requested explicitly: the automatic pass at 09:40 was skipped for the 101-file limit, so nothing since `4ec2c07a` has been looked at. Two things landed after your last review that are worth your attention:

- `b0de28ef` — the fix for your P1 approval-isolation finding. Delivery is now scoped to the parking conversation via `ApprovalRequest.ctx.sessionId`, failing closed on an ownerless request, and the guard subscription is released with the last handler. Please check the fail-closed path and the decide path specifically.
- `61cf3f9b` — `origin/main` merged in, resolving a real semantic conflict on `RunContext` (main's `turnId` unioned with this branch's `user`/`context`/`messages`).

The full four-gate is running against `61cf3f9b` now; results will follow in a separate comment.

yousefh409:
### Full four-gate on the current tip `61cf3f9b` (main merged in)

Only scoped gates had run on this tip, so this is the first time the whole suite has seen the merged tree — worth it, because it surfaced something the scoped gates could not.

- **build**: exit 0 (25/25)
- **typecheck**: exit 0 (45/45)
- **lint**: exit 0 (1 pre-existing `demo-bank` warning)
- **test**: **57/58 turbo tasks — 8805 passed | 89 skipped**, 32m5s

The merge resolution holds up under the full suite: **core 966**, **vendo 2096 | 23**, **harnesses 324 | 10**, **guard 257 | 3**, **store 376 | 2**, **agents 62**. Those counts are up from the previous tip exactly where main added coverage, which is what a clean union of the two `RunContext` sides should look like.

**The two capacity flakes from the last run are gone.** `demo-bank` 111 passed and `@vendoai/redteam-e2e` 24 passed | 1 skipped, both green in this run on an idle machine — confirming they were contention, not defects.

**The one red is a 1-millisecond wall-clock flake, and I verified that rather than assuming it.** `@vendoai/knowledge` → "the per-TURN verification budget (K15 T3) > gives the deep retry only what the chat verification left": `expected 1701 to be less than or equal to 1700`. The test gives the turn a 2000ms budget, has the first verification `await setTimeout(300)`, then asserts the retry's remaining budget is `<= 2000 - 300`. It pins an exact wall-clock boundary, so a timer firing a fraction under 300ms — or an elapsed measurement truncating 299.6 to 299 — leaves 1701 and fails by one. Nothing in this branch touches that arithmetic.

Evidence: that single case passes **5/5** consecutive isolated runs, and the package is **124/124, exit 0** standalone. It failed only inside the 32-minute parallel run.

Same family as the `demo-bank` 240s hook budget I flagged earlier: both pin wall-clock margins that a loaded runner can't honour. Each deserves its own small test-only change — widening that boundary, or measuring consumed budget from the same clock the timer uses. I have deliberately not touched either here; test changes belong in their own commit, not smuggled into a fix.

### Where this leaves the PR

Gates are green apart from one proven-flaky assertion. The **BYO acceptance leg passed live 9/9**; the **Cloud leg stays blocked** on the P4 store hold. Greptile's automatic re-review was skipped for the 101-file limit, so I've tagged it explicitly — **the P1 fix and the main merge have not yet been reviewed by anything but me.**

**Not merging.** Outstanding: that re-review, and a maintainer's call on the two wall-clock-flaky tests.

yousefh409:
### Independent adversarial review of the P1 fix

Greptile's automatic re-review was skipped for the 100-file limit, the `@greptile-apps review` tag has produced nothing in ~90 minutes, and its retrigger link needs an authenticated browser session — so the P1 security fix had been reviewed by nobody but its author. I ran a separate read-only reviewer against it, briefed to **refute** the claim that the leak is closed rather than to confirm it. **This is an internal review, not Greptile's** — that one is still outstanding.

**Outcome: the cross-session leak is closed on every path it could find. No security defect survived.** What it actually checked, rather than assumed:

- **Every surface an approval or its id crosses is owner-scoped** — event delivery (`session.ts:143`), `approvals.decide` (answers `not-found` to a foreign subject, indistinguishable from a bad id), `approvals.pending`, the wire routes (all use the *authenticated* principal), `byoApprovals.read` on all three branches, `stream()` (rides the parking turn's own writer — no subscription to scope), the MCP door, and `onApprovalDecision` (resumes under the call's own stored ctx). Ids are `apr_` + `randomUUID()`, and guessing one buys nothing given the subject checks.
- **The fixed site is the only instance of the pattern.** `session.ts:157` is the sole production subscriber to `onApprovalRequested`; every other hit is an interface declaration or a test. **`packages/vendo` never subscribes at all.** I re-ran that grep myself to confirm — this was the finding that mattered most, because a second unscoped subscriber would have meant the P1 was still live and merely unreported.
- **Fail-closed is sound, not fail-broken.** `RunContext.sessionId` is *required* in both the TS interface and the zod schema (`run-context.ts:70`, `:113` — verified directly), and the sole emitter stamps it unconditionally from the live ctx, so no live emission can be ownerless. The optional typing exists only for rows persisted before the field, and persisted rows are never re-emitted. Non-chat venues (automations, schedules, box, MCP) mint deliberately non-thread session ids and deliver through their own channels, so nothing legitimate is starved.
- **Subscription lifecycle holds** across the interleavings — stale disposers can't release a newer subscription, double-release no-ops, re-subscribe after full release works. A session dropped with handlers attached still retains its subscription, but that is strictly better than before, when it could never be released at all.
- **Thread is the right boundary** for this event. The user-wide view already exists as principal-scoped `GET /approvals`, and a same-user decide by id still works from anywhere. Subject-only scoping would also have closed the P1 but would broadcast an approval card into every open thread — noisier, no safer.

Suites re-run during review: agents **62/62**, guard **259 passed / 1 pre-existing skip**.

### Two adjacent findings — both pre-existing, neither introduced here, neither blocking

- **`packages/automations/src/engine.ts` is a second writer into the guard's `vendo_approvals` collection** (`engine.ts:63` vs `guard.ts:49`), minting approval rows at `:1113` and `:1541` **without `refs`**, and at the `:1541` site without `ctx.sessionId`. Those rows are invisible to `approvals.pending` and to `#supersedeApprovedSiblings` (both query by refs), yet the guard's TTL sweep does process them. It may well be intentional, since the engine surfaces and decides them by id through its own capture flow — but two writers with different invariants in one collection is fragile, and the concrete consequence is that **a denied engine approval can never void an approved sibling**. Worth an owner's eyes; unreachable from this fix.
- **`on("approval", h)` dedups by function identity** (a `Set`), so registering the same callback twice yields one entry and two disposers, and the first disposer kills delivery while the caller still believes a registration is live. Demonstrated with a scratch test run outside the repo. Identical semantics before the fix, so not a regression — but if `on()` is meant to be called from re-render-prone code with a stable callback, counted registration would be safer.

No tracked file was modified by the review and nothing was pushed.

**Still not merging.** Outstanding: Greptile's own re-review of `61cf3f9b`, and a maintainer's call on the two wall-clock-flaky tests.

yousefh409:
**Live BYO acceptance — current main (1bb535bb6), examples/ai-sdk-agent, real browser (Playwright-driven Chromium), direct `ANTHROPIC_API_KEY` from Infisical, no `VENDO_API_KEY` anywhere.**

## Verdict: the tool door is proven end-to-end. ✅

**1. Guarded read** — "What's the weather in Paris?" → model called `vendo_host_get_weather`, the cautious policy ran it (risk `read`), and the model narrated the real returned data (62°F, cloudy). Plain data, no envelope — as designed.

**2. Approval round trip (the critical proof)** — "Email that trip report to boss@example.com." → model called `vendo_host_send_trip_report` (risk `write`), the guard **parked** it and returned a `vendo/approval-ref@1` envelope, which rendered as a real approval card in the host's own chat: "Needs your approval · Send trip report · Makes changes", with the real tool inputs (to/subject/body) displayed. Clicked **Approve** in the browser → card resolved in place to "Approved — ran" with the executed result: `To: boss@example.com · Delivered: Yes` — i.e. the parked server action actually executed on approval. Screenshots captured at parked and executed states (byo-2-approval-parked.png, byo-3-approval-executed.png in the acceptance session).

**3. Generated app (`vendo_create_app`)** — the door itself works: tool called, `vendo/app-ref@1` envelope returned, inline embed rendered, build streamed over the wire, failure surfaced gracefully with a Try-again card. But the build **failed 4/4 runs**: first from the example's random weather fixture tripping the honesty checks (fixed in #821), then persistently from generation quality on the default BYO model — invalid wire code and hard-coded values the gen checks rightly block. Worse, the chat model then narrated a fabricated success each time. Both filed as #822.

Server log excerpts (guard audit lines, park/approve/execute, and all four build-failure reasons) are preserved in the acceptance session transcript.

PR #798, opened 2026-08-05
feat(apps,vendo,harnesses): reading a file off a box is the seam's business, not each adapter's

yousefh409:
## Why

Reading a file off a sandbox was **adapter-private**. Three implementations
existed with an identical shape, hidden behind
`satisfies SandboxMachine & Record<string, unknown> as SandboxMachine` casts,
and two test doubles had no `files` at all — five private spellings (or
absences) of one contract, sitting on the seam a built app's SOURCE has to
cross on its way out of the box.

Blueprint §3.2 + §12.2, risk §16.4. §16.4's rule is that partial promotion is
two ways to read a file, so all of it lands here in one change with the
conformance suite extended in the same commit.

## What the interface now says

`SandboxMachine.files` — `read`, `write`, `list` — is declared in
`packages/apps/src/sandbox.ts`. The signature is the one all three real
adapters already had; nothing was invented. The documented answers:

- `read` REJECTS for a path the box does not hold — never empty bytes, because
  a silently empty source file is a lost app.
- `read` hands bytes back UNCHANGED: no text decode, no BOM strip, no
  line-ending normalization. Box content is untrusted, and the layer above can
  only verify it against the hash in the app's row if these bytes are those
  bytes.
- `write` creates or replaces the whole file and creates the directories on the
  way to it. It never appends.
- `list` is ONE level and names only, and rejects for a directory the box does
  not hold — same rule as `read`.

## Two real disagreements the promotion exposed

Both were invisible while `files` was private, and both are now pinned by the
shared conformance suite:

1. **The Vendo Cloud list route answers deeper than one level.** With the
   adapter's depth fold reverted, the mock-console conformance leg fails with
   `expected [ 'nested/two.bin', 'one.txt' ] to deeply equal [ 'nested', 'one.txt' ]`.
   The fold is load-bearing, not decoration.
2. **A missing directory rejected on real e2b** (`[not_found] path not found:
   lstat /tmp/vendo-conformance-files/nope`) **while both in-memory fakes
   answered `[]`.** `[]` is how a mistyped source directory reads as "this app
   has no files". The seam now rejects everywhere: the fakes reject (a flat map
   cannot hold an empty directory, so "nothing under this prefix" IS "no such
   directory"), and the Cloud adapter rejects when the console answers no
   entries — loud in the safe direction, with the console's inability to
   distinguish absent from empty recorded at the call site.

## Hostile cases: what is pinned here and what belongs a layer up

Box content is untrusted data — a compromised or merely buggy in-box agent
controls every path and every byte crossing this seam. Pinned in the
conformance leg:

- **Byte fidelity.** The round-trip payload is deliberately hostile: a NUL, a
  CRLF, a bare CR, and three sequences that are not valid UTF-8. Verified
  byte-for-byte through a real e2b box.
- **Entries that look like paths.** `list` returns names only — proven by the
  reverted-fold red above, which is exactly a provider answering a path where a
  name belongs. (A POSIX filename cannot itself contain `/`.)
- **Absent paths and absent directories.** Both reject.

Deliberately NOT pinned here, with reasoning:

- **Paths escaping the app directory (`../`, absolute paths outside it).** This
  seam is the box's whole filesystem, not a partition of it: the bootstrap
  legitimately writes `/app/server.js` and `/tmp/vendo-conformance.pid`.
  Confinement already lives one layer up, at the layer that decides which paths
  may become an app's source — `app-source.ts:97-106` refuses an absolute key
  or any `.`/`..`/empty segment, with the comment "Refused here rather than at
  write time because a `../` key is a checkout escape". Adding a second copy
  down here would be §0's second door for one rule, and would break the
  bootstrap.
- **Bytes that disagree with a caller's metadata.** This seam carries no
  metadata — no hash, no length. The comparison is P0's fail-closed gate at
  `app-source.ts:175,182` ("refusing to check out content that is not the
  content stored"). What this layer owes that gate is bytes it did not touch,
  which is now the pinned byte-fidelity case above.

## §0: what was consolidated or removed

Total concept count goes down.

- Two redundant `files` casts on the real adapters (`files` moved above the
  adapter-private line in both).
- The `files`-shaped half of the Cloud wire test's private-surface cast.
- The `files` cast in three live bootstraps (`e2b.live`, `secrets-egress.live`,
  `sandbox.live`) — they now call `machine.files.write` directly.
- **Three copies of the fakes' in-memory file semantics collapsed into one**
  `inMemoryBoxFiles` (`packages/apps/src/testing/box-files.ts`), shared by
  `fake-sandbox`, `fake-sandbox-stateful` and `fake-box`. The two fakes that had
  no `files` at all now get the same one, so they cannot drift.
- `packages/vendo/src/sandbox.ts`: the public `url()` moved ABOVE the
  adapter-private line, where it always belonged.

**The `satisfies … Record<string, unknown> as SandboxMachine` cast still stands
at `e2b/index.ts` and `vendo/sandbox.ts` — deliberately.** It now carries only
`exec`, which stays adapter-private and out of scope. Removing it would force
`exec` onto the public seam, the opposite of the brief.

`SandboxMachineLike` in `@vendoai/harnesses/claude-code` carries `files` and is
still a structural subset, so the subpath's type surface did not widen.

## §0 receipt: the consolidation caught its own bug, once instead of five times

A review found a real defect in `inMemoryBoxFiles.list`: it treated the root's
prefix as `""` instead of `"/"`, so it sliced nothing off an absolute stored path,
`"/app/a.txt".split("/")[0]` came back `""`, and the empty-name filter dropped it.
**`list("/")` answered `[]` on a box that demonstrably held files** — a root
listing claiming an empty box, which on this seam means "this app's source has no
files". Exactly the silent-wrong-answer class this project keeps closing, and it
was pre-existing: `fake-sandbox.ts` shipped that identical line before this PR.

**It was one fix in one file.** Before the `inMemoryBoxFiles` consolidation in
this same PR, that line lived in every fake that had a `list`, and this finding
would have been five separate fixes in five separate files — with five chances to
fix four of them. The conformance suite now pins the root case for every
implementation at once. That is the measurable payoff of "reuse, never duplicate":
the consolidation is not overhead attached to the feature, it is what made the
next bug cheap.

Checked rather than assumed: `e2b`'s `list` is pure delegation to
`sandbox.files.list(dir).map(entry => entry.name)` with no prefix arithmetic, so
it was structurally immune. The Cloud adapter's depth fold handles the root
correctly but its not-found throw lacked a root exemption — fixed. The mock
console's route built `"//"` for `dir=/` and matched nothing; that is the mock's
own path arithmetic, not console semantics (no filesystem or route treats `"//"`
as root), so fixing it removes an artifact rather than shaping the counterparty
to fit an assertion.

## Verification

**The root-listing fix in commit 2 is WRITTEN BUT NOT YET RUN.** Fleet load
policy has test execution queued behind a shared verification slot, so the
red/green evidence for the new root leg does not exist yet and is not claimed
here. Everything below describes the first commit, which was fully verified.

TDD: the conformance expectations went in first and could not compile against
the un-promoted interface (`error TS2339: Property 'files' does not exist on
type 'SandboxMachine'`, five sites). Three deliberate breaks, each seen red and
restored:

| Break | Red |
|---|---|
| Cloud `list` returns the console's entries unfolded | `expected [ 'nested/two.bin', 'one.txt' ] to deeply equal [ 'nested', 'one.txt' ]` |
| `inMemoryBoxFiles.read` answers empty bytes for a missing path | `promise resolved "Uint8Array[]" instead of rejecting` |
| `inMemoryBoxFiles.list` answers `[]` for a missing directory | `promise resolved "[]" instead of rejecting` |

**Live e2b: green.** The file-seam conformance leg ran against a real E2B
sandbox (`E2B_API_KEY` from Infisical, piped straight in) — 1 passed. That is
what settled the missing-directory question.

**Live Vendo Cloud: could not run, and this is not green.** The console refuses
the prod `VENDO_API_KEY` with `VendoError: Invalid API key.` at `create()`
(`adapter-conformance.ts:249`), before any `files` call, so it says nothing
about the file seam. External credential state, not this change; not chased
further, because hunting a working production credential through the secret
store is the 2026-07-28 exposure pattern. Cloud file-seam coverage therefore
rests on the mock-console lane over the real adapter wire, which the reverted
fold proves is a genuine test.

Tests were run FILE-SCOPED and capped both ways
(`VITEST_MIN_FORKS/MAX_FORKS=1/2`, `VITEST_MIN_THREADS/MAX_THREADS=1/2`,
`--minWorkers=1 --maxWorkers=2`), never `pnpm test`:

- `@vendoai/apps` — 17 files, 249 passed (conformance, fixtures,
  machine-lifecycle, machine-runtime, box-agent, box-turn-routes, e2b,
  egress-failopen, egress-flow, grant-env-reinjection, redaction, schedules,
  secrets-injection, served-apps, served-orgs, access, building-apps)
- `@vendoai/vendo` — `sandbox.test.ts` + `app-source.e2e.test.ts`, 33 passed
- `@vendoai/harnesses` — `materialize.test.ts` + `claude-code.test.ts`, 85 passed
- `@vendoai-fixtures/integration` — `machine-skin.e2e.test.ts`, 1 passed

Package-scoped `build` and `typecheck` green for `@vendoai/apps`,
`@vendoai/vendo`, `@vendoai/harnesses` (exit 0 each). `dependency-guard` OK (15
packages) and `portability-gate` all legs green. **The full suite and full build
are the coordinator's gate at the merge boundary, not run here** — per fleet
policy, four parallel builders running root builds is what spiked this box.

## TEST EDITS

Stated separately, as they are their own change:

- `packages/apps/src/adapter-conformance.ts` — the new conformance leg (this is
  the point of the PR, not incidental).
- `packages/apps/src/e2b/e2b.live.test.ts`,
  `packages/apps/src/e2b/secrets-egress.live.test.ts`,
  `packages/vendo/src/sandbox.live.test.ts` — bootstrap now calls
  `machine.files.write` through the public seam; the private cast shrank to
  `exec` only.
- `packages/vendo/src/sandbox.test.ts` — the private-surface wire test keeps all
  its wire assertions; its title and cast now say `files` is public and only
  `exec` is private.
- `fixtures/integration/src/machine-skin.e2e.test.ts` — its inline
  `SandboxMachine` double gained a `files` member that throws, because that
  journey is the fn/callback boundary and never the box's disk. This is the one
  test file the promotion actually broke at typecheck.

Not touched, deliberately: ~10 other test files hold inline
`const machine: SandboxMachine = {…}` literals with no `files`. Every package's
`tsconfig.json` excludes `src/**/*.test.ts`, so they are never typechecked and
nothing under test calls `.files` on them. Editing ten files for no verification
gain is churn.

---
## Summary by cubic
Promoted sandbox file I/O to the public seam via `SandboxMachine.files` (read/write/list) and added a shared conformance test to keep all providers in sync. This removes duplicate adapter logic, enforces byte-fidelity, and standardizes list semantics.

- **New Features**
  - Declared `files` on `SandboxMachine` in `@vendoai/apps` with clear rules: read rejects missing paths, write replaces whole files, list is one-level names only.
  - Added a conformance leg that verifies byte-for-byte reads (including invalid UTF-8) and list behavior; verified against real e2b.
  - Introduced `inMemoryBoxFiles` for all fakes to share identical semantics; `machine-lifecycle` now exposes `files`; adapters implement `files` and bootstraps write via `machine.files`.

- **Bug Fixes**
  - Cloud adapter now folds deep list responses to one-level names to match the seam.
  - Listing a missing directory now rejects across providers and fakes (no empty array on absent dirs).
  - Listing the root (`"/"`) now returns top-level names instead of empty: fixed in in-memory fakes and the Vendo console mock.

<sup>Written for commit 233aeb0dd018efb3e123f3b2d0cff59b788bb86f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/798?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/sandbox-file-seam.md
+++ b/.changeset/sandbox-file-seam.md
@@ -1,0 +1,47 @@
+---
+"@vendoai/apps": minor
+"@vendoai/vendo": minor
+"@vendoai/harnesses": minor
+---
+
+Reading a file off a sandbox is part of the seam, not each adapter's private
+business.
+
+`SandboxMachine.files` — `read`, `write`, `list` — is now declared on the public
+interface in `@vendoai/apps`. It already existed three times with an identical
+shape, hidden behind `satisfies SandboxMachine & Record<string, unknown> as
+SandboxMachine` casts in the e2b and Vendo Cloud adapters and on the fake, and
+was missing entirely from two other test doubles: five private spellings (or
+absences) of one contract, on the seam a built app's SOURCE has to cross.
+
+The interface now states the answers all of them have to agree on:
+
+- `read` REJECTS for a path the box does not hold — never empty bytes, because a
+  silently empty source file is a lost app.
+- `write` creates or replaces the whole file and creates the directories on the
+  way to it. It never appends.
+- `list` is ONE level and names only: entries directly in `dir`, a subdirectory
+  as its own name, never a path and never recursive. It rejects for a directory
+  the box does not hold, exactly as `read` does.
+- `read` hands bytes back UNCHANGED — no text decode, no BOM strip, no
+  line-ending normalization — because box content is untrusted and the layer
+  above verifies it against the hash in the app's row.
+
+The shared conformance suite (`@vendoai/apps/adapter-conformance`) pins all of
+it in one leg that every adapter runs, so no provider can drift. Verified live
+against a real e2b sandbox, including a payload of NULs, bare CRs and invalid
+UTF-8.
+
+Two disagreements the promotion exposed, both invisible while `files` was
+private: the Vendo Cloud list route answers deeper than one level, so the Cloud
+adapter folds the depth away at the seam; and a missing directory rejected on
+real e2b (`[not_found] lstat …`) while both in-memory fakes answered `[]`,
+which is how a mistyped source directory reads as an app with no files. The
+seam now rejects everywhere.
+
+What went away: two redundant `files` casts on the real adapters, the
+`files`-shaped half of the Cloud wire test's private-surface cast, the
+`files` cast in three live bootstraps, and three copies of the fakes'
+in-memory file semantics (now one `inMemoryBoxFiles`). `SandboxMachineLike` in
+`@vendoai/harnesses/claude-code` carries `files`, still structurally and
+without widening the subpath's imports. `exec` stays adapter-private.
--- a/.changeset/sandbox-file-seam.md
+++ b/.changeset/sandbox-file-seam.md
@@ -32,7 +32,15 @@ it in one leg that every adapter runs, so no provider can drift. Verified live
 against a real e2b sandbox, including a payload of NULs, bare CRs and invalid
 UTF-8.
 
-Two disagreements the promotion exposed, both invisible while `files` was
+The consolidation paid for itself immediately: a review found that the
+in-memory `list` treated the root's prefix as `""` rather than `"/"`, so it
+sliced nothing off an absolute path and dropped every name as blank — `list("/")`
+answered `[]` on a box full of files. Before `inMemoryBoxFiles` that line existed
+in every fake that had a `list` and would have been a separate fix in each. It
+was one fix in one file, and the conformance suite now pins the root case for
+every implementation.
+
+Two further disagreements the promotion exposed, both invisible while `files` was
 private: the Vendo Cloud list route answers deeper than one level, so the Cloud
 adapter folds the depth away at the seam; and a missing directory rejected on
 real e2b (`[not_found] lstat …`) while both in-memory fakes answered `[]`,
--- a/fixtures/integration/src/machine-skin.e2e.test.ts
+++ b/fixtures/integration/src/machine-skin.e2e.test.ts
@@ -110,6 +110,14 @@ function fakeBox(): { sandbox: SandboxAdapter; setEnv(env: Record<string, string
       const result = await handler();
       return { status: 200, headers, body: encoder.encode(JSON.stringify(result)) };
     },
+    // The skin journey is the fn/callback boundary, never the box's disk — the
+    // seam member is here so this double stays a whole SandboxMachine, and it
+    // fails loudly rather than answering a file it does not hold.
+    files: {
+      async read(path) { throw new Error(`the machine-skin box holds no files (${path})`); },
+      async write(path) { throw new Error(`the machine-skin box holds no files (${path})`); },
+      async list(dir) { throw new Error(`the machine-skin box holds no files (${dir})`); },
+    },
     async url(port?: number) { return `https://${port ?? 8080}-fake_box_machine.skin.test`; },
     async snapshot() { return "fake:machine-skin"; },
     async stop() { /* sleep */ },
--- a/packages/apps/src/adapter-conformance.ts
+++ b/packages/apps/src/adapter-conformance.ts
@@ -244,6 +244,47 @@ export const sandboxAdapterConformance = (
       TEST_TIMEOUT_MS,
     );
 
+    it("writes, reads back byte-for-byte, lists one level, and rejects what it does not hold", async () => {
+      const adapter = await harness.makeAdapter();
+      const machine = track(await adapter.create({ env: { PORT: "8080" } }));
+      // /tmp, not the app root: every provider's box has it and it is writable
+      // without bootstrap, so the file seam is provable on a bare machine.
+      const dir = "/tmp/vendo-conformance-files";
+      // Deliberately hostile bytes: a NUL, a CRLF, a lone CR, and three
+      // sequences that are not valid UTF-8. Box content is UNTRUSTED — the
+      // layer above verifies a built app's source against the hash in its row,
+      // which it can only do if this seam hands back the bytes UNCHANGED. No
+      // text decode, no BOM strip, no line-ending normalization.
+      const hostile = new Uint8Array([0, 13, 10, 13, 0xff, 0xfe, 0x80, 65, 13, 0]);
+
+      await machine.files.write(`${dir}/one.txt`, "first");
+      await machine.files.write(`${dir}/nested/two.bin`, hostile);
+
+      const read = await machine.files.read(`${dir}/one.txt`);
+      expect(read).toBeInstanceOf(Uint8Array);
+      expect(decoder.decode(read)).toBe("first");
+      expect(await machine.files.read(`${dir}/nested/two.bin`)).toEqual(hostile);
+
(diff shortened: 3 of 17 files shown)


yousefh409:
Confirmed and fixed in 233aeb0d. The finding is real, and I traced the exact mechanism rather than taking it on faith:

`dir = "/"` gave `prefix = ""`, so a stored `/app/a.txt` went `slice(0)` → `/app/a.txt` → `.split("/")[0]` → `""`, and the `name !== ""` filter dropped it. **`list("/")` returned `[]` on a box that demonstrably held files** — a root listing claiming an empty box, which on this seam means "this app's source has no files". It is pre-existing: `fake-sandbox.ts` shipped that identical line before this PR.

**The fix, at the root cause:** box paths are absolute, so the root's prefix is `"/"`, not `""`. Both spellings collapse onto one `root` boolean on one line, so `list("/")` and `list("")` cannot answer differently by construction — stronger than testing them apart. The not-found throw added a root exemption: the root exists on every box, with files or without.

**Checked, not assumed, for the other implementations:**
- **e2b** — pure delegation (`sandbox.files.list(dir).map(e => e.name)`), no prefix arithmetic. Structurally immune.
- **Cloud adapter** — its depth fold handles the root correctly, but the not-found throw lacked the root exemption. Fixed.
- **Mock console** — its route built `"//"` for `dir=/` and matched nothing. That is the mock's own path arithmetic, not console semantics: no filesystem or HTTP route treats `"//"` as root. Fixed as an artifact removal, deliberately *not* as counterparty behaviour invented to make my assertion pass.

**The test is in the conformance suite**, so it runs against all five implementations rather than one fake. Asserted by containment, since a real box's root also holds `bin`, `etc`, `home`: `list("/")` must contain `tmp`, and no entry may be empty or contain a `/` — that last clause is the bug's exact signature.

**`list` on a genuinely non-existent directory rejects.** Pinned in the previous commit, and settled by probing real e2b rather than by preference: it answers `[not_found] path not found: lstat …: no such file or directory`. Both in-memory fakes had answered `[]`. The root is the sole exemption.

**Verification status, stated plainly:** this fix is written and pushed but **not yet run** — fleet load policy has test execution queued behind a shared verification slot. No red/green evidence for the new root leg exists yet and none is claimed. The first commit was fully verified, including live e2b.

PR #811, opened 2026-08-05
feat: a build can be corrected while it is still building (mid-build steering, §10.2)

yousefh409:
## A build can be corrected while it is still building

Mid-build steering (blueprint §10.2). `session.send()` / `interrupt()` gain their
first steering caller; nothing is built beside them. **No new card, pill, toast,
preview surface, event vocabulary, composer, queue or send path.**

### Queue policy as built — queue-and-deliver-at-next-step

A message typed while a turn is in flight still parks in the composer's **existing
single queued slot**, unchanged. It is then *offered* to the running turn. The
Agent SDK's own already-open input stream does the queueing, so we build no second
queue — that is where "at-next-step" comes from.

| | What the user sees |
|---|---|
| **Steer lands** | Their words become a normal user turn in the transcript; the same `.fl-queued` chip flips to **Sent · added to the reply in progress**, `Send now` disappears (nothing left to send). It is **not** re-sent at turn end. |
| **Steer does not land** | Today's behaviour, untouched: **Queued · sends when the reply finishes**, then flushes on the busy edge. Nothing lost, nothing duplicated. |

`interrupt()` stays the explicit stop. **Steering never cancels.** No new stop UI.

The chip's copy states the message's *fate* only — never anything about how the
result looks. The build's own reply is the only thing entitled to describe the build.

### The §1 harness-contract change — NEEDS RATIFICATION

One **optional** member on `Turn` (`packages/core/src/harness.ts`):

```ts
readonly onSteer?: (handler: (text: string) => Promise<boolean>) => void;
```

- **Inbound**, so deliberately *not* a `HarnessEvent` — that union is closed and
  describes what a harness *says*. A steer is what a harness is *told*; merging
  them would be the second event vocabulary §0 forbids. `HarnessEvent` is untouched.
- **Optional**, so every existing harness and driver is unchanged by construction.
- **One mechanism, not two** — a registered callback, no parallel async iterable.
- The handler answers whether *this* message landed, because a harness that can
  steer in general still cannot when the thing it drives has just finished.
  Not registering **is** the "no" — which is why there is **no capability
  protocol** anywhere in this change.

Second, smaller widening (not §1): `HarnessRuntimeDeps.liveTurn`'s published
object gains `steer`. Reused rather than adding a deps member — "publish the turn
now in flight to this process's own doors" is already what that hook means.

### Three hazards, each measured before it was fixed

**1 · Early settle (fatal, silent).** Every user message the SDK answers produces
its own `result`, and `send()` settled on the next one it saw. A naive steer
therefore resolved the *original* `send()` at message 1's result → the box door
marked the message done → the poll loop returned → **the turn ended on the wire
while the box was still working, and the steer's own output went to a state nobody
polls.** N steers add exactly N results, so N are swallowed. `interrupt()` clears
the count, or a stopped turn would hang to the 15-minute message budget.

Proven by writing the naive version first and watching the test go red:
`expected true to be false` — the caller's promise had settled while the steered
turn was still held open.

**2 · Transcript seq collision (verdict: REAL, and different from the brief's
guess).** `persistTurn` writes one row per message at its **index** in the turn's
own array. A side-channel write to the store cannot know that index, so it lands
at the seq the assistant message then claims. Measured with the runtime's real
persist path: `m_steer` at seq 1, assistant **also** at seq 1. The store reads
`ORDER BY seq` with **no tiebreak**, and `schema.ts` calls seq "the ONLY ordering
authority" — so the user's steer and the reply it caused had **undefined order**.
Not an overwrite: a collision.

Fixed by appending to the in-flight turn's own message list, which the existing
turn-end pass then persists once, in place.

**3 · Contract** — above.

### The seam, both ends, no stub on either side

- `ClaudeSession.steer` → the SDK's open inbox (`claude-turn.ts` still imports
  **nothing**, verified — its output is copied verbatim into a machine image).
- Box door route `/session/:id/steer` — extends the existing regex; only the
  message *in flight* can take one.
- `SessionMachine.steer` in **both** homes: an HTTP hop to the real box door, and
  a direct push into the live local session. A seam with one implementation lies.
- `POST /threads/:id/steer` modelled exactly on the liveness beat:
  principal-scoped, and a **foreign or unknown thread id answers exactly what an
  idle thread answers** — no oracle.
- `packages/vendo/src/server.test.ts` drives the whole chain with the real wire
  route, real registry, real runtime and a real `onSteer` harness, then reads the
  transcript back through the real `GET /threads/:id`.

### `interrupt()` was never unused — correcting the blueprint

The blueprint says `send()`/`interrupt()` "exist unused". `interrupt()` already had
real callers via the abort path, so **no stop affordance was added**:

- **local**: direct call (`local.ts:305`, plus the abort listener).
- **box**: `box.ts` does *not* call the session — it posts
  `/session/:id/interrupt` and the **door** calls `session.interrupt()`.

Now pinned by a test, proven failable by breaking the abort check.

### Real-browser proof

`packages/ui/e2e/steering.spec.ts`, driven by the package's own Playwright against
the in-process wire server. Both specs pass; the preview server was reaped
immediately after.

- `packages/ui/e2e/screenshots/steering-landed-mid-build.png` — the moment of
  steering during a live build: chip reads `SENT · added to the reply in progress`,
  the user's correction is in the transcript, the build is still streaming and
  `Stop` is still offered.
- `packages/ui/e2e/screenshots/steering-fallback-queued.png` — the unchanged path:
  `QUEUED · sends when the reply finishes` with `Send now`.
- `packages/ui/e2e/screenshots/steering-settled-in-history.png` — **needs
  re-capture**: the `[stream-long]` fixture answers with 100 paragraphs so the
  settled viewport sat past the steer. The spec fix (scroll into view) is in this
  commit; the capture needs the fleet browser slot.

### §0 — what went down, not up

No new concepts. The queued slot, the `.fl-queued` element, the box door, the
`SessionMachine` port, the liveness registry module and the `liveTurn` publish hook
were all **extended, not duplicated**. Two things were consolidated rather than
added beside: the steer sink rides the **existing** `liveTurn` publication instead
of a new runtime dep, and the landed/queued states are **one element with two
labels** instead of a second component. `interrupt()`'s "unused" status is removed
from the docs by being disproved.

### Gate

Green post-rebase: package-scoped `build` and `typecheck` for all five touched
packages, `dependency-guard` (16 packages), `portability-gate` (all legs).

**Not yet re-run post-rebase:** the four vitest slices and the browser spec — all
were green *before* `origin/main` advanced twice under this branch. Held because
the fleet allows two test slots and D5 has a merge-blocking failure waiting on
one. Exact commands are in the lane's parked note. **Do not merge until they are
re-run.**

### Test edits, every one declared

All marked `⚠️ TEST EDIT` in place. All are test **doubles/infrastructure** kept
honest with the port they stand in for — no assertion was weakened:

- `box-turn-routes.test.ts`, `claude-code.test.ts`, `local-session.test.ts` — the
  session doubles gained `steer` (and now mirror the real session's refusal when no
  turn is in flight); the fake box records `steers` and `interrupted` so the chain
  is observable.
- `claude-session.test.ts` — the fake SDK gained an optional `gate`, the only way
  to observe that a caller's promise is still *pending* while the session works.
- `runtime.test.ts` — the fixture accepts `liveTurn`.
- `wire-server.ts` — the `/threads/:id/steer` route and a `[steerable]` per-turn
  marker, **off by default** so every suite written against the turn-end flush
  keeps meaning what it meant.
- `steering.test.tsx` releases its held turn in `afterEach`, so one failed
  assertion cannot cascade into a hook timeout.

No existing assertion was changed. `composer.test.tsx`, `conversation-registry.test.tsx`,
`export-surface.test.ts` and `hooks.test.tsx` all still pass unmodified.

---
## Summary by cubic
Let users correct a reply while it’s still streaming. A message typed mid-turn is offered to the build in progress; if it lands, it becomes a normal user turn and is not re‑sent at the end.

- **New Features**
  - Queue policy: queue-and-deliver-at-next-step. The SDK’s open input stream queues the steer; `interrupt()` remains the only cancel.
  - API/UI: `POST /threads/:id/steer` (principal‑scoped) and box route `/session/:id/steer` return `{ landed }`. The composer offers steer while busy, flips the chip to “Sent · added to the reply in progress,” hides “Send now,” and never resends a landed message; attachments/grounding are not steered.
  - Runtime/SDK: optional `Turn.onSteer(handler)`, `ClaudeSession.steer(): boolean`, and `SessionMachine.steer` for both local and box paths. Published via `liveTurn`; `vendo` registers a per‑turn steer sink and dispatches through `steerActiveTurn`.

- **Bug Fixes**
  - Prevented early settlement by counting extra `result` messages; `interrupt()` clears the count so stop ends promptly.
  - Removed transcript seq collisions by appending steered user messages to the in‑flight turn’s message list and persisting once at turn end; the UI inserts the steered message before the assistant reply for stable order.
  - Beats: no `finishing` on a steered result, and narration resets so the rework can re‑narrate its phase.

<sup>Written for commit 87b0508e82dd10d1382f4013180bf5a5d5cf2d9f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/811?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/apps/box/turn-routes.mjs
+++ b/packages/apps/box/turn-routes.mjs
@@ -357,7 +357,7 @@ export const createSessionRoutes = (options = {}) => {
         return { status: 202, body: { messageId: await startMessage(payload) } };
       }
 
-      const match = /^\/session\/([^/]+)\/(poll|interrupt)$/.exec(pathname);
+      const match = /^\/session\/([^/]+)\/(poll|interrupt|steer)$/.exec(pathname);
       if (match === null) return { status: 404, body: { error: `unknown route: ${pathname}` } };
       const state = messages.get(match[1]);
       if (state === undefined) return { status: 404, body: { error: `unknown message: ${match[1]}` } };
@@ -365,6 +365,20 @@ export const createSessionRoutes = (options = {}) => {
       if (match[2] === "poll") {
         const cursor = Number.isInteger(payload?.cursor) ? payload.cursor : 0;
         return { status: 200, body: await poll(state, cursor, payload?.waitMs) };
+      }
+      if (match[2] === "steer") {
+        // The user typed while this message was being answered. Straight into
+        // the live session, which hands it to the model at its next step — the
+        // box queues nothing, because the SDK's own input stream already does.
+        if (typeof payload?.prompt !== "string" || payload.prompt.trim() === "") {
+          return { status: 400, body: { error: "prompt must be a non-empty string" } };
+        }
+        // Only the message IN FLIGHT can take one: a finished message has no turn
+        // to fold the words into, and the host's own queue is the fallback. Said
+        // as an answer rather than an error — "it did not land" is a fact the
+        // host acts on, not a failure.
+        const landed = state === current && session?.steer(payload.prompt) === true;
+        return { status: 200, body: { landed } };
       }
       // The user hit stop. The SESSION survives — only this turn is cut short,
       // which is the whole reason a live session interrupts instead of aborting.
--- a/packages/apps/src/box-turn-routes.test.ts
+++ b/packages/apps/src/box-turn-routes.test.ts
@@ -38,7 +38,22 @@ const scripted = (body: (input: Any, prompt: string) => Promise<void>) => {
   const factory = (input: Any) => {
     opens.push(input);
     return {
-      async send(prompt: string) { await body(input, prompt); },
+      async send(prompt: string) {
+        input.__inFlight = true;
+        try {
+          await body(input, prompt);
+        } finally {
+          input.__inFlight = false;
+        }
+      },
+      // ⚠️ TEST EDIT — the double must speak the whole `ClaudeSession` port, and
+      // `steer` joined it. It records what it was handed and, like the real
+      // session, refuses when no turn is in flight.
+      steer(prompt: string) {
+        if (input.__inFlight !== true) return false;
+        (input.__steers ??= []).push(prompt);
+        return true;
+      },
       async interrupt() { input.__interrupted = true; },
       async end() { input.__ended = true; },
     };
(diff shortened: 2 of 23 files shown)


yousefh409:
### Rebased onto `main@86fcd157` — two conflict sites reconciled once, plus the steer×beats seam

`main` advanced twice under this branch (phase0 → `98eba22a2` after #810's beats/stream-resume, then → `86fcd157` after a `store/` refactor). Rebased with pinned shas each time; the `git diff --stat origin/main..HEAD` gate confirms **exactly the 23 steering files** and nothing else.

**Site 1 — `claude-turn.ts` `result` branch (steer × #810 beats).** These collided at one boundary. Resolution: the `finishing` beat fires **only on the final result** (`else if (steeredResults === 0)`) — never on a steered/intermediate one, or the build would claim "Finishing up" while the correction's rework is still ahead (§3.4, no progress-lie); `narrated.clear()` runs on **every** result including steered ones, so the rework re-narrates its phase (the mockup's scene-3 "Regrouping by client"). New red-first assertion in `claude-session.test.ts` pins both halves (proven red by reverting the guard: `finishing` then fired twice).

**Site 2 — `wire/threads.ts` (steer × #810 stream-resume).** Independent: steer *injects* input via the registry, resume *replays* the recorded byte stream — so a rejoining client sees steered output for free through the same recorded Response (noted inline). Import union of both; `POST /threads` body taken wholesale from main (never touched); both `GET /threads/:id/stream` and `POST /threads/:id/steer` kept.

### Real-user course-change proof (Yousef's standing rule)

The split-view workspace (§10.2's heavy-build surface = the mockup's scene 3): type a correction mid-build → it lands as a user turn **and a fresh `Regrouping by client` beat appears in the Building rail** while the build streams on. The wire fixture cannot re-plan, so it faithfully mirrors the real box's causal chain (steer → a new `building` beat into the open stream) rather than inventing planning — stated honestly in the spec. Proven red-first (reverting the fixture's beat-on-steer fails the assertion). Screenshots: `steering-landed-mid-build.png` (rail beat + SENT chip + still streaming), `steering-settled-in-history.png` (both user turns in order), `steering-fallback-queued.png` (unchanged queue path).

All four slices + the browser spec re-run green on `86fcd157`: apps 63 · harnesses 114 · vendo server 124 · ui 48 · e2e 2/2.

PR #827, opened 2026-08-05
feat(automations)!: per-trigger model, fail-loud consent, one brain, one scheduler

yousefh409:
## What

An app carries `triggers[]` — per-trigger grants, sponsorship, and on/off, instead of one grant blob per app.

- **Park removed.** A permission miss fails loud with **Grant & re-run**; the effect-lineage ledger makes re-runs safe.
- **One brain.** Agentic firings ride `@vendoai/agents` (away entry). The legacy loop no longer serves automations.
- **One scheduler.** `vendo.json` cron folded into doc triggers (−375 lines).
- **Authoring skill + refusal law.** Irreversible asks are refused crisply.
- **Consent honesty.** Declared tools only; anything withheld from unattended runs is never asked for.

## Security

- Adversarial attack suite: 9 properties, all green — including fire-time trigger isolation.
- Redteam: away-authority unweakened.
- Four data-shape migrations: sponsorship, armed, refs (incl. BYO adapters), schedule cursor.
- Interchange import fix — pre-train archives no longer silently lose their automations.

## Proof

- Central full gate green: **build 25/25 · test 58/58 · typecheck 45/45 · lint**. Arc across gates: 16 → 3 → 0 failures.
- Fresh-eyes real-browser walk of every user journey: consent card, fail-loud → grant & re-run, kill switch on-row, agentic firing watched live, authoring + send-refusal.
- Changesets included (automations major; apps major; others minor).

Screenshots in a comment below.

## Known limitations / fast-follow

1. Authoring a **second** automation via chat is blocked by the plan compiler's one-`<Server>`-per-plan rule (`core/src/genui/plan/compile.ts:347-350`). Model/engine/panel fully support multi-trigger (proven via wire + UI on seeded apps); the fix belongs to the UI-gen pipeline rebuild.
2. Agentic re-runs can still repeat a non-destructive write — model-driven call ids defeat positional effect identity. Irreversible tools never run unattended regardless.
3. Two walks are env-gated: console (needs `HOSTED_DATABASE_URL`) and real-box manifest cron (needs `VENDO_BOX_TEMPLATE`).
4. The agent may author aggressive crons (`* * * * *` authored for a daily ask; renders raw — `humanizeCron` wildcard fallthrough). No cadence guardrail exists.
5. An add-ask that avoids all add-vocabulary, plus a lazy planner, can still replace. The mechanical floor covers the vocabulary; the rest is model choice.
6. The legacy agent loop survives solely for `vendo_delegate` (Mastra / AI-SDK packs). Deleting it is a fast-follow that rewires delegation onto the runtime.
7. Pre-existing export drops (`placements` / `source` / `componentTools` in the interchange allow-list) predate this train — documented, not fixed.
8. `demo-bank` / `demo-accounting` have no typecheck scripts; an arity break only surfaced at the next build.
9. Proposal not applied: CSS `clamp` on run-summary rendering for stored/steps rows.

## Verification

- [x] Screenshots attached (see comment)

---
## Summary by cubic
Moves automations to a per-trigger model, replaces park/resume with fail‑loud + “Grant & re-run”, and unifies agent and scheduler paths for simpler, safer automation.

- **New Features**
  - Per-trigger automations: `AppDocument.triggers[]` with required `id`; grants, sponsorship, cursors, webhooks, and runs are keyed by `(appId, triggerId)`.
  - API changes: `enable/disable/dryRun/adopt(appId, triggerId, ctx)`; `list()` returns per-trigger state.
  - Fail‑loud consent: missing permission ends the run with `needs-permission`, writes a capture, and the UI offers “Grant & re-run”.
  - Effect lineage: re-runs inherit `lineageId` so completed effects aren’t repeated; re-runs fire the persisted trigger definition (`__trigger`) that originally fired, so positional call ids stay stable even if the declared trigger changed; audit rows include `trigger.id` and `lineageId`.
  - Agentic firings: run via `@vendoai/agents` `awayRunner` (one non-interactive harness run per firing).
  - One scheduler: machine `vendo.json` schedules are folded into doc triggers and fired by the automations tick.
  - Honest consent: arm-time asks follow declared agentic tools; irreversible tools are refused at authoring; unattended listings show the service dispatcher only when per‑slug grants exist.
  - Store/schema: per-kind trigger refs (`trigger_kind_schedule`, `trigger_kind_host_event`); grants carry `trigger_id`; legacy triggers, armed rows, and cursors auto‑migrate; pre‑list sponsorships migrate on read; imports that used `trigger` normalize to `triggers[]`.
  - UI: one row per trigger with its own toggle and Stop button; clearer run status/summary; schedule labels name UTC; live polling.

- **Migration**
  - Replace `trigger` with `triggers[]` and give each trigger an `id` (use `DEFAULT_TRIGGER_ID`/`"main"` for single-trigger apps).
  - Update calls to `automations.enable/disable/dryRun/adopt` to pass `triggerId`.
  - Remove `AppsRuntime.schedules` and `SCHEDULE_STATE_COLLECTION`; call the single `/api/vendo/tick`.
  - If you use a custom store adapter, add per-kind app refs and persist `trigger_id` on grants; reserved adapters handle this automatically.
  - Machine apps: rely on manifest sync to convert `vendo.json` schedules into triggers; do not run a second scheduler.

<sup>Written for commit b0042cf183f0d88083fa5c78d7c3ac1e3ee24401. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/827?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
## Evidence — real-browser walk

Every journey below was walked in a real browser on the demo hosts (Maple / Cadence); the last two are the console half from the tandem PR.

**One app, two triggers, independent switches — the daily schedule is on, the host-event trigger is off**



---

**After a reload the states hold and stay independent — first trigger now off, second on and waiting on its own permission**



---

**Consent honesty — the card asks for exactly the 2 tools the trigger declared, and nothing else**



---

**Fail loud — a permission miss is a Failed run that names what it needed, not a silent park**



---

**Grant & re-run on the failed row, with the standing-access card naming the exact permission**



---

**After granting: a fresh Succeeded run above the failed one — the effect-lineage ledger makes the re-run safe**



---

**Kill switch on the trigger row — Stop appears on the running trigger while its sibling stays untouched**



---

**Agentic firing on @vendoai/agents, watched live — the run summarises what it found in its own words**



---

**Console runs view, grouped per trigger (tandem PR: runvendo/vendo-web#275)**



---

**Console run detail — a failed run names the tool nobody has allowed yet**

yousefh409:
@greptile-apps review

yousefh409:
## Merged `main` into the train (integration notes)

`main` moved 33 commits while this train was in the shop, and the PR was
`CONFLICTING` — which is also why GitHub had created no CI run for it. Merged
`main` in rather than rebasing, so the train's own history stays reviewable.

**Three textual conflicts, resolved:**

1. `packages/apps/src/runtime.ts` — the train widened `armAutomation` to
   `(appId, triggerId, ctx)`; `main` added `files?: FilesAdapter` next to it.
   Both kept. Every one of the six call sites already passes the trigger id.
2. `packages/guard/src/guard.ts` — a doc comment on `effectBaseKey`. The train
   moved the run component to the firing (`ctx.trigger.lineageId`); `main`
   documented why the new `ctx.turnId` is deliberately unused. Merged text keeps
   the lineage rule and `main`'s turn-id caveat, minus the duplicated reasoning.
3. `packages/apps/src/generation/lanes.test.ts` — both sides touched the same
   test. `main`'s rewrite (`experimentalServedApps` → `servedProxyPath`) wins for
   that test; the train's new machine-free-rungs test and `main`'s new
   served-narrows-box test both survive. All three assert today.

**Three semantic breaks git merged cleanly and neither side could see:**

4. `examples/demo-bank/src/demo-script/console-seed.ts` (new on `main`, #819)
   seeds six app documents with the pre-list singular `trigger:`, which no longer
   typechecks against `AppDocument`. Converted to `triggers: [{ id:
   DEFAULT_TRIGGER_ID, ... }]` — exactly what the read-time normalizer would have
   produced. This is limitation (8) in the description landing in practice:
   demo-bank has no typecheck script, so it surfaced only at `next build`.
5. `fixtures/automations-e2e/src/manifest-foldin.e2e.test.ts` — `main` made
   `files` a required member of `SandboxMachine`, and this fixture's box double
   predates it. Given the same loud-rejection stub the sibling fixture
   (`fixtures/integration/src/machine-skin.e2e.test.ts`) already uses; the
   fold-in journey never touches the box's disk. Fixtures typecheck their tests,
   which is why this one caught it and the packages did not.
6. `packages/vendo/src/app-rebuild.e2e.test.ts` (new on `main`) seeds a pre-list
   row and asserted `doc.trigger` survives commit after commit. Under this train
   the document door normalizes that stored row into `triggers[{ id: "main", … }]`,
   so the singular field reads `undefined` by design. The property the test guards
   — a repeated source commit must not gratuitously drop the trigger — is
   untouched and still asserted; only the spelling moved to `triggers`. The seed
   deliberately still writes the OLD shape, so the test now also holds the
   migration door across those same commits. Called out because it is a test
   change: it encodes the contract this train replaces on purpose.

**Gate re-run on the merge result:** `pnpm build` 27/27 · `pnpm typecheck` 48/48 ·
`pnpm lint` · `pnpm test` — full suite, not a subset, because a merge this size
is exactly where a green subset lies.

yousefh409:
Verified and fixed in the push above — thanks, this one was real, and it was aimed at exactly the property this train exists to provide.

**Reproduced first, independently of the artifacts.** New case in `packages/automations/src/engine.test.ts`: step 0 completes, step 1 fails on a missing permission, the author inserts a step *ahead* of step 0, then the run is re-run. Before the fix the completed call came back renumbered (`_0_read` → `_1_read`), so its receipt was never consulted — red for precisely the reason you gave.

**Root cause is the assumption, not the identity scheme.** `runSteps` already documents why the call id is positional rather than by `step.id`: step ids are not validated unique, and two steps sharing one would share a call id and *skip* a mutation — a worse failure than repeating one. Its justification ends "…just as stable across a re-run, since the re-run reads the same step list." That was the bug: `runs.rerun` resolved `triggerOf(app.row.doc, run.triggerId)` — the definition as it stands *now* — while keeping the original lineage. The step list the comment relies on was never the one the re-run read.

**Fix: re-run the trigger that fired.** A run persists `__trigger` (the validated definition it actually fired) exactly the way it already persists `__event`, and `runs.rerun` fires that. The declared trigger is still resolved and still gates the re-run — it is what proves the trigger exists and is armed — but it is no longer what gets fired. Runs written before this fall back to the declared trigger, i.e. today's behavior. This makes the existing comment's assumption true rather than adding a second identity scheme.

Semantics worth stating outright: a re-run now re-fires the definition that failed, so an edit made after the failure is not picked up by that re-run. That is the coherent reading of a lineage — same firing, same event, same steps — and the remedy for an edited automation is its next firing or a dry run, neither of which shares the failed run's ledger. It also leaves the **Grant & re-run** journey untouched, since granting a permission does not edit steps.

Gates on the fix: `@vendoai/automations` 106/106 (48 in `engine.test.ts`, every existing re-run case included), then full `pnpm build` 27/27 · `pnpm typecheck` 48/48 · `pnpm lint` · `pnpm test` across the monorepo.

One note on scope, since it is adjacent: limitation (2) in the description — an *agentic* re-run can still repeat a non-destructive write because model-driven call ids defeat positional identity — is a different hole and is unchanged by this. Irreversible tools never run unattended regardless.

---

Separately, CI's `conformance` job caught one more thing on the merge commit, fixed in the same push: `packages/ui/e2e/wave3-consumer-voice.spec.ts` stopped an automation by patching the app-level `enabled`/`stopped`, which this train moved onto each trigger (sponsorship is per `(app, trigger)`). The panel still renders **Stopped** and the consumer sentence exactly as before — the spec now stops the trigger. Called out because it is a test change. Worth noting for the repo generally: that suite runs only in the `conformance` job via `test:browser`, which `pnpm test` does not include, so a local full gate cannot see it.

yousefh409:
@greptile-apps review

PR #833, opened 2026-08-06
chore: scoped local gate (test:affected); CI owns the full suite

yousefh409:
## Summary

Task 5 of the lighter-testing plan (Worker B slice): local dev no longer runs
the full test suite. CI is the gate of record.

- Root `package.json`: new `test:affected` script —
  `turbo run test test:ui --affected` (matches `pnpm test`'s task selection,
  including UI browser smoke) with the same VITEST_MIN/MAX_FORKS/THREADS caps
  as `test`, plus `TURBO_SCM_BASE=origin/main` so the comparison is always
  "differs from main as landed" and can't drift with a stale local `main`
  branch ref (see Resolved decisions below).
- `CLAUDE.md`: Commands section gains `pnpm test:affected`; Rules section's
  pre-PR gate now reads `pnpm build && pnpm test:affected && pnpm typecheck && pnpm lint`
  with "never run the full suite locally"; Tests section's `pnpm test`
  paragraph now describes CI's sharded ownership and points local runs at
  `test:affected`. Seam-testing, sibling-distDir, poll-budget, and
  VITEST_MIN/MAX lessons are untouched (verbatim). Checked for stale
  `demo-template`/four-suite-extras references (both were plan artifacts that
  never survived PR #831) — none present, no fix needed.
- `packages/ui/e2e/README.md`: tier table and "NOT in CI" table now match
  what actually exists on this branch after PR #831 (the deletion sweep)
  removed 14 of the 15 specs the plan named for the CI browser gate, plus
  both probationary specs entirely. The gate is unchanged at smoke + 13 files
  (14 total, per PR #834's report — nothing was left to add); "NOT in CI" is
  now exactly keyboard/screenshots/eng-222/live-voice (+ mcp-shim, which runs
  in CI under its own job).

**Note on scope drift:** the plan also had me delete `scripts/serial-gate.sh`
and remove its `.gitignore` entries. PR #831 had already done both before I
rebased onto it — same end state, arrived at by a different PR, zero
`serial-gate`/`gate-failures` references remain either way (grep-confirmed).

## Resolved decisions

**`TURBO_SCM_BASE`.** `turbo run test --affected` doesn't scope correctly by
default, and the originally-pre-approved fallback (`TURBO_SCM_BASE=main`)
didn't fix it — it resolves to whatever the *local* `main` branch ref points
to, which can be stale/diverged from `origin/main` (shared across worktrees,
never fast-forwarded), silently marking nearly every package affected instead
of erroring. Approved fix: `TURBO_SCM_BASE=origin/main` — matches the spec's
intent ("differs from main as landed") and can't go stale the way a local
branch ref can. Re-verified on a clean scratch branch off `origin/main` with
a single-line comment edit in `packages/kit/src/index.ts`: scope resolved to
exactly `@vendoai/box-template`, `@vendoai/kit` for the `test` task (core/ui
only rebuilt as upstream deps, correctly not re-tested).

**`test:ui` coverage (greptile P1, real).** `test:affected` only ran the
`test` task, omitting the browser-smoke `test:ui` task that full `pnpm test`
runs — a regression that only broke browser smoke in the affected package
could pass the local gate. Fixed: now runs `turbo run test test:ui --affected`.
Verified `test:ui` is defined only in `packages/ui/package.json`; turbo
silently skips it everywhere else (confirmed via a live run, not just
`--dry`), so this doesn't inflate scope for unrelated changes. Verified
against an actual `@vendoai/ui` edit: correctly fans out to ui's real
dependents (`demo-bank`, `kit`, `vendo` all declare the dependency) — wide,
but genuine, not a scoping bug.

**greptile P1 on single-branch clones (dismissed, not fixed).** See PR
comment — `TURBO_SCM_BASE=origin/main` is the deliberate choice already
covered above; a `--single-branch` clone lacking `origin/main` isn't this
repo's supported setup, and the alternative (a stale local `main`) is a more
common real failure mode here.

## Test plan

- [x] `pnpm install && pnpm build` clean
- [x] `pnpm lint` green
- [x] `pnpm typecheck` green
- [x] Zero remaining `serial-gate`/`gate-failures` references (grep)
- [x] `test:affected` sanity-checked twice: base scoping (kit → 4 packages)
  and task selection (`test:ui` correctly included/skipped per package)
- [x] greptile-apps findings triaged: 2 real and fixed, 1 dismissed with reason

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -18,7 +18,7 @@ generated UI in a sandboxed, brand-native surface.
 
 ## Commands
 
-- `pnpm install` · `pnpm build` · `pnpm test` · `pnpm typecheck` · `pnpm lint` (turbo-cached)
+- `pnpm install` · `pnpm build` · `pnpm test` · `pnpm test:affected` (scoped to changed packages — the local default) · `pnpm typecheck` · `pnpm lint` (turbo-cached)
 - Demos: `pnpm --filter demo-bank dev` (Maple) · `pnpm --filter demo-accounting dev` (Cadence)
 
 ## Vendo Cloud
@@ -50,8 +50,9 @@ generated UI in a sandboxed, brand-native surface.
 - Never commit to `main`; branch and open a PR.
 - UI-affecting changes are verified in a real browser with screenshots in the
   PR. Tests and typecheck alone don't count.
-- `pnpm build && pnpm test && pnpm typecheck && pnpm lint` must be green
-  before any PR.
+- Local gate = `pnpm build && pnpm test:affected && pnpm typecheck && pnpm lint`
+  on the touched scope. The FULL suite runs only in CI — the PR's green `ci`
+  check is the gate of record; never run the full suite locally.
 
 ## Tests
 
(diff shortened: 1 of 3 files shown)


yousefh409:
Verdicts on the three findings above (verified against code, not taken on faith):

1. **P1 — local gate omits UI browser smoke (`test:ui`).** Real. `pnpm test` runs `turbo run test test:ui`; `test:affected` only ran `test`, so a browser-smoke regression in the affected package could pass locally. Fixed: `test:affected` now runs `turbo run test test:ui --affected`, matching `pnpm test`'s task selection. Verified `test:ui` is only defined in `packages/ui/package.json` — turbo silently skips it for every other package (confirmed live, not just via `--dry`), so this doesn't inflate scope for unrelated changes.

2. **P2 — CI browser-gate count/hermetic-shard language stale.** Real, and already in flight independently: PR #831 (deletion sweep) removed 14 of the 15 specs the original plan named for the gate, and PR #834 confirms nothing was left to add — the gate is unchanged at smoke + 13 files (14 total). Fixed `packages/ui/e2e/README.md` line 8 to say exactly that, and dropped the `ci-cache-shards`/hermetic-shard language along with the two now-deleted probation rows (`cdn-packages.spec.ts`, `appframe-devserver.spec.ts` no longer exist as files).

3. **P1 — `TURBO_SCM_BASE=origin/main` fails in a `--single-branch` clone without that ref.** Dismissing, not fixing. This is the deliberately-chosen design, not an oversight: the alternative (leaving the default, or pinning to a local `main`) was tested and found to silently mark ~all packages affected whenever the local `main` ref is stale relative to `origin/main` — a much more common failure mode in this repo's actual workflow (shared local branches across worktrees) than an intentionally-restricted `--single-branch --branch <feature>` clone, which isn't this repo's documented or supported setup (a standard `git clone` always has `origin/main`). Worker A's companion PR (#834) makes the same choice in CI for the same reason. Not adding fallback/detection logic for the single-branch case — that's complexity nobody asked for, for a clone flag this repo doesn't support.

New commit: e9b2fc393.

PR #835, opened 2026-08-06
feat(store): transcripts + harness state ride StoreOps — hosted stores serve harness turns (S1)

yousefh409:
## What this changes

The three store helpers a harness turn needs demanded a raw SQL handle, so a
key-only (hosted-store) deployment could not serve harness turns and silently
fell back to the legacy chat path. This is **S1** of the one-path project: the
transcript and harness-state helpers gain hosted support. Workspace is S2.

- **`VendoStore` gains `ops?: StoreOps`.** `hostedStore` already exposed `.ops`,
  so it satisfies the member structurally — no change there.
- **One internal selector, `backendOf(store, what)`** (`packages/store/src/helpers/backend.ts`):
  the SQL handle when there is one (same database, one hop shorter), the store's
  own 32-op surface when there is not, and a named `VendoError("not-implemented")`
  naming *"a SQL handle or a StoreOps-capable store"* only when the store offers
  neither. Nothing above the store package can tell the two apart — **no caller
  changed**.
- **Transcripts over the wire as-is (D2):** `transcripts.putMessage` for the
  write, `transcripts.getThread` for the read. Ownership is enforced against the
  thread record's `subject` exactly as the SQL join enforces it against
  `vendo_threads`: a foreign or absent thread reads as `[]` and refuses writes.
  A guarded (`expectedRevision`) edit has no wire expression, so it is refused
  loudly rather than downgraded to last-write-wins — no runtime caller asks for
  one.
- **Harness state** rides the wire's `harness` family under the SAME slot the SQL
  half uses (`harness_state:<threadId>`, keyed by the thread's owner), so §1.3 —
  one slot per thread, a foreign harness *destroying* rather than shadowing it,
  the slot dying with its thread — holds identically on both backends.
- **`harness-turn.ts`** refusal reworded: thrown only when the store has neither
  a SQL handle nor ops, and it names both options.
- **`server.ts`** `storeServesHarnessTurns` now answers true for an ops-carrying
  store (still a no-I/O probe: a WeakMap lookup plus a property read).

## Proof

New tests only; no existing test was touched.

- `packages/store/src/thread-messages.ops.test.ts` and
  `packages/store/src/harness-state.ops.test.ts` — **one behavioral suite per
  helper, run against three backends**: `sql` (real PGlite/Postgres),
  `memory-ops` (core's `memoryStoreOps` reference), and `local-ops`
  (`createStoreOps` — the 32-op contract served off the same real database).
  48 tests.
- `packages/vendo/src/hosted-transcript.live.test.ts` — **the seam with no stub
  on either side**: writes through the real `threadMessageStore` over a real
  `hostedStore` against the **real console** on Yousef's Vendo Cloud account,
  then reads back through a *second, freshly constructed* client that has never
  seen the write. Same for harness state. Gated on `VENDO_API_KEY` having
  content, so CI and a keyless clone stay green.

**Prove-red done:** disabling the ops leg in `backendOf` turns **31 of the 48**
new store tests red (the 17 survivors are the `sql` cases and the two
neither-backend refusals). Restored, all 48 pass.

## Gate

| check | result |
| --- | --- |
| `pnpm --filter @vendoai/store test` | 481 passed, **2 pre-existing environment failures** — `durability.drill` / `split-brain.drill`, which resolve a fixture via `new URL(...).pathname` and break on a worktree path containing a space (`Cool%20Code`). Unrelated to this diff; the store vitest config already calls this drill out as space-path-sensitive. |
| `pnpm --filter @vendoai/store typecheck` | green |
| `pnpm --filter @vendoai/vendo typecheck` | green |
| affected vendo suites (`harness-wire`, `turn-id.e2e`, `org-policy`, `hosted-store`) | 87 passed |
| `hosted-transcript.live.test.ts` (real console) | 4 passed |
| `scripts/dependency-guard.mjs` | green |

Full suite runs centrally at the merge boundary, not here.

## Two things the reviewer must decide on

**1. The console's `putMessage` appends — it does not edit by id.** Measured
live: re-writing an already persisted message under its own id is refused with
`thread … carries two messages with the id "m_2"; message ids must be unique
within a thread`. The console implements `putMessage` as read-thread → push →
put-whole-thread, so it trips the duplicate-id guard.

That is the **approval flip**: when an approval resolves, the runtime re-writes
the already persisted assistant message under its own id (`persistTurn` diffs
against the pristine copy and re-upserts what changed). Until the console makes
`putMessage` replace-by-id, a hosted deployment fails that write. It fails
loudly rather than duplicating history, and there is no wire op that expresses
an edit any other way — **the fix is console-side.** Structure §2's D2 line
("putMessage for insert AND edit-by-id") does not hold against today's console.

Recorded as an executable fact: an `it.fails` in the live test, which passes
while the console is wrong and turns **red the moment it is fixed**. Core's
`memoryStoreOps` shares the gap (its `putMessage` appends unconditionally, and
the wire conformance suite only asserts append), which is why the flip case in
the store suite runs against `sql` + `local-ops` — the two backends that do
carry id semantics. Conformance files belong to S2/S3, so I did not touch them;
they should gain an edit-by-id case with the console fix.

**2. Interim window with S2.** With the probe now accepting ops-capable stores,
a hosted deployment is routed to the harness path — but `workspaceStore` (S2's
file) still resolves `dbFor` and throws, so `sqlDoors()` still fails for hosted
until S2 lands. S1 and S2 must land together; S1 alone on `main` would move
hosted chat from "legacy path, works" to "harness path, errors".

Also noted: `seq` is not sent over the wire (`transcripts.putMessage` carries
none — the service appends at `max(seq)+1`). `persistTurn` walks the canonical
array in index order and writes only what changed, so appended order *is* seq
order. Documented at the branch.

---
## Summary by cubic
Hosted stores can now serve harness turns by moving transcripts and harness state to the `StoreOps` surface. This removes the SQL-only requirement and keeps behavior consistent across backends.

- **New Features**
  - `VendoStore` adds `ops?: StoreOps`; `backendOf` selects SQL or `ops`, and throws only if neither is available.
  - Transcripts use wire ops (`transcripts.putMessage` / `transcripts.getThread`) with ownership checks; guarded edits (`expectedRevision`) are rejected over the wire.
  - Harness state rides the same slot (`harness_state:<threadId>`, keyed by thread owner) with identical clearing and cascade rules on both backends.
  - `server.ts` probe now accepts ops-capable stores; `harness-turn.ts` error mentions both SQL and `StoreOps` options.

- **Migration**
  - Hosted deployments will route to the harness path; `workspaceStore` (S2) must land with S1 to avoid hosted errors.

<sup>Written for commit 384527c76a925b740ef40268fa463795fa50ea10. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/835?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/transcripts-and-harness-state-over-storeops.md
+++ b/.changeset/transcripts-and-harness-state-over-storeops.md
@@ -1,0 +1,46 @@
+---
+"@vendoai/store": minor
+"@vendoai/vendo": minor
+---
+
+Transcripts and harness state ride StoreOps, so a hosted store can serve a
+harness turn.
+
+`threadMessageStore` and `harnessStateStore` opened with `dbFor(store)` and threw
+"Unknown VendoStore handle" for anything `@vendoai/store` did not mint — which is
+every key-only deployment. So `storeServesHarnessTurns` answered false for them
+and the host silently fell back to the legacy chat path: hosted deployments could
+not use `harness:` at all.
+
+- `VendoStore` gains an optional `ops?: StoreOps`. The Cloud `hostedStore` already
+  exposed one, so it satisfies the member with no change.
+- One internal selector, `backendOf`, decides for every store-shaped helper: the
+  SQL handle when there is one (same database, one hop shorter), the store's own
+  32-op surface when there is not, and a named `not-implemented` refusal only when
+  the store offers neither. Nothing above the store package can tell the two
+  apart — no caller changed.
+- Transcripts ride the wire as-is: `transcripts.putMessage` for the write,
+  `transcripts.getThread` for the read, ownership enforced against the thread
+  record's subject exactly as the SQL join enforces it against `vendo_threads`.
+  A foreign or absent thread reads as empty and refuses writes, as it does
+  locally. A guarded (`expectedRevision`) edit has no wire expression and is
+  refused loudly rather than downgraded to last-write-wins; no runtime caller
+  asks for one.
+- Harness state rides the wire's `harness` family under the SAME slot the SQL
+  half uses (`harness_state:<threadId>`, keyed by the thread's owner), so §1.3's
+  rules — one slot per thread, a foreign harness destroying rather than shadowing
+  it, the slot dying with its thread — hold on both backends.
+
+The harness-turn refusal now names both options instead of only SQL, and the
+route probe accepts an ops-capable store.
+
+Proven where it counts: one behavioral suite for each helper runs against three
+backends (real Postgres/PGlite, core's `memoryStoreOps`, and the local 32-op
+backend), and a live seam test writes through the real helper over a real
+`hostedStore` against the real console and reads it back on a second,
+freshly-constructed client — no stub on either side.
+
+Known gap, recorded as a live `it.fails` rather than a comment: the console's
+`transcripts.putMessage` appends instead of editing by id, so re-writing an
+already persisted message (the approval flip) is refused there. The fix is
+console-side; the local backends already do the right thing.
--- a/packages/store/src/harness-state.ops.test.ts
+++ b/packages/store/src/harness-state.ops.test.ts
@@ -1,0 +1,161 @@
+/**
+ * Durable harness state (build contract §1.3), proven against every backend it
+ * can sit on — the transcript suite's sibling, same shape and same reason.
+ *
+ * `harnessStateStore` demanded a SQL handle, which is why a key-only deployment
+ * re-seeded a session-owning harness on every turn. It picks a backend now
+ * (`backendOf`); this file is the proof that §1.3's rules — one slot per thread,
+ * keyed by its OWNER, a foreign harness DESTROYING rather than shadowing it, and
+ * the slot dying with its thread — read identically whichever backend answers.
+ */
+import { VendoError, type Principal, type StoreOps } from "@vendoai/core";
+import { memoryStoreOps } from "@vendoai/core/conformance";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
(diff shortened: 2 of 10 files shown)


yousefh409:
Review sweep — triaging the one finding on this PR ("Workspace storage rejects StoreOps-only harnesses", `packages/store/src/workspace.ts:78`).

**False against the tip branch this lands on (#858, `one-path/s3-surgical-undo`).** It was true of S1 in isolation: at `384527c` `workspaceStore` still opened with `dbFor(store)`, so a StoreOps-only store passed `storeServesHarnessTurns` and then died in workspace init. S2 closed exactly that hole — `workspace.ts` now branches on `backendOf(store, "the workspace")` and hands a non-SQL store `workspaceOpsRows(backend.ops)`, the same façade over the 32-op contract. Verified on the tip; `packages/vendo/src/hosted-workspace.live.test.ts` covers the composed path.

This is an artefact of reviewing an inner slice of a stack against `main`. This PR is closing as landed via #858; the finding needed no fix beyond the S2 commit it already has.

yousefh409:
Landed via #858 (squashed as b576ab9f8). S1 is the first of the three commits in that merge unit; its one bot finding was triaged in the comment above (false against the tip — S2's `backendOf` branch closes it).

PR #848, opened 2026-08-06
refactor(vendo,core)!: packs die; capability arrives as tools + skills

yousefh409:
## Why

A pack was a labelled bundle of four lists — `tools`, `skills`, `checks`,
`components` — merged at boot. Every one of those four already had a home of
its own:

| pack slot | where it actually lands |
| --- | --- |
| `tools` | the one tool registry |
| `skills` | the `/host/skills` workspace mount |
| `checks` | the checking floor (`apps.checks`) |
| `components` | the catalog (`catalog` / `<VendoRoot components>`) |

So the noun bought a `Pack` type, a `PackProvider` function shape, a
`definePack` handle, a `PackContext`, a client-side second import of every pack
module, a `DEFAULT_PACKS` list, a "you left apps() out" warning, and a `packs:`
config key — and zero capability. It was a fifth concept sitting on top of four
that already worked. A developer integrating Vendo should never have to learn
it: they already know what a tool is and what a skill is.

Packs shipped six days ago (#189c813) and were on the delete list before anyone
outside built one.

## What changed

**Capability arrives on the keys a host already writes.**

```ts
createVendo({
  tools: [...hostDeclarations, checkReportTool],  // declarations AND executables
  skills: [buildingComplianceReports],            // NEW — mounted at /host/skills
  apps: { checks: [noUnmaskedAccounts] },         // unchanged
  catalog: { RetentionBadge },                    // unchanged
});
```

**The shape chosen for widening `tools`.** One key, a union, discriminated by
`execute`:

```ts
tools?: readonly (ExtractedTool | ToolDefinition)[];
```

Considered and rejected: a sibling `agentTools?` key. The union is the smaller
surface — a dev learns one word, `tools`, for "the things the agent can do",
and the two shapes were never ambiguous (`ExtractedTool` carries a `binding`
the registry executes through; `ToolDefinition` carries the function itself).
`selectHostTools` filters to declarations for `.vendo` semantics; the executable
half becomes a composition contribution. `tools:`-wins-over-`profile.tools`
precedence is byte-for-byte unchanged.

**The collision check survives, as plain code.** `mergeCapability` in
`packages/vendo/src/capability/merge.ts` is the boot merge minus the pack noun:
two contributors claiming one tool or skill name still throws at boot naming
both, one contributor declaring a name twice still says so in its own words,
the skill-name grammar and the `/host` companion-path projection check still
run, and the registry-marker forgery defence (`from-registry.ts`) is untouched.
`hostToolCollision` (was `hostPackToolCollision`) still refuses a contributor
that claims one of the deployment's own extracted tool names, still with no
I/O, still across all three `profileDir` spellings. All of their tests moved
with them.

The one thing the merge no longer does is claim CHECK and COMPONENT names —
there is no longer a second contributor of either, and host checks/catalog
entries were never validated through the pack merge, so nothing a host does
changes behaviour.

**apps and automations mount themselves.** `packs/apps.ts` and
`packs/automations.ts` are gone; app generation contributes its tools and its
`building-apps` skill directly in composition, and automations contributes THE
LAW's judgment rule (which moved to `@vendoai/automations` as
`unattendedIrreversibilityCheck`, since it is that block's law). Two new
switches:

- `apps: false` — `vendo_make` and the `vendo_apps_*` tools are absent from the
  registry, `building-apps` is absent from the mount, and `/apps/**` answers
  not-found. Honest absence: the agent cannot build apps and says so, rather
  than being handed tools that refuse. The host's own `vendo.apps` handle stays
  — unmounting is about what the agent and the wire offer, not about deleting
  your server code's API.
- `automations: false` — `/automations/**`, `/runs/**` and `/webhooks/**`
  answer not-found, `vendo.emit` refuses loudly, nothing fires locally, and the
  unattended-irreversibility rule leaves the reviewer's rubric.

Both mount by default.

## Equivalence

`packages/vendo/src/default-composition.test.ts` composes a bare
`createVendo()` and asserts the tool-name set and the skill-name set against
**literals** transcribed from the composition before the refactor. Proven able
to fail: dropping one app tool and the `building-apps` skill turned both tests
red, restoring them turned both green.

## One real bug the seam test caught

Sharing one key between two shapes has an obvious trap, and the rewritten
outside-authored fixture walked straight into it: a `tools:` carrying only
executables filtered to an EMPTY declaration list, which counted as "the host
set the slot" — silently erasing every `.vendo/tools.json` declaration AND
blinding the boot collision gate, which then read no file and passed
everything. Fixed in `selectHostTools` (708f69b), pinned by the three
host-tool-collision cases in the fixture and by a focused unit case in
`config-consolidation.test.ts`, proven red by reverting the one-line guard.

## Deviations from the plan

- `automations?: false`, not `false | {…}`. There are no existing host-facing
  automations options to carry, and an empty object type is a meaningless
  second spelling. Widening later is not a breaking change.
- Changeset marks `@vendoai/vendo` and `@vendoai/core` **major**, not
  minor/patch — removing `packs`, `definePack`, `Pack` and `PackProvider` is a
  removal, and #827 marked its own removal major on the same convention.
- The `unattended-irreversibility` judgment rule now joins the reviewer's rubric
  BY DEFAULT. It previously required an explicit `packs: [apps(), automations()]`
  (`DEFAULT_PACKS` was `[apps()]`), so in practice nobody had it. It rides the
  subsystem now, and the subsystem mounts by default. Tools and skills are
  unchanged, which is what the equivalence test pins.
- `packs/components.ts` and the `<VendoRoot packs>` prop are deleted alongside
  the listed files — both existed only to read a pack's `components` slot.
  Components reach the client through `components` exactly as before.
- `docs/archive/config-migration-table.md` is edited despite being archived:
  `handler-options.docs.test.ts` enforces key parity against it.

## Gate

`pnpm build && pnpm test:affected && pnpm typecheck && pnpm lint`, run twice,
green except for pre-existing failures unrelated to this branch — all three
proven so:

- Locally, `@ai-sdk/anthropic` module resolution in `dev-creds/model.test.ts`,
  and `%20`-escaped file URLs in the core packaging and store durability
  drills (a space in this worktree's path). Both reproduce on the untouched
  tree, and CI — which runs on a clean path — passes every one of them.
- In CI, `@vendoai/apps#test:coverage` fails an 88% line ratchet at 85.82%.
  That ratchet has been stale and riding a turbo cache hit on main; this PR
  changes `@vendoai/core` upstream of apps, so the task actually executed for
  the first time in a while. The apps diff here is a type-only rename, and
  reverting apps to the base commit measures the identical 85.82%. Full
  evidence in a PR comment below. Deliberately not "fixed" by lowering the
  threshold — that call belongs in its own change.


yousefh409:
### ⚠️ The one red check is a stale ratchet this PR un-cached, not a regression

`test-rest (group-b)` fails on:

```
ERROR: Coverage for lines (85.82%) does not meet global threshold (88%)
@vendoai/apps#test:coverage: … exited (1)
```

**This branch did not cause it**, and here is the evidence rather than the claim:

1. The entire `packages/apps` diff on this branch is a type-only rename. There
   is not one runtime line in it:

   ```
   $ git diff e1d7686ec -- packages/apps | grep '^[+-]' | grep -v '^[+-] *\(\*\|//\)'
   -import type { PackSkill } from "@vendoai/core";
   +import type { Skill } from "@vendoai/core";
   -export const buildingAppsSkill: PackSkill = {
   +export const buildingAppsSkill: Skill = {
   ```

2. Running the gate on this tree with `packages/apps` reverted to the base
   commit produces the **identical** number — `85.82%`.

3. `git diff e1d7686ec origin/main -- packages/apps` is empty, so main's own
   apps code measures 85.82% as well.

4. Main's CI has never actually run this gate. On the last green main run
   (31071290063, job 92519595374):

   ```
   ##[group]@vendoai/apps:test:coverage
   cache hit, replaying logs 644f7f0a2fce59a0
   ```

   This PR changes `@vendoai/core`, an upstream dependency of `@vendoai/apps`,
   so the turbo hash changed and the task genuinely executed here
   (job 92519458444):

   ```
   ##[group]@vendoai/apps:test:coverage
   cache miss, executing 1209a592e05ce58a
   ```

The 88% ratchet in `packages/apps/vitest.config.ts` has been stale for a while
and riding a cache hit. The first PR to touch anything upstream of
`@vendoai/apps` was always going to surface it.

I deliberately did **not** lower the threshold — that is weakening an
acceptance gate, and it is not this PR's call to make. Raising apps line
coverage by 2.2 points is real, unrelated work. Biggest gaps if someone wants
to close it: `generation/sections.ts` (55%), `generation/validation/islands.ts`
(82%), `generation/validation/validate.ts` (87.5%).

Needs a decision before this can go green: write the missing apps tests, or
re-ratchet the floor deliberately in its own change.

<sub>Every other check passes, including `integration`, `typecheck-lint`, and all
`test-shards` for vendo/store/ui.</sub>

yousefh409:
### Resolved — the ratchet is met by new tests, not by moving the ratchet

`packages/apps` line coverage: **85.82% → 88.54%** against the unchanged 88%
floor. `test-rest (group-b)` now passes.

Four files that had no tests at all, each picked because it is real behaviour a
silent change would break — not because it was cheap to cover:

| file | before | after | what is now pinned |
| --- | --- | --- | --- |
| `src/sandbox-ladder.ts` | 0% | 100% | THE adapter precedence ladder: explicit adapter wins, E2B beats a Vendo key, a whitespace key is not a key (the trim `vendo doctor` agrees on), and a half-installed BYO sandbox refuses at boot instead of silently riding Cloud (0.4.4 defect C) |
| `src/app-source.ts` | 6.99% | 100% | checkout→commit as a **round trip** through one workspace: path confinement (`../` cannot reach another app), fail-closed hash/byte verification, org-vs-user address derivation, spill past the inline cap and back, and "still there but would not read" keeping its entry instead of reading as a deletion |
| `src/checking/floor.ts` | 14.28% | 100% | `createAppFloor` — the paint-seam floor whose own docblock says it was "structurally dead for every other author". Now proven to expand inline tool refs and produce real binding errors, with a side-by-side against a bare `compileWire` so the regression is observable |
| `src/testing/scripted-model.ts` | 77.14% | 100% | the shipped fixture ~40 suites measure against, driven through the real `ai` SDK (`generateText`/`streamText`) rather than by hand |

**No mock-the-counterparty tests.** The env is driven for real; `e2bInstalled()`
really probes the module resolver (the absent case names a package that
genuinely is not installed, using the specifier seam `e2bInstalled` already
exposes and `e2b/installed.test.ts` already uses); the scripted model is driven
by the actual SDK; and checkout/commit share ONE workspace so the two ends can
genuinely disagree.

**Proven able to fail.** One mutation per new test file, each watched go red and
then reverted:

| mutation | tests that went red |
| --- | --- |
| drop `.trim()` in the ladder's env reader | 2 — the whitespace-key cases |
| `if (await workspace.exists(…))` → `if (false)` | 2 — "stale beats gone", and the sibling-files case |
| `compileWire(text, wireCompileOptionsFor(…))` → `compileWire(text)` | 3 — the whole shared-dialect group |
| drop the 40-char name cap | 1 — the display-title cap |

One production line changed: `selectSandbox` now takes the optional
`e2bSpecifier` the probe underneath it already accepted, so rung 2's refusal is
reachable without stubbing. Everything else is tests.

PR #863, opened 2026-08-06
refactor(vendo,agent)!: delete createAgent and @vendoai/agent — one path (S6)

yousefh409:
The last slice of **one path**: `@vendoai/agent` ceases to exist. Hosted stores
serve harness turns (S1–S3) and the survivors moved (S5), so the only reason the
legacy chat engine existed — a hosted deployment that could not serve a harness
turn had to fall back somewhere — is gone. The harness runtime is now the only
engine, and `POST /threads` has one door.

Base: `one-path/s5-survivors` merged with `one-path/s3-surgical-undo` (one
trivial conflict in `server.ts`, resolved as S5's shape carrying S3's ops-aware
probe text; both sides then deleted by this slice).

**~8,000 lines deleted. Nothing a client can see changes.**

---

## What moved, what died

| | |
| --- | --- |
| `packages/agent/**` | deleted whole — `createAgent`, `runner.ts`, `agent.stream`, `ids.ts`, `prompt.ts` (the umbrella already had its copy), `test-helpers.ts`, 22 suites |
| `agent/src/threads.ts` (+ `mintThreadId`, inlined) | → `packages/vendo/src/threads.ts`, its one remaining consumer's package (`Thread` carries ai's `UIMessage`; the store block deliberately has no `ai` dep) |
| `Vendo.agent`, `WireDeps.agent`, `VendoAgent` | gone. `WireDeps.harness` is required and carries both halves: `Pick<HarnessTurns, "stream" \| "threads">` |
| `storeServesHarnessTurns` | deleted. A store with neither SQL nor StoreOps refuses **that turn** inside the door, naming both ways to give it a home. Boot stays up. |
| `fixtures/chat-e2e/**` | deleted with its subject — see the mapping below |
| bench `agent-context` | drives `createVendo(...).harness.threads.list` |
| dependency-guard, corpus/existing-agents pack lists, `FORBIDDEN_SERVER_MODULES`, root `conformance` script, README, docs-site, `docs/quickstart.md` + both `.docs-check.ts` mirrors | updated |

## The keyless-error parity fix (S5's finding)

`wireErrorMessage` is the one gate the legacy door put in front of every failure.
The runtime consulted it only for failures the harness *reported* as an `error`
event; for a harness that **threw**, and for anything that failed before the
harness ran (`execute` rejecting — credential mint, toolset build, workspace
mount), it substituted the constant `HARNESS_FAILED`.

So a keyless deployment migrated onto the harness path was told *"Something went
[...]

PR #864, opened 2026-08-06
fix(store,vendo): check-round fixes — refusal atomicity, create-CAS parity, dead surface (S8)

yousefh409:
Four confirmed defects from the one-path check round. Every fix was proven red
on the pre-fix tree first, including against the real console.

## 1. The no-SQL/no-ops refusal was a half-write

`createHarnessTurns().stream` called `sqlDoors()` — where the refusal is raised
— AFTER `threads.persist`, which rides the adapter seam and therefore succeeds
on a store that can keep neither the transcript nor the workspace. Every refused
turn left a `vendo_threads` row carrying the user's message on a deployment that
can never answer it, and a retrying client kept growing that thread.

The doors are now resolved before the first write.

RED (pre-fix): `expected [ { id: 'thr_refused', … } ] to deeply equal []`
GREEN: adopted `harness-refusal-atomicity.test.ts` passes, with `harness-wire`,
`harness-threads` and `harness-system-prompt` still green.

*Behaviour note for review:* tour mode (`config.scripted`) returns between the
persist and the old `sqlDoors()` call, so it is now refused on a no-SQL store
too instead of quietly working. That path's documented `agent.stream` fallback
was deleted in S6; `tours.e2e.test.ts` and `tours/tours.test.ts` are green. Say
the word if tour mode should be exempted.

## 2. A strict mount dropped its guard on a CREATE

`/orgs` commit policy is strict compare-and-swap. The SQL backend enforces it
for a create — `landRow` compares `prepared.prior?.revision ?? null` against the
checkout base. The hosted backend required a number:

```ts
const strict = options?.strict === true && typeof options.expectedRevision === "number";
```

`WorkspaceStoreFs.commit` sends `index.get(path)?.revision ?? null`, and for a
path the caller never checked out that is `null`. So `strict` went false, the
entry crossed the wire unguarded, and the console's `files.put` silently
overwrote the colleague who created the shared file first.

`expectedRevision` now has three states end to end — a number compares, `null`
means "this path must not exist yet", the absent field is unguarded:

- `store-wire.ts` accepts `null`
- the local ops backend passes it to `landRow`, which already compared correctly
- the memory reference reads an absent path as `null` so it matches
- the conformance kit gains the create-only case, holding **both** backends to it
- console counterpart: runvendo/vendo-web#282 (merged + deployed)

RED, proven three ways:
- memory conformance: `the workspace moved on under create-cas.json` — the FIRST
  create was refused, because `undefined !== null`
- SQL conformance: `workspace entry create-cas.json has a non-numeric expectedRevision`
- **live, against the real console**: `expected { status: 'ok' } to deeply equal
  { status: 'conflict' }` — Bob's write really did overwrite Alice

GREEN: 53/53 memory, 50/50 SQL, and the live seam test passes.

## 3. A stale `it.fails`

`hosted-transcript.live.test.ts` still carried the KNOWN-GAP `it.fails` for the
approval flip. The console gap is closed, so it is a plain `it` and the comment
is gone. It passes live.

## 4. Dead surface

Eleven exports left `@vendoai/harnesses` with no reader anywhere — verified with
ripgrep over `packages/`, `examples/`, `fixtures/`, `corpus/`, `docs/`,
`docs-site/` and `scripts/`; every caller reaches them relatively, and neither
barrel uses `export *`. `StoreBackend`, `GENERIC_TURN_ERROR`,
`WORKSPACE_BYTES_KEY` and the two byte converters never left their file either.
The `./vendo` subpath is untouched.

## Live suite (after the console deploy)

| file | result |
|---|---|
| `hosted-workspace.live.test.ts` | 6/6 (incl. both S3 per-path undo cases) |
| `hosted-workspace-cas.live.test.ts` | 1/1 |
| `hosted-transcript.live.test.ts` | 4/4 |

**11/11 green, no `it.fails`, no skips beyond env-gating.**

## Gate

`build`, `typecheck` and `lint` (dependency-guard + portability) are clean.

`pnpm test` shows 5 failures **local to this worktree's checkout path, which
contains a space** — `packages/core` packaging e2e (3) and the two
`packages/store` SIGKILL drills both fail on `Cool%20Code` URL-decoding, and two
`dev-creds/model.test.ts` provider-resolution cases fail on node_modules layout.
None touch changed code, and the base branch (#863) is green in CI.

The `automations-e2e` fixture flakes in this environment; I A/B'd it 3 runs per
side on the same machine: **base 0a2da8bbd failed 5, then 1, then 0**, while
**this branch's tip failed 0, 0, then 1** — same tests, pre-existing, and
marginally better here. `fixtures/integration` is 51/51 with one server-boot
ECONNREFUSED that passes on re-run.

---
## Summary by cubic
Deletes `@vendoai/agent` to serve all chat turns through the harness runtime, fixes refusal atomicity, enforces strict create-CAS across backends, and preserves crafted wire errors so users get actionable guidance. Prevents half-writes on non-SQL stores and stops silent overwrites of shared files.

- **Bug Fixes**
  - Refusal is atomic in `@vendoai/vendo`: resolve doors before any write, so refused turns no longer persist `vendo_threads` rows. Adds `harness-refusal-atomicity.test.ts`.
  - Strict mount create-CAS parity across backends: `expectedRevision` now supports number | null | absent end to end (`@vendoai/core`, `@vendoai/store`). `null` guards creates; wire accepts `null`; local ops and hosted backends refuse overwriting creates. Adds conformance case and `hosted-workspace-cas.live.test.ts`.
  - Harness runtime surfaces specific failures: `wireErrorMessage` now keeps Vendo-crafted errors (provider 401/login guidance, Cloud meter) via `specificWireErrorMessage`; adds `packages/harnesses/src/provider-401.test.ts`. Hosted approval flip is now a plain passing expectation.
  - Rebuilt `prompt-block-forgery.test.ts` helpers in `@vendoai/vendo` to remove the deleted agent dependency; assertions unchanged.

- **Migration**
  - `@vendoai/agent` is deleted. Serve turns through the harness path (`@vendoai/harnesses` + `vendo()`); `Vendo.agent` and `VendoAgent` are removed. Use the harness surface (`Vendo.harness`) and `@vendoai/agents` runners where needed; threads types move to `@vendoai/vendo` (`./threads`).
  - Non-SQL/non-ops stores no longer fall back to the legacy engine; each turn is refused with an explicit message.
  - For strict mounts, send `expectedRevision: null` when creating a file you didn’t check out, to guard against overwriting a colleague’s create.
  - `@vendoai/harnesses` drops unused exports (approval/transcript helpers, tool-bridge helpers, tool-search helpers, capability-miss helpers). Update imports to supported APIs or local modules.

<sup>Written for commit 3e2a76ef03b57df77ec8015dd66362f4a5ba2ef8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/864?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
## Fresh-eyes cold walk of the merged one-path work

Cold verification by an agent that built none of this, on `origin/main` @ `f7c6da2ab`, after `pnpm install && pnpm build`. Real browser (Playwright), real keys, Yousef's Vendo Cloud account. Screenshots live on the runner at the paths below (GitHub can't take image uploads from the CLI, so each is described).

**Verdicts**

| Walk | Result |
|---|---|
| 1 — keyless local OSS (demo-bank, BYO Anthropic key, local PGlite) | **PASS with 2 findings** |
| 2 — hosted, key-only (`VENDO_API_KEY`, no store slot) | **PASS** (slow, 1 noisy log) |
| 3 — live provider matrix (`packages/harnesses/src/vendo/live.test.ts`) | **PASS** (Anthropic leg; the other two providers are env-gated) |

---

### Finding 1 — BLOCKER for the keyless story: the persisted failure line loses the `vendo login` guidance

The whole point of the persisted turn error is that a keyless host reloads the page and still sees what to do. The live banner is right; the thing that survives the reload is not.

- **Expected** (and asserted in `packages/harnesses/src/provider-401.test.ts:141`, "a keyless turn is told to run `vendo login`, whichever half of the turn died"): persisted message = `Vendo: Vendo found no model key. Set ANTHROPIC_API_KEY / … or run \`vendo login\` … (validation)`
- **Observed** on the wire, one POST to `/api/vendo/threads` with no model key:

```
data: {"type":"error","errorText":"Vendo: Vendo found no model key. Set ANTHROPIC_API_KEY / OPENAI_API_KEY / GOOGLE_GENERATIVE_AI_API_KEY in .env.local (with the matching @ai-sdk provider installed), or run `vendo login` for a free dev key. Production always needs a real server-side key. (validation)"}

data: {"type":"data-vendo-turn-error","data":{"message":"Something went wrong on my side, so I stopped."}}
```

The transient chunk is specific; the **persisted** part is the generic constant. After reload the user sees a red chip reading "The response didn't finish" and nothing actionable.

**Why the green test doesn't catch it.** `failedTurn()` in that test drives a harness that *throws*, which lands in `runtime.ts:542` where `specificWireErrorMessage(error)` sees the real `VendoError` and wins. The shipped `vendo()` harness never throws here — `vendo.ts:375` catches and *yields* `{type:"error", message: wireErrorMessage(error)}`. On that path `runtime.ts:515` `writeError(...)` re-enters the SDK's `onError` (the code comment at `runtime.ts:576-579` already knows it does) *before* line 519 runs; `onError` at line 592 receives a plain `Error`, not a `VendoError`, so it records `HARNESS_FAILED` and flips `turnErrorRecorded`, and line 519's specific record is then a no-op. Matching evidence: no `[vendo] harness stream error` line in the server log, because `text !== surfaced` was false.

The seam that is mocked is the harness itself — exactly the class of gap called out in `CLAUDE.md`.

Evidence: `/tmp/one-path-proof/w1-06-keyless-failure-line.png`, `/tmp/one-path-proof/w1-07-FINDING-persisted-failure-generic-after-reload.png`, `/tmp/one-path-proof/walk1b-keyless.log`.

### Finding 2 — thread delete has no user-facing affordance

The walk asks "delete works". Over the wire it does: `DELETE /api/vendo/threads/:id` with `content-type: application/json` → 200, and the row is gone from `GET /threads` and from the rail after reload. But there is no way for a user to reach it: `CenterChats` (`packages/ui/src/chrome/center/rail.tsx:297`) renders each thread as a bare button — no hover menu, no context menu, no overflow item — and `packages/ui/src` contains no "Delete" string at all outside the generated jail bundle. `client.threads.delete` and `useThreads().remove` exist with no caller.

Evidence: `/tmp/one-path-proof/w1-04a-thread-rightclick.png` (right-click on a thread row selects text; no menu), `/tmp/one-path-proof/w1-05-thread-deleted-list-empty.png` (list empty after the wire-level delete).

### Finding 3 — cold `pnpm build` can leave the dev server unbootable

On a warm turbo cache, `@vendoai/ui#build` is a cache hit, so its `prebuild` (`build:jail`) never runs. `turbo.json` declares outputs as `dist/**` only, but `build:jail` generates into **`src/`** (`packages/ui/src/tree/jail/runtime-bundle.gen.ts`), so the cache restores `dist` and the source file stays missing. `pnpm --filter demo-bank dev` then 500s on every page with `Module not found: Can't resolve './runtime-bundle.gen.js'`. `pnpm --filter @vendoai/ui run build:jail` fixes it. Pre-existing, not one-path, but it is the first thing a fresh clone hits after a cached build.

Evidence: `/tmp/one-path-proof/00-finding-jail-bundle-500.png`.

### Minor observations

- React key warning in the thread rail on every load: `Each child in a list should have a unique "key" prop. Check the render method of \`CenterChats\`` — shows as the persistent "1 Issue" badge in every screenshot.
- The persisted failure chip is `white-space: nowrap; text-overflow: ellipsis` at 144px, so even the 26-character generic sentence is clipped to "The response didn't f…". Whatever Finding 1 restores will not be readable in that chip as it stands.
- Hosted boot prints the "Vendo Cloud is the hosted store for this deployment…" paragraph ~20+ times.
- During the hosted app build the server logged `[vendo] render seam: source did not reach the store … has no row to hold its source` twice, yet the source *did* land (verified below). Misleading error line.
- A hosted app build took **3.9 min** end to end (`POST /api/vendo/threads 200 in 3.9min`).

---

### Walk 1 — keyless local (BYO Anthropic key, `MAPLE_STORE=local`)

Chat answered for real (`Your checking account balance is $9,412.20`, one tool call, 2.5s); the thread appears in the rail and its messages survive a reload; delete works over the wire (see Finding 2); with the key removed the turn fails and the failure persists across a reload (see Finding 1); with the key restored, Regenerate produces a clean answer and the stored thread has **no** `data-vendo-turn-error` part left.

### Walk 2 — hosted, key-only

`VENDO_API_KEY` alone, store slot unset. Log confirms `model: VENDO_API_KEY (Vendo Cloud) → vendo via the Cloud gateway` and the hosted store. Chat answered for real over the store wire. Asked it to build an app ("Dining Watch"); the app rendered live in the thread. Verified independently with a **fresh store-wire read using only the key** (no host process involved), `POST https://console.vendo.run/api/v1/store/records/…/list`:

- `vendo_threads` → my thread `thr_e9e471c5…` created 12:36:28Z
- `vendo_apps` → `app_b026a550…` "Dining Watch" with its tree
- `vendo_workspace_files` → `["vendo-demo","/user/apps/app_b026a550…/app.vendo"]` — the workspace file, with its `<App name="Dining Watch">` source

Then killed and restarted the server: the thread, its transcript, and the rendered app all came back from the Cloud. I did not use the console UI (no browser auth on this runner) — the store-wire read is the substitute, as allowed.

### Walk 3 — live provider matrix

```
npx vitest run src/vendo/live.test.ts   # VITEST_MIN/MAX_FORKS=1/2, MIN/MAX_THREADS=1/2, ANTHROPIC_API_KEY set
✓ provider matrix live — Anthropic > runs the shared approval, blocked-outcome, and view-part journey  8901ms
Test Files 1 passed (1) · Tests 1 passed | 2 skipped (3)   EXIT=0
```

All three *legs* of the journey (parked approval that runs once, blocked call that never runs, tree that reaches the screen) pass against a real Anthropic model. The other two matrix entries are separate providers and are env-gated: OpenAI needs `OPENAI_API_KEY` and the proxy needs `VENDO_TEST_PROXY_URL`/`_KEY`, which are not in the key file. For completeness I re-ran with the `OPENAI_API_KEY` from the key file and it fails at 316ms with `AI_APICallError … Incorrect API key provided` (HTTP 401) — a **stale credential in the key file**, not a code defect. That leg is currently unprovable.

---

### Screenshot manifest (`/tmp/one-path-proof/`)

| File | What it shows |
|---|---|
| `00-finding-jail-bundle-500.png` | Finding 3 — demo-bank 500s on every page after a cached build (missing jail runtime bundle) |
| `w1-01-composer-typed.png` | Keyless-local Maple, Ask Maple workspace, question typed into the composer |
| `w1-02-first-answer.png` | Real answer "$9,412.20" with "Did 1 thing · 2.5s"; thread now listed under TODAY |
| `w1-03-thread-persisted-after-reload.png` | Same thread and both messages restored after a full page reload |
| `w1-04a-thread-rightclick.png` | Finding 2 — right-click on a thread row offers no menu (only text selection) |
| `w1-05-thread-deleted-list-empty.png` | Rail empty after the wire-level DELETE and a reload |
| `w1-06-keyless-failure-line.png` | Finding 1 — no model key: red "The response didn't f…" chip, no guidance |
| `w1-07-FINDING-persisted-failure-generic-after-reload.png` | Finding 1 — the same generic chip survives a reload; still no `vendo login` |
| `w1-08-retry-after-key-restored.png` | Key restored → Regenerate gives a clean 3-tool answer, no stale failure line |
| `w2-01-hosted-empty-workspace.png` | Hosted key-only boot, Vendo workspace empty state |
| `w2-02-hosted-answer.png` | Hosted turn answers for real ("Get account" → $9,412.20) over the store wire |
| `w2-03-hosted-app-build-progress.png` | "I'll build this now" — app build running against Cloud |
| `w2-04-hosted-app-built.png` / `w2-05-hosted-app-wait2.png` | Same turn still building at ~1.5 and ~2.5 min (see 3.9 min note) |
| `w2-06-hosted-app-turn-complete.png` | "Dining Watch" rendered in-thread with real transaction rows |
| `w2-07-thread-persists-across-restart.png` | After killing and restarting the server: thread, transcript and rendered app all back from the Cloud |

Also on the runner: `walk1-dev.log`, `walk1b-keyless.log`, `walk1c-restored.log`, `walk2-hosted.log`, `walk2b-restart.log`, `walk3-live.log`, `walk3-live-both.log`, `w2-storewire-workspace-files.json`.

No product code was changed. All processes started for this walk were killed.

PR #889, opened 2026-08-06
feat: placement core — rows, runtime, wire, client, slot read

yousefh409:
Placement moves off `doc.placements` and onto rows in the generic records
collection. A slot can now show a build that is still running, one that failed,
and a ready app — and pinning is a Vendo write, so Maple's hand-rolled
`/api/demo/pin` route is deleted.

- rows: `packages/apps/src/placements.ts` (`vendo_placements`, id = subject+slot, refs `{subject, slot}`)
- runtime: `place` / `unplace` / `placements`; `create({ slot })` claims the slot at mint; `delete` sweeps
- wire: `GET /apps/placements?slots=…` (registered before the `/apps/:appId/*` catch-all), `POST /apps/:id/place`, `POST /apps/:id/unplace`
- client + ui: `apps.place/unplace/placements`; one placement poller per client behind `useSlotApp(slot) → { appId, status }`; `VendoSlot` mounts only a READY app
- pin: `usePinAction` awaits `apps.place`, then announces, then calls `onPin`

Seam proof: `packages/vendo/src/placements-seam.e2e.test.ts` — placed through the
real wire route, read back through the real `GET /apps/placements`, and a
slot-targeted create watched from BUILDING to READY with nothing stubbed.

No UI-visible change to a slot's rendering in this PR (the building/failed slot
states land in PR2), so no screenshots.

### Behaviour removed on purpose

The fork mint no longer writes `doc.placements`; the assertion pinning that
write (`packages/apps/src/fork-pin.test.ts`) is deleted as spec-removed
behaviour. `doc.placements` itself stays on `AppDocument` and its legacy
classifier keeps its tests — deleting the field is a separate change.

### Known-red, pre-existing

`packages/vendo/src/dev-creds/model.test.ts` fails 2 cases (`@ai-sdk/anthropic`
module resolution from a temp dir). Verified identical at the base commit
`2e0573978` with this branch checked out — unrelated to placement and not fixed
here. Everything else is green: build 24/24, typecheck 45/45, lint 3/3,
`test:affected` 41/42 tasks with those two cases the only failures.

---
## Summary by cubic
Moves placements off app documents into rows with new placement APIs and wire routes. Slots can now show building and failed apps; pinning uses Vendo’s write, and clients share a single placements poller.

- Rows in `vendo_placements` keyed by `{subject, slot}`; enforced one app per slot.
- Runtime: `place`, `unplace`, `placements`; `create({ slot })` claims the slot at mint.
[...]

PR #897, opened 2026-08-06
feat: agent addressing — vendo_make{slot}, pin/unpin, door withholdTools

yousefh409:
Second PR in the slots stack. **Base is `yousefh409/slots-placement-core` (PR1), not `main`.**

PR1 shipped the placement rows and the runtime verbs. This PR teaches every AGENT door to address a slot on the user's page.

## What's in it

**1. `vendo_make` grows one optional `slot`.** Honoured on BOTH engines the one front door routes to (#862 made `vendo_make` one engine with a routing seam behind it):

- **assembly** — `runtime.place({ app, slot })` the moment the stored row is confirmed. `place()` gates on the app record, and assembly is what writes that record, so this is the first instant the claim can legally be made.
- **escalated build** — `create({ slot })`, which claims at mint, so a build still running occupies the place the caller aimed at instead of appearing out of nowhere minutes later.

Threading `slot` into `create()` alone would have shipped a **dead feature** for the default path — assembly serves most asks and never calls `runtime.create`. Both engines are asserted, in `packages/apps/src/agent-tools.test.ts` and again over the real MCP door.

On a CHANGE (`app` present), `slot` is refused by name: silently moving an existing app would evict whatever holds that slot off the back of an edit nobody aimed there. The error names `vendo_apps_pin`.

**2. `vendo_apps_pin` / `vendo_apps_unpin`.** Put an app the user already has into a slot and take it back out. Aimed by the app's id OR the name the user said, like every other door in this registry. Pin reports what it replaced as `evicted`. Risk `write` — a placement row is small and reversible.

Neither is offered to an unattended run. `PRESENCE_ONLY_TOOLS` (core) joins THE LAW's projection in `projectableForRun`: a tool whose whole effect is on a person's screen has nothing to act on when nobody is looking, and rearranging a dashboard someone comes back to is a change they never saw being made. **Keyed on the NAME, not the grade** — grading them `destructive` to buy the withholding would lie to every other reader of the label (policy rules, consent card, automations planner).

**3. `McpDoorConfig.withholdTools`.** Names one door never offers, checked **before** the `vendo_` prefix bypass. The bypass exists so a HOST's curated menu cannot delete Vendo's own plumbing; this is the deployment's own decision about its own door. Listing, call path, and the apps ride-along path all read the one set, so a withheld name cannot be invisible and callable. Curation, not security: a withheld name answers with the same in-band not-found an unknown name gets.

## Pinned surfaces that moved

Three existing test files pin surfaces this PR adds to, and moved for that reason only:

- `packages/vendo/src/mcp-door-outside-agent.e2e.test.ts` — the sorted set an outside agent sees gains `vendo_apps_pin` / `vendo_apps_unpin`
- `packages/vendo/src/default-composition.test.ts` — `DEFAULT_TOOL_NAMES` gains the same two
- `packages/vendo/src/vendo-make.e2e.test.ts` — the `vendo_make` schema pin gains `slot`

## Proof

`packages/vendo/src/placement-tools.e2e.test.ts` is new and crosses the seam in both directions with no stub on either side: the write goes in over the real MCP door (register to authorize to token to JSON-RPC) through the real guard-bound registry; the read comes back out of the real store's `vendo_placements` rows. The projection is measured on two real turns, present and away.

Three revert-and-watch runs, each confirming the test fails for the right reason and only that reason:

| Reverted | Went red | Message |
| --- | --- | --- |
| `...(slot === undefined ? {} : { slot })` on the escalated `create` | only the ESCALATED unit test | `expected [] to deeply equal [ ObjectContaining{...} ]` |
| `runtime.place(...)` on the assembled branch | only the slot-targeted e2e | `expected [] to have a length of 1` |
| `!PRESENCE_ONLY_TOOLS.has(name)` in the projection | only THE LAW e2e | `expected [ ..., 'vendo_apps_pin', ... ] not to contain 'vendo_apps_pin'` |

## Two contradictions found during recon

1. **"withheld via the existing projection" + "risk `write`" cannot both be true without a name-keyed rule.** `projectableForRun` withholds on `risk === "destructive" | "ungraded"` only, so a `write` is projected into every unattended run. A ctx-aware `descriptors(ctx)` would be dead on the composed path — packs are gone (#848) and the apps surface is a static capability contribution (`toolsFromRegistry(...)` in `server.ts`), while the actions registry's `descriptors()` takes no ctx at all. Hence `PRESENCE_ONLY_TOOLS` inside `projectableForRun`: same mechanism, second predicate, keyed on name.

2. **B1's "claim at mint" no longer covers the engine that serves most asks, so the row is written per engine.** PR1's mint-time claim lives inside `create({ slot })`, and `place()` gates on the app record, which does not exist at mint — so nothing PR1 ships can claim a slot at the mint line. **The cost:** a slot-targeted make whose ASSEMBLY fails leaves the slot untouched instead of showing the failure there (the builder path is unaffected — it still shows `building`, then `failed`). Honouring B1's letter on both engines would need PR1 to expose its mint-time claim as a runtime verb, which is a PR1 contract change, not this PR's to invent. Flagging rather than improvising.

## Gate

`pnpm build`, `pnpm typecheck`, `pnpm lint` — all green.

`pnpm test:affected` — green except three **pre-existing** failure groups, none of them touched by this PR:

- `packages/vendo/src/dev-creds/model.test.ts` (2 cases) — the known red on clean `main`, same as PR1 (#889).
- `packages/core/src/packaging.e2e.test.ts` (3 cases) — verified pre-existing by stashing this branch's changes and re-running on the untouched base.
- `packages/store/src/durability.drill.test.ts` + `split-brain.drill.test.ts` (2 cases) — `packages/store` is untouched by this PR (`git diff --stat base..HEAD -- packages/store` is empty). Root cause is `new URL(...).pathname` keeping percent-encoding, so a checkout path containing a space spawns `.../Cool%20Code/.../drill-writer.mjs` and Node cannot resolve it. An artifact of the worktree this ran in, not of the change; left alone under the scope fence.

---
## Summary by cubic
Agents can now place apps in specific page slots. `vendo_make` accepts a slot, new pin/unpin tools move existing apps, and `withholdTools` lets deployments hide tools; presence rules keep page-only actions out of unattended runs.

- New Features
  - `@vendoai/apps`: `vendo_make` adds optional `slot`. Assembly places on row confirm; escalated builds claim at mint. Refused on edits that pass `app`.
  - `@vendoai/apps`: `vendo_apps_pin` and `vendo_apps_unpin` to place/remove existing apps by id or exact name; pin returns `evicted`. Risk `write`.
  - `@vendoai/core`: `PRESENCE_ONLY_TOOLS` keeps `vendo_apps_pin`/`vendo_apps_unpin` off unattended runs (keyed by name, not grade).
  - `@vendoai/mcp`: `McpDoorConfig.withholdTools` to never offer named tools, checked before the `vendo_` prefix bypass; listing, calls, and apps ride-along all honor it.

- Migration
  - No breaking changes; `slot` is optional.
  - For doors without a page UI, set `withholdTools` to include `vendo_apps_pin` and `vendo_apps_unpin`.

<sup>Written for commit c0277018cb7d91c03dfb5a1d42818ded668f94f7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/897?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/agent-addressing-slots.md
+++ b/.changeset/agent-addressing-slots.md
@@ -1,0 +1,34 @@
+---
+"@vendoai/core": minor
+"@vendoai/apps": minor
+"@vendoai/mcp": minor
+---
+
+Agents can address a place on the page.
+
+`vendo_make` takes one optional `slot`, honoured on both engines the one front
+door routes to: an assembled screen claims the slot the moment its row lands,
+and an ask that escalates to a real build claims it at mint — so a build that is
+still running occupies the place the caller aimed at instead of appearing later
+out of nowhere. On a CHANGE, `slot` is refused by name: silently moving an
+existing app would evict whatever holds that slot off the back of an edit nobody
+aimed there.
+
+Two new tools do the moving. `vendo_apps_pin { app, slot }` puts an app the user
+already has into a slot and reports what it replaced as `evicted`;
+`vendo_apps_unpin { app, slot }` takes it out and leaves the app itself alone.
+Both aim by the app's id OR the name the user said, like every other door in
+this registry, and both are graded `write` — a placement row is small and
+reversible.
+
+Neither is offered to an unattended run. `PRESENCE_ONLY_TOOLS` (core) joins THE
+LAW's projection: a tool whose whole effect is on a person's screen has nothing
+to act on when nobody is looking, and rearranging a dashboard someone comes back
+to is a change they never saw being made. Keyed on the name, not the grade, so
+policy rules and consent cards still read an honest `write`.
+
+`McpDoorConfig.withholdTools` — names one door never offers, checked BEFORE the
+`vendo_` prefix bypass. The bypass protects the product's plumbing from a HOST's
+curated menu; this is the deployment's own decision about its own door. Curation,
+not security: a withheld name answers with the same in-band not-found an unknown
+name gets.
--- a/packages/apps/src/agent-tools.test.ts
+++ b/packages/apps/src/agent-tools.test.ts
@@ -11,7 +11,9 @@ import { agentToolDescriptors } from "./agent-tools.js";
 import { createApps, type AppsRuntime } from "./index.js";
 import {
   authoringAssembler,
+  basicLanguageModel,
   bindTools,
+  fakeBoxSandbox,
   guardFixture,
   memoryStore,
   seedAppRow,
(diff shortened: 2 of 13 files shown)


yousefh409:

PR #908, opened 2026-08-06
fix(apps,guard,mcp,ui): risk-round findings — placement masking, atomic eviction, presence-only law, withheld tools

yousefh409:
Risk-check round on `yousefh409/slots-display-picker` @ 64e6f786. The checker's
probes are cherry-picked in as `ba47f46a3` (untouched except where noted under
finding 7); this branch is the product-code answer to them. **All six findings
are fixed and all of the checker's assertions pass.**

## 1 — `GET /apps/placements` leaked a masked app's title and build status

`entryFor` did a raw `apps.get(row.appId)`, so a viewer whose grant was taken
back kept reading a placed app's **title** and its **live build status** long
after `open`/`get`/`list` had all gone back to `not-found`.

**Fix** (`packages/apps/src/runtime.ts`) — every entry now passes the same
`holds(appId, ctx, "viewer", record)` check those paths use, and a slot the
caller may no longer view is dropped from the answer. The owner fast path means
single-player placements stay at one store read per row. The **no-record** arm
is deliberately untouched: a build in flight has no document to mask, and both
fields it returns come off the caller's own placement row.

**Proves it:** `placement-risk.test.ts` › "keeps naming an app the caller lost
access to, after open() has gone back to not-found" ✅

## 2 — `place()` was read-then-write, so the eviction receipt could lie

Two concurrent places into one slot both answered `evicted: undefined` while one
of them was silently displaced.

**Fix** (`packages/apps/src/placements.ts`) — a place is one decision:
`atomic.insertIfAbsent` for an empty slot, `atomic.compareAndSwap` on the slot
pointer's revision otherwise, and the loser retries against the winner.
`RecordStore.atomic` is present on every adapter Vendo ships for a generic
collection (PGlite `records.ts`, hosted store, the conformance memory double); a
BYO adapter without it keeps read-then-write, which is what `AtomicRecordStore`'s
own contract says absent adapters do.

**Proves it:** `placement-risk.test.ts` › "loses one of two concurrent places
into the same slot and reports no eviction" ✅

## 3 — `unplace()` could delete the row that replaced it

A placement is now **two rows**:

| row | id | meaning |
|---|---|---|
| pointer | `plc:<subject>:<slot>` | who holds the slot, under which token. Never deleted; the single CAS arbitration point, so finding 2's receipt is untouched. |
| live | `plcv:<subject>:<slot>:<token>` | exists iff *that* placement still holds the slot. |

Occupied means "the live row the pointer names exists". A clear is
`delete(liveId(token))` and nothing else, and tokens are never reused — so a
stale client's delete can only ever address **its own** placement. Keyed on the
shared `(subject, slot)` id alone, that same delete landed on whatever now
occupied the id, which is exactly the row it must not touch.

The live row is written *before* the pointer swings to it, so a slot never reads
empty mid-replace; a placer that loses the CAS takes its own unnamed row back
out rather than leaving an orphan.

Pointers live in their own generic collection (`vendo_placement_slots`), which
keeps `vendo_placements` at exactly one row per **live** placement — the count
`packages/vendo/src/placement-tools.e2e.test.ts` reads across the seam (green,
unchanged). Neither collection is reserved or dedicated, so both route to the
shared `vendo_records` table: no migration, and `erase.bySubject` sweeps both on
`refs @> {subject}`.

**Proves it:** `placement-risk.test.ts` › "clears the slot the winner just took
when a place lands inside unplace's read-then-delete" ✅

> **Correction on the record.** My first pass shipped a one-row fix (delete
> scoped to the appId via `RecordStore.claim`) and reported this probe as
> unsatisfiable, with a two-branch argument: a shared-id delete fails the last
> assertion, an atomic compare-and-delete fails the middle one. Both branches
> are real — I ran them — but the argument assumed the row you *arbitrate* on
> and the row you *delete* must be the same row. They need not be. The probe is
> a specification for version-keyed row ids, and it was right. The one-row
> impossibility result does survive in that narrower form: under one row per
> slot, lines 138 and 142 of the probe are mutually exclusive.

## 4 — `vendo_make {slot}` placed from a run with nobody there

`vendo_make` is honestly graded `read`, so no projection withheld it and the
guard's risk-keyed defence never fired — while its `slot` made the same evicting
write `PRESENCE_ONLY_TOOLS` exists to keep off an unattended run.

**Fix** (`packages/apps/src/agent-tools.ts`) — a slot is presence-only at the
source: with nobody there the build still runs and the slot is simply not
claimed. Plain makes keep their honest `read` grade; the "you aimed a new app at
a slot on an EDIT" refusal still reads the raw `slot`, because that is wrong
however present the person is.

**Proves it:** `placement-risk.test.ts` › "rearranges the page anyway — the
presence rule catches the pin tool and not this" ✅

## 5 — the presence-only law had one layer; now it has two

`projectableForRun` only decides what a run is **offered**. A standing automation
grant, a resumed step, or a harness that calls without listing reaches
`execute()` by name regardless — and these tools are honestly `write`, so the
choke point's risk-keyed test never spoke for them.

**Fix** — `presenceOnlyCall()` in `packages/core/src/grant-sets.ts` states the law
as a **call** (the pin/unpin names, plus a make that carries a slot — the slot
rides in the arguments, so only the call can say). `packages/guard/src/guard.ts`
consults it at the same choke point that already refuses destructive and
ungraded work, with the same `UNATTENDED_DESTRUCTIVE_REASON` and the same
park-and-replay exemptions.

**Proves it:** `guard/test/security/unattended-presence-only.test.ts` — all four
cases ✅

## 6 — `withholdTools` was ignored on the turn-bearing leg

A door with `turnCredentials` has a second credential space on the same mount.
`#handleMcp` resolved a turn bearer first, then listed `turnTools(state.turn)`
and handed `#callTool` straight to `turn.tools.call` — neither consulted
`#withheld`. A name the deployment said it never offers was both advertised and
callable there.

**Fix** (`packages/mcp/src/door.ts`) — the same `#withheld` set filters the turn
listing and refuses the turn call, with the identical `not-found` in-band error
the OAuth leg produces. No parallel mechanism, no new error shape.

**Proves it:** `door.test.ts` › "createMcpDoor withholdTools on a turn-bearing
session" — both cases ✅ (whole file green).

## 7 — not a bug; the checker's case is removed

The failed arm of `VendoSlot` **intentionally** replaces host fallback markup —
that is what makes "Try again" and "Clear this slot" reachable at all, and the
building-vs-failed asymmetry is deliberate. Per the conductor's ruling I deleted
only the case the checker added for this
(`packages/ui/test/chrome/slot-states.test.tsx`) and left a comment in its place
stating the rule so nobody re-adds it.

## Gate

`pnpm build && pnpm test:affected && pnpm typecheck && pnpm lint` — build ✅,
typecheck ✅ (45/45), lint ✅ (one pre-existing demo-bank warning).

Tests: **no new reds.** The only failures are pre-existing on the base tip —
`core/packaging.e2e` ×3, `store` durability/split-brain drills ×2,
`vendo/dev-creds/model` ×2. 47 of 50 task targets green, and the three red
packages are exactly those three.

`apps` 1073 passed / 19 skipped (all four placement probes green) · `guard` 269
· `mcp` 129 · `ui` 1087.

## Test-file changes, stated out loud

- `packages/apps/src/placement-rows.test.ts` — the refs case pinned the physical
  id `plc:user_ada:home-hero` inside `vendo_placements`; the live row's id now
  carries its token, so it is found the way the erase cascade finds it (by refs)
  and the pointer's refs are asserted too. A second case pins the
  one-live-row-per-slot count directly.
- `packages/ui/test/chrome/slot-states.test.tsx` — finding 7 above.

No checker assertion was weakened, skipped, or rewritten.

---

# Round 2 — the cases the pointer/live split opened

Re-check of this branch @ `2ff2676a`. The checker's round-2 probes are
cherry-picked in as `b101ef847` (untouched); everything below is the
product-code answer. **All five findings are fixed and every one of the
checker's assertions passes.** One conductor ruling is implemented on top,
with its own test.

## 2.1 — a `place()` that dies mid-flight stranded its live row

`place()` writes the live row first and swings the pointer second, so the slot
never reads empty mid-replace. The cost nobody paid for: until the pointer
lands, **nothing names that row** — and anything that stopped the swing left it
there forever. Nothing read it, nothing collected it, and every retry added one
more, against the "exactly one live row per slot" count the seam readers take.
`place()`'s own comment claimed the row "is taken back out before retrying";
no code did that.

**Fix** (`packages/apps/src/placements.ts`) — the pointer swing is its own
function, and a swing that throws takes the unnamed live row back out before the
error reaches the caller. The comment now says what the code does, including the
part that was always true: losing the CAS retries under the *same* token, so the
row already down is the one the next attempt names and there is nothing to
collect.

**Proves it:** `placement-risk-2.test.ts` › "leaves no live row behind for the
retry to pile on top of" ✅

## 2.2 — deleting an app left its pointer behind forever

The split made `delete` clear only the **live** row, so the pointer kept the
dead app's id, its `placedBy` subject and its timestamp. Nothing in the tree
ever removed a pointer: one accumulated per slot a person had ever used, and
only a full subject erase reached it. Before the split, delete removed the
placement outright.

**Fix** — the clear takes the pointer with it, and **only while the pointer
still names the token being cleared**. That token check is what keeps this from
re-opening finding 3 of round 1: a place that took the slot between the read and
the write installs a new token, so its pointer is not ours to remove and the
slot stays with the app that won it.

**Proves it:** `placement-risk-2.test.ts` › "takes its pointer with it, not just
its live row" ✅ — and round 1's `placement-risk.test.ts` › "clears the slot the
winner just took when a place lands inside unplace's read-then-delete" is still
green, which is the case that would have caught a careless pointer delete.

## 2.3 — deleting a SHARED app left a failure card on other people's pages

`runtime.delete` swept `placementRows.list(ctx.principal.subject)` — the
**deleter's** rows. Mia deletes an app Ada had placed on her own page: Ada's row
survives with no record behind it, reads `building` and then, past the build
window, `failed` — and a failed slot **replaces** the host's own markup until a
person taps "Clear this slot". The §9.4 mask added in round 1 does not catch it;
that arm runs only where an app record exists, and here there is none.

**Fix** — placement rows now carry `refs.app_id` (the same ref name
`egress-approval.ts` clears an app by), and `runtime.delete` sweeps by **app**
through a new `placementStore.clearForApp(appId)`. A shared app sits in slots
belonging to people the deleter cannot enumerate; those are exactly the pages
that were being left holding it.

The read path is deliberately unchanged. "No record, past the build window ⇒
`failed`" is what the build watchdog exists to surface, and it cannot tell a
deleted app from a build that never landed — so the honest fix is to stop
leaving the row, not to stop reporting it.

**Proves it:** `placement-risk-2.test.ts` › "clears the rows OTHER people hold
on it, not only the deleter's" ✅

## 2.4 — the write trimmed the slot id and the read did not

`requireSlot` trims on every write; the read path did not trim at all, so
`placements({ slots: [" hero "] })` could not see what `place(" hero ")` wrote.
The wire's own parser trims each name, which is what hid it from every test
until a host called the runtime or the client directly.

**Fix** (`packages/apps/src/runtime.ts`) — the read filter goes through the
**same** `requireSlot` every write does. One normalizer, both sides.

**Proves it:** `placement-risk-2.test.ts` › "answers a `slots` filter that is
spelled exactly as the write was" ✅

## 2.5 — a comma in a slot id was writable and permanently unreadable

Both halves shipped: the client joined mounted slot ids with `,` and
percent-encoded the whole join, and the route split the decoded value on `,`. A
slot id containing a comma therefore came back as two slot names that do not
exist — an app in a slot the page can never resolve, and a person looking at an
empty slot with no way to fix it.

**Fix** — encoding per item, which is what the transport always meant. The
client percent-encodes each id **before** the join (`slotsQuery` in
`packages/ui/src/client-impl.ts`), so the separator can never be an id's own
comma; the route decodes each item after the split (`packages/vendo/src/wire/apps.ts`).
Hand-written URLs are unaffected — text that is not valid percent-encoding
stands for itself, so `?slots=home-hero%2Csidebar` still means two slots.

**Proves it:** `wire/apps.placements.test.ts` › "asks for the slot id the client
was given, comma and all" ✅ (real client → real route, nothing stubbed between
them)

## Conductor ruling — an unattended `vendo_make` with a slot BUILDS

Round 1 put a slot-bearing `vendo_make` into `presenceOnlyCall`, so the guard's
choke point refused the whole call. That is too much: it silently breaks every
automation that legitimately creates a screen. **Placement is what requires a
person present; creation is not.**

**Implemented** — `presenceOnlyCall` (`packages/core/src/grant-sets.ts`) names
the pin tools and only those. A slot-bearing make goes through and the *slot* is
dropped, at the tool's own door in `packages/apps/src/agent-tools.ts` — which is
where round 1 already dropped it. That was the disagreement the checker flagged:
`agent-tools.ts` said "the build still runs and the slot is simply not taken"
while `guard.ts` refused the call outright. Both comments now state the same
rule.

**Proves it:** `placement-tools.e2e.test.ts` › "builds a slot-bearing make on an
unattended run and takes no slot" — a real away firing over the composed host:
the call crosses the same guard-bound registry that refuses the pin tool two
cases above, holding the real captured authority an away run needs (05 §6 —
parked on the first firing, granted by one present tap, spent by the second).
The app comes back `ready` and is in the caller's app list; the store's
placement rows are empty. Verified red on the old rule (`denied`) and green
with the fix.

## Round-2 gate

`pnpm build && pnpm test:affected && pnpm typecheck && pnpm lint` — build ✅,
typecheck ✅ (45/45), lint ✅ (dependency-guard, portability-gate and eslint;
one pre-existing demo-bank warning).

Tests: **no new reds.** The only failures are the pre-existing ones named on the
base tip, each re-run in isolation to confirm it — `core/packaging.e2e` ×3,
`store` durability/split-brain drills ×2, `vendo/dev-creds/model` ×2.
`fixtures/mcp-e2e` also went red under the concurrent turbo run and is green
standalone (19 passed / 5 skipped) — contention, not a regression.

Per-package: `apps` 1078 passed / 19 skipped (was 1073 — the five new cases) ·
`guard` 269 · `ui` 1087 · `vendo` 2091 passed / 33 skipped, 2 known red.

## Round-2 test-file changes, stated out loud

- `packages/apps/src/placement-rows.test.ts` — the refs case now expects
  `app_id` alongside `subject`/`slot` on both rows, because finding 2.3 makes
  that ref the key app deletion sweeps on. A new case pins `clearForApp` across
  two subjects.
- `packages/vendo/src/placement-tools.e2e.test.ts` — the composed host takes an
  optional per-turn action so a test can drive a real CALL through the same
  guard-bound registry it already measures the listing on, plus the ruling's
  case above.

No checker assertion was weakened, skipped, or rewritten in either round.

### DIFF
--- a/.changeset/slots-risk-round.md
+++ b/.changeset/slots-risk-round.md
@@ -3,6 +3,8 @@
 "@vendoai/core": patch
 "@vendoai/guard": patch
 "@vendoai/mcp": patch
+"@vendoai/ui": patch
+"@vendoai/vendo": patch
 ---
 
 Risk-round fixes on the placement stack.
@@ -25,12 +27,40 @@ Risk-round fixes on the placement stack.
   reused, so a stale client's delete can only ever hit its own placement.
   Pointers live in their own generic collection (`vendo_placement_slots`), so
   `vendo_placements` still holds exactly one row per live placement.
-- A `vendo_make` that names a `slot` is presence-only, like the pin tool: the
-  slot claims a place on somebody's page and evicts whatever held it, so an
-  unattended run still builds but takes no slot. Plain makes are not regraded.
+- A `vendo_make` that names a `slot` still builds when nobody is there, and
+  simply takes no slot. Placement is what needs a person present; creation is
+  not, and refusing the whole call would silently break the automations that
+  legitimately create screens. `presenceOnlyCall` now names the pin tools and
+  only those.
 - The presence-only law now has the second layer the destructive half has
   always had. `projectableForRun` only decides what a run is OFFERED; the
   guard's choke point refuses a presence-only call outright, so a standing
   automation grant can no longer rearrange a page with nobody watching.
 - `withholdTools` holds on both legs of an MCP mount. A turn-bearing session
   listed and could call a name the deployment said it never offers.
+
+Second round, on the pointer/live-row split the first round introduced.
+
+- A `place()` that dies between writing its live row and swinging the pointer
+  no longer strands that row. Until the pointer lands nothing names the row, so
+  a failed swing takes it back out; every retry used to leave one more row that
+  nothing read and nothing collected, against the "exactly one live row per
+  slot" count the seam readers take. `place()`'s comment claimed a cleanup no
+  code performed; it now describes what the code does.
+- Deleting an app leaves no placement behind. The clear removed only the live
+  row, so the pointer kept the dead app's id, its `placedBy` and its timestamp
+  forever — nothing in the tree ever removed a pointer, and one accumulated per
+  slot a person ever used. The pointer now goes too, and only while it still
+  names the token being cleared, so a place that took the slot in between keeps
+  it.
+- Deleting a SHARED app clears it out of everyone's slots, not just the
+  deleter's. Placement rows carry `refs.app_id` (the ref `egress-approval.ts`
+  clears an app by) and the sweep is by app, so a co-owner's delete can no
+  longer leave a permanent "didn't build" card standing over another person's
+  host markup.
+- `placements({ slots })` normalizes the slot id the same way every write does,
+  so `place(" hero ")` is findable as `" hero "`. Only the write trimmed before.
+- A slot id containing a "," survives the wire. The client joined slot ids on
+  "," and the route split on ",", so such an id was writable and permanently
+  unreadable. Each id is now percent-encoded on its own inside the list and
+  decoded per item after the split.
--- a/.changeset/slots-risk-round.md
+++ b/.changeset/slots-risk-round.md
@@ -17,9 +17,14 @@ Risk-round fixes on the placement stack.
   "nothing was replaced" while one of them was silently displaced. The write
   now compare-and-swaps on the row's revision and the loser retries against the
   winner's row.
-- `unplace()`'s delete is scoped to the app the caller named, at the store, so a
-  stale client cannot clear the app that replaced it. Adapters that expose
-  `RecordStore.claim` compare and delete in one statement.
+- `unplace()` can no longer clear the app that replaced it. A placement is now
+  two rows: a pointer at `plc:<subject>:<slot>` that says who holds the slot
+  under which token (the single CAS arbitration point), and a live row at
+  `plcv:<subject>:<slot>:<token>` that exists only while that placement holds
+  it. A clear deletes the token'd row and nothing else, and tokens are never
+  reused, so a stale client's delete can only ever hit its own placement.
+  Pointers live in their own generic collection (`vendo_placement_slots`), so
+  `vendo_placements` still holds exactly one row per live placement.
 - A `vendo_make` that names a `slot` is presence-only, like the pin tool: the
   slot claims a place on somebody's page and evicts whatever held it, so an
   unattended run still builds but takes no slot. Plain makes are not regraded.
--- a/.changeset/slots-risk-round.md
+++ b/.changeset/slots-risk-round.md
@@ -1,0 +1,31 @@
+---
+"@vendoai/apps": patch
+"@vendoai/core": patch
+"@vendoai/guard": patch
+"@vendoai/mcp": patch
+---
+
+Risk-round fixes on the placement stack.
+
+- `GET /apps/placements` is an app read. `entryFor` raw-read the app row, so a
+  viewer whose grant was taken back kept reading a placed app's title and its
+  live build status after `open`/`get`/`list` had all gone back to not-found.
+  Every entry now passes the same `can()` check those paths use, and a slot the
+  caller may no longer view reads as empty.
+- `place()` is one decision, not read-then-write. The eviction receipt used to
+  be a separate read, so two places into the same slot could both answer
+  "nothing was replaced" while one of them was silently displaced. The write
+  now compare-and-swaps on the row's revision and the loser retries against the
+  winner's row.
+- `unplace()`'s delete is scoped to the app the caller named, at the store, so a
+  stale client cannot clear the app that replaced it. Adapters that expose
+  `RecordStore.claim` compare and delete in one statement.
+- A `vendo_make` that names a `slot` is presence-only, like the pin tool: the
+  slot claims a place on somebody's page and evicts whatever held it, so an
+  unattended run still builds but takes no slot. Plain makes are not regraded.
+- The presence-only law now has the second layer the destructive half has
+  always had. `projectableForRun` only decides what a run is OFFERED; the
+  guard's choke point refuses a presence-only call outright, so a standing
+  automation grant can no longer rearrange a page with nobody watching.
+- `withholdTools` holds on both legs of an MCP mount. A turn-bearing session
+  listed and could call a name the deployment said it never offers.
--- a/packages/apps/src/agent-tools.ts
+++ b/packages/apps/src/agent-tools.ts
@@ -4,6 +4,7 @@ import {
   VENDO_MAKE_TOOL,
   VENDO_TOOL_TITLES,
   appIdSchema,
+  isUnattended,
   VENDO_VIEW_STREAM,
   VendoError,
   makeReceiptSchema,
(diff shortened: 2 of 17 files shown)


yousefh409:

PR #917, opened 2026-08-06
polish(ui): one busy vocabulary, one component family, one theme mapping — and the agent asks to connect

yousefh409:
## What this is

The Vendo agent surface read like several products stitched together: two visual languages for "in progress", two component families racing to render the same tag, the theme→CSS-variable mapping hand-written four times, and a connect flow that only appeared *after* the agent had already failed in front of the user. This deletes the duplicates, gives the theme one owner, makes the Kit the only component family, and rebuilds the connect moment so the agent asks before it acts.

Net: **+2725 / −2779** across 140 files — the project removes more than it adds.

## What ships

**Polish (what a user sees)**
- Maple's "Or try this" pill tier is gone, and so is the SDK mode that stacked pills under cards with a hardcoded label. One suggestion tier; hosts pass cards *or* pills.
- One busy vocabulary: three bouncing dots while thinking, checklist beats while tools run. The gray "Working…" pill above the composer is deleted — the between-steps gap is now a beat line at the transcript tail, and it only appears when the turn has real tool activity and the silence outlasts 800ms (it used to flash "Working… 0.5s" under a finished answer).
- The fabricated `Connected Gmail.` bubble no longer appears in the user's own voice; the resume travels as hidden agent context.

**Connect asks first (new capability)**
- New agent tool `request_connection(toolkit, reason)` returns the *existing* `connect-required` outcome, so the wire and the render path are unchanged — the agent can now offer a connection in prose instead of hitting a wall first. Declining collapses the card to a "Skipped" row that still offers Connect, and the agent adapts.
- OAuth opens in a centered popup created **synchronously in the click handler** (browsers judge popups by call-stack provenance; the old code awaited the broker first, which is exactly what gets blocked) and the opener closes it on success. Blocked-popup and timeout states are designed, not accidental.
- 🔒 **Security fix:** the broker's `redirectUrl` was navigated with no scheme validation. The popup is opened without `noopener` by design (so we can close it), so it inherits our origin — a `javascript:` URL would have executed in our own document with `opener` reachable. Now rejected at one choke point before any navigation, surfacing as the ordinary "nothing changed, try again" refusal.

**Centralization (why the drift stops)**
- One theme→CSS-variables mapping, in `@vendoai/core`. `ui` re-exports it (host API unchanged); both hand-copies in `@vendoai/mcp` become thin serializers and now emit the **full** variable set — density and motion were silently dropped inside the MCP door before this. A seam test pins all three contexts to an identical variable set.
- Two live bugs fixed on the way: the Kit's fallback values had `surface` and `background` **swapped** (unthemed hosts rendered inverted), and phantom `--vendo-space-*` variables that nothing anywhere produced are gone.
- The legacy component family (`tree/primitives.tsx`, `tree/branded.tsx`) is retired; the Kit is the only family and `KIT_SPECS` the only name source. The Kit gains `Card`; `Table` is not rebuilt (`DataTable` is a strict superset and is taught instead); `Skeleton` becomes private chrome. The renderer resolved the *legacy* component for 15 shared names while generated island code resolved the *Kit* one — same tag, two widgets, gone.

## ⚠️ Scope event

The approved plan dropped `Tabs` from the generated-UI vocabulary, on the premise that nothing produced wire tabs. **That premise was false:** `packages/apps/src/generation/skeleton.ts` emits a `Tabs` node whose children are every panel of the app, so following the plan as written would have rendered every multi-tab generated app as a bare tab bar with no body. Caught before any code was written. `Tabs` was kept and rebuilt on the Kit with the legacy children-as-panels contract, and `test/tree/tabs-skeleton-seam.test.tsx` now renders the generator's exact node shape through the real renderer so the two can never drift again.

## Proof

Five slices, each gated in its own worktree, then one internal risk round (6 real findings, all fixed and re-verified clean) and a cold end-to-end walk by a prover that built none of it.

- Live, on the real Vendo Cloud account: the connect card offered by `request_connection` in **5 of 5** fresh conversations (provenance read back off the persisted thread through the real door — not merely "a card appeared"), popup opening in 157ms on the genuine Google consent page, decline collapsing correctly with the agent adapting, and the blocked-popup fallback completing.
- The first cold walk returned NOT-SHIP on two counts, both fixed here: the offer fired only 2/6 because the new teaching sat in a prompt section the engine path never receives while another section it *does* receive told the model to point users at a button instead; and a hidden context message could become a persisted thread title (the marker lived only in `ui`, while titles are minted in `agent`).
- Screenshots attached below.

## Not in this PR — pre-existing `main` bugs the walk found

Reproduced on `origin/main`, deliberately untouched here:
1. **"New chat" is dead once you've spoken in the current thread** — the next message silently appends to the old thread. A thread minted by *sending* lands in `minted`, but New chat clears `selected` (`vendo-page.tsx:285`), so the key never flips, `VendoThread` never remounts, and `activeThreadIdRef` keeps the old id. Most common path in the product; worth its own fix.
2. Accounts panel: each connect attempt leaves a duplicate "Gmail — Expired" row, and closing the popup leaves "Reconnecting…" stuck for the full 120s poll (nothing watches `popup.closed`).
3. React unique-key warning on every thread-rail render (`CenterChats`).
4. The connect card vanishes on thread reload, leaving prose that references a button that is gone.

Also observed, generation-engine quality rather than this project: a generated tab body shipping the text "This section could not be built", raw cents (`643178`) rendering beside correct currency, tables clipped mid-value with no scroll affordance, and a chart claiming "no dining spend" above eight dining transactions.

## Test-change note

`test/chrome/affordances-eng225.test.tsx` was retargeted, stated out loud: it pinned the retired one-shot `window.open(redirectUrl)` mechanism by stubbing `window.open` to `null`, which under the new design *means* "popup blocked". Its intent ("the broker flow reaches the real OAuth URL") is now asserted on the two-phase mechanism. A 51-export / 81-token sweep across the test tree found no other file pinning a retired mechanism.

---
## Summary by cubic
Unifies the UI and agent: the Kit is the only component family, theme→CSS variables map once in `@vendoai/core`, and the agent asks to connect before it acts via a centered, secure popup. Rebased to `main` with agent work under `@vendoai/vendo`; connector discovery tools are always active and the ask now triggers reliably.

- **New Features**
  - The agent learns before it asks: when an outside service is needed, it calls `list_connections` then uses `request_connection`; connector discovery tools are always-active (not loadout-gated) so the ask always appears.
  - Connect popup opens synchronously in the click, centers, navigates on redirect, and closes from the opener; blocked-popup and timeout states are handled.
  - One component family: the Kit only. Legacy prewired/branded components are retired. `Card` added; `DataTable` is the table surface; `Tabs` kept with the child-panels seam pinned.
  - One theme mapping in `@vendoai/core` (`defaultVendoTheme`, `resolveTheme`, `colorSchemeForBackground`, `themeCssVariables`); `@vendoai/ui` re-exports; `@vendoai/mcp` and the shim serialize the same set.
  - Simplified busy moments: three dots while thinking; a `WorkingBeat` line appears only for real tool gaps (>800ms). Demo “try this” pills and their machinery are removed.

- **Bug Fixes**
  - Security: reject non-HTTP(S) `redirectUrl` before navigation to prevent `javascript:` executing in our origin.
  - Hidden agent context stays hidden: no thread titles from it, agent-context-only user messages render without a bubble, and snippets exclude it; the connect continuation rides agent context.
  - Fixed theme fallback inversion (surface/background), removed phantom `--vendo-space-*`, and ensured density/motion reach MCP pages.
  - Engine path now teaches and triggers the ask; prompt truth aligned with behavior. Validation allows `$path` literal bindings in typed Kit props; island ambient/reserved-name checks read `KIT_COMPONENT_NAMES`; references to `Table` updated to `DataTable`.
  - Rebase to `main`: agent paths moved to `@vendoai/vendo`; generated demo manifests synced, `host_demo_chips_list` removed, and the `host_demo_reset_create` srcHash corrected.

<sup>Written for commit 38fef3f0271a0f387afae582bc1fde79f8218791. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/917?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
## Browser proof inventory

Every UI-affecting change was walked in a real browser against the live Vendo Cloud account. Images are on the build machine at the paths below (GitHub's attachment endpoint needs a browser session, so they cannot be pushed from CLI — happy to drag them in, or re-run any walk on request).

**Connect asks first — 5 of 5 fresh conversations, provenance verified**
Each trial: new conversation, Gmail disconnected, prompt "draft me an email summarising this week's spending". A card appearing is not the claim — each was confirmed to originate from a `request_connection` tool part read back off the persisted thread through the real door.
- `/tmp/fixproof-offer-1.png` … `/tmp/fixproof-offer-5.png` — the five trials
- `/tmp/v5-proof-card.png` — the card itself: plain-language access line ("Connecting lets us read and send mail as you."), Connect + Not now
- `/tmp/v5-proof-popup.png` — centered popup, real `accounts.google.com` → "Sign in to continue to Composio", opened 157ms after click
- `/tmp/v5-proof-skipped.png` — decline collapses to a one-line "Skipped — Gmail isn't connected" that still offers Connect; agent adapts in prose
- `/tmp/v5-proof-blocked.png` — popups blocked: the "Open sign-in in a new tab" fallback renders and the poll keeps running
- `/tmp/fixproof-titles.png` — every thread created in the run has a clean title; no `[vendo:context]` leak

**One busy vocabulary**
- `/tmp/fresh-proof-2-t1600.png` — three bouncing dots before any text (no metaball)
- `/tmp/fresh-proof-2-t18000.png` — checklist beats inside the transcript, `fl-beat-working` → `fl-beat-done`; **no pill above the composer at any sample**
- `/tmp/fresh-proof-2-t90000.png` — after settle: "✓ Did 2 things · 65.0s", no lingering "Working…"
- `/tmp/fresh-proof-1b-launcher.png`, `/tmp/fresh-proof-1c-fullpage.png` — landing on both mounts: greeting + starter cards, no chip row, no "Or try this"
- `/tmp/v2-proof-mixed.png` — a mixed suggestions array renders as ONE unlabelled chip row (`chipRows:1, tryRow:0, tryLabel:0`, `chipsParent: fl-landing`)

**One component family**
- `/tmp/v4-proof-generated.png` — generated app end-to-end on live data; `data-kit` on every built-in and `document.querySelectorAll('[data-primitive]').length === 0` page-wide
- `/tmp/v4-proof-skeleton.png` — the forming skeleton still paints mid-stream after `Skeleton` went private (4 live `[data-skeleton]` captured at t=20.4s)

**One theme mapping**
- `/tmp/v3-proof-maple.png` — Maple visually unchanged; `.vendo-root` carries 32 variables
- `/tmp/v3-proof-unthemed.png` — the swap bug fixed: an unthemed Kit renders white page / `#f7f7f8` surfaces instead of inverted
- `/tmp/v3-proof-mcp-compact.png` + `/tmp/v3-proof-mcp-comfortable.png` — the MCP connect page now responds to density (10px comfortable → 7px compact) where it previously ignored it

**Known and unrelated:** one console warning appears throughout — React's unique-key warning from `CenterChats` on every thread-rail render. Reproduced on `origin/main`; listed in the PR body as pre-existing.

PR #919, opened 2026-08-06
fix(apps): a slot-targeted make claims its slot at mint on the assembly path too

yousefh409:
## The B1 gap

Decision **B1** of the approved design: a slot-targeted `vendo_make` claims its
slot the moment the appId is minted, so the slot shows the build immediately, and
a terminal failure shows the honest failure card **in the slot** (dismiss =
unplace).

The cold walk found this held on the **escalated builder** path only. On the
**fast screen-assembly** path — the engine that serves most asks — the placement
row was written *after* assembly succeeded. Two consequences, both observed in
the real product:

- **(a)** an empty slot showed nothing for the whole of a fast make, instead of
  the skeleton the slot exists to show;
- **(b)** a **failed** assembly wrote **no row at all**, so the slot showed
  nothing and the person learned of the failure only in the conversation.

## The fix

The claim now rides the **mint**, at the one place that mints the id — the
`vendo_make` front door, before it routes — so both engines are claimed for from
the same instant.

- `runtime.ts`: the mint-time claim is extracted into one internal `claimSlot`
  write. `create` keeps its own `slot` (for callers that mint inside it) and now
  goes through that same write; the front door no longer passes `slot` down,
  because it already claimed.
- `agent-tools.ts`: `claimSlot` fires right after the mint, before the assembler
  runs. The two terminal-failure exits (assembly came back empty; escalation with
  no sandbox) now land the **same terminal record a failed build leaves**, via
  `markUnbuilt`. `entryFor` therefore reads one thing and the claimed slot turns
  into the honest failure card the instant either engine gives up, rather than
  holding a skeleton until the 5-minute build window ages out.

Both new capabilities are internal wiring on `AgentToolsDataDependencies` (the
channel the runtime already uses to construct its own tool registry) — **no
public `AppsRuntime` surface added**.

Round-2 properties preserved, deliberately:

- the read path still cannot distinguish "app deleted" from "never landed" — the
  tombstone is a **write-side** answer, and app deletion still sweeps rows by
  `refs.app_id` through `clearForApp`, so a placement whose app no longer exists
  still renders as nothing placed;
- a placement whose app exists and FAILED renders the failure card — that is the
  path this PR extends to the assembly engine;
- the tombstone is a tombstone: `list()` already skips `buildFailed` documents,
  so it answers the slot and never joins the user's apps (asserted).

No assertion was weakened or removed.

## Proof 1 — unit/integration (`packages/apps/src/agent-tools.test.ts`)

Two new tests, both reading back through the runtime's real `placements()` path:

- **claims at MINT** — the reading is taken *from inside the assembler*,
  mid-flight: `{ slot, app, title: "", status: "building" }`, then the same row
  goes `ready` when assembly lands.
- **terminal failure** — an assembler that answers `unavailable` leaves
  `status: "failed"` in the slot with the receipt's own title, the app is absent
  from `list()`, and `unplace` (dismiss) clears it.

**Revert-and-watch, both directions:**

| reverted | result |
| --- | --- |
| drop the mint-time `claimSlot` | 4 red, incl. `expected [] to deeply equal [{ slot: 'dashboard.hero', … }]` — the slot is empty through the make |
| drop `markUnbuilt` | 1 red: `- "status": "failed"` / `+ "status": "building"` — the stuck skeleton this fixes |

Restored, 29/29 green.

## Proof 2 — seam (`packages/vendo/src/placement-tools.e2e.test.ts`)

Extended, nothing stubbed on either side. The failure is **real, not injected**:
this host's screen agent can serve one ask, so the second `vendo_make` reaches the
same real MCP door (register → authorize → token → JSON-RPC) and comes back with
nothing that renders.

- the write goes over the real MCP door;
- the row is read out of the real store's `vendo_placements`;
- the **status** is read back over the real `GET /apps/placements` route a browser
  polls — `status: "failed"`, title matching the receipt.

The ready direction now also reads through that same route in the existing
first test (added assertion, nothing changed).

**Revert-and-watch:** with `markUnbuilt` dropped, the seam test goes red with
`- "status": "failed" / + "status": "building"` and `- title: "match my invoices
to payments" / + title: ""`. Restored, 6/6 green.

## Gate

`pnpm build` ✅ · `pnpm test:affected` ✅ (only the known pre-existing reds) ·
`pnpm typecheck` ✅ · `pnpm lint` ✅

Known pre-existing reds, unchanged by this PR: `core/packaging.e2e` ×3,
`vendo/dev-creds/model` ×2, `store` drills ×2. `fixtures/mcp-e2e` ran clean.

Changeset: `.changeset/slot-claim-at-mint.md` (`@vendoai/apps` patch).

---
## Summary by cubic
Fixes slot-targeted `vendo_make` so it claims the slot at mint on both assembly and builder paths. Slots now show the build immediately, and terminal assembly failures render as failed in the slot.

- **Bug Fixes**
  - Claim happens at the `vendo_make` front door right after id mint; `create` uses the same internal `claimSlot`.
  - Added `markUnbuilt` so unlanded asks leave the same tombstone as a failed build; the slot shows failed and `list()` excludes it.
  - Wired via `AgentToolsDataDependencies`; no public `AppsRuntime` changes in `@vendoai/apps`.
  - Tests updated, including e2e verifying status via real `GET /apps/placements`.

<sup>Written for commit c81126c24e3ef831b35ddd3cbe3fd4c3a9f418ca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/919?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/slot-claim-at-mint.md
+++ b/.changeset/slot-claim-at-mint.md
@@ -1,0 +1,19 @@
+---
+"@vendoai/apps": patch
+---
+
+A slot-targeted `vendo_make` claims its slot at MINT on the assembly engine too,
+not only on the escalated builder.
+
+The row used to go down only after assembly RETURNED, so on the engine that
+serves most asks a slot the person was already looking at stayed empty for the
+whole of the make — the skeleton the slot exists to show never appeared — and an
+assembly that failed wrote no row at all, so the slot showed nothing and the
+failure existed only in the conversation.
+
+Both engines are now claimed for from the one place that mints the id: the
+`vendo_make` front door, before it routes. And an ask no engine landed leaves the
+same terminal record a failed build leaves, so the claimed slot turns into the
+honest failure card (dismiss = unplace) the instant assembly gives up, rather
+than holding a skeleton until the build window ages out. That record is a
+tombstone like any other failed build's — it never joins the user's apps.
--- a/packages/apps/src/agent-tools.test.ts
+++ b/packages/apps/src/agent-tools.test.ts
@@ -4,11 +4,12 @@ import {
   VENDO_TOOL_TITLES,
   toolDescriptorSchema,
   type RunContext,
+  type ScreenAssembler,
   type ToolRegistry,
 } from "@vendoai/core";
 import { describe, expect, it } from "vitest";
 import { agentToolDescriptors } from "./agent-tools.js";
-import { createApps, type AppsRuntime } from "./index.js";
+import { createApps, type AppsRuntime, type PlacementEntry } from "./index.js";
 import {
   authoringAssembler,
   basicLanguageModel,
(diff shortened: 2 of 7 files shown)


yousefh409:

Issue #874, opened 2026-08-06
doctor act-as pass message claims 'host verification' but the probe only round-trips vendo's own resolver


amrib247:
﻿## Summary
The doctor act-as check prints **"actAs mint + host verification live round-trip"** on pass — but the probe never touches the host's session verification. It mints with vendo's own preset and verifies at vendo's own `/doctor/act-as/echo` via vendo's own `deps.principal`. Mint and verify share the same preset, secret, and hoisted `@auth/core`, so the check is a self-round-trip. Consequence: on hosts whose real middleware cannot verify the minted material (e.g. next-auth v4), doctor reports green while every real away-mode call (automations, MCP door) will fail host auth.

## Affected
vendoai 0.7.0; still present at main `7313b873`.

## Evidence
- `packages/vendo/src/server.ts:1952-1975` — synthetic away ctx, vendo-bound tool
- `packages/vendo/src/wire/doctor.ts:23-29` — echo bound to vendo's own route; `:93-98` — verified via `deps.principal`
- `packages/vendo/src/cli/doctor.ts:628` — the "host verification" pass message

## Suggested fix
Either exercise a genuinely host-authenticated endpoint with the minted material (e.g. one extracted read tool), or rename the check/message so it doesn't claim host verification it never performs.

amrib247:
﻿Fix implemented (PR to follow after local verification). The pass message no longer claims host verification: it now reads "actAs mint round-trip verified by the composition's own principal resolver — host middleware is not exercised, real away calls still depend on it." The E-AUTH-004 docs entry likewise states the probe's vendo-side scope. The deeper redesign (probing a genuinely host-authenticated endpoint with the minted material) is tracked as follow-up design work rather than this fix.

yousefh409:
Closing as moot — #1385 retired doctor's live actAs probing entirely; doctor is static-only now.

Issue #873, opened 2026-08-06
doctor: declined actAs mint indistinguishable from unconfigured seam (both E-AUTH-007)


amrib247:
﻿## Summary
`vendo doctor`'s act-as probe collapses two very different states into one answer: a mint that **declined** (e.g. `authJs()` wired but secret unset, or a user-resolver that rejects the synthetic doctor subject) and an actAs seam that is **absent** both map to `not-implemented` → E-AUTH-007 "actAs is not configured; pass createVendo({ actAs })". A correctly wired host gets told to configure what it already configured. In the same family: on E-AUTH-004 the wire replaces the registry's actionable error (which can include an install instruction) with a generic string.

## Affected
vendoai 0.7.0; still present at main `7313b873`.

## Evidence
- `packages/actions/src/runtime/registry.ts:394` (declined, disposition `actAs:"declined"`) vs `:624` (absent) — same error code, different disposition + message
- `packages/vendo/src/wire/doctor.ts:113` — branches on `code === "not-implemented"` alone, discarding both discriminators
- `wire/doctor.ts:122-128` — clobbers the actionable message (cf. `registry.ts:397`)

## Suggested fix
Branch on the actAs disposition: keep E-AUTH-007 for the truly-absent seam, add a distinct code/message for "configured but declined the doctor principal" and pass the registry's real message through on E-AUTH-004.

amrib247:
﻿Fix implemented (PR to follow after local verification). The act-as probe executor calls the actions registry directly (no guard binding), so the actAs disposition passthrough survives on the outcome — the wire now branches on it: a configured seam that DECLINED answers `act-as-declined` (409, carrying the registry's own reason) while a truly absent seam keeps `act-as-not-configured` (501). Doctor maps declined to a new warning code **E-AUTH-008** ("actAs is configured and declined the doctor's synthetic principal — expected when a subject→user resolver only mints for real users") instead of telling a wired host to wire actAs, and E-AUTH-004 now carries the wire's failure reason instead of a generic line. Documented in agents/verify.

yousefh409:
Fixed at the wire level (tests named for #873); the CLI-side probe was retired by #1385.

Issue #872, opened 2026-08-06
Principal-resolver failures return generic 501; authJs resolves secret/module before cookie inspection


amrib247:
﻿## Summary
Two related principal-resolution failure modes produce terrible DX:
1. When the authJs preset cannot initialize (missing `@auth/core`, missing secret), the throw happens **before any cookie inspection** — so every request including anonymous ones fails, and the wire's catch-all converts it to a generic **501 `Internal Vendo error`** with the actionable message demoted to the server console. Observed as 973× 501s on `GET /api/vendo/connections` with nothing useful in the browser.
2. Once initialized on a next-auth v4 host, the v4 session cookie (`next-auth.session-token`) is simply never found (`getToken` reads `authjs.session-token` only) — signed-in users silently resolve as anonymous, so per-user features reject them (`403 connecting external accounts requires a signed-in user; sign in first` — while visibly signed in).

## Affected
vendoai 0.7.0; still present at main `7313b873`.

## Evidence
- `packages/vendo/src/auth-presets/auth-js.ts:83-88` — secret/module resolution before cookie inspection
- `packages/vendo/src/wire/context.ts:141` — no try/catch around the resolver
- `packages/vendo/src/server.ts:1531-1536` — catch-all → `internalError()` → 501, real message to console only
- `packages/vendo/src/connections.ts:53-56` — ephemeral principal → blocked

## Suggested fix
(a) Catch principal-resolver errors and answer with the resolver's actionable message as a 401/503 auth-config error instead of generic 501. (b) Defer secret/module resolution until a session cookie is actually present so anonymous traffic keeps working. (c) Consider detecting a v4-named cookie and logging a one-time, loud "v4 session cookie seen — authJs() is v5-only" hint.

amrib247:
﻿Fix implemented (PR to follow after local verification). Two halves:
1. The authJs preset now short-circuits to "no session" before any module/secret work when no Auth.js session cookie is present — a misconfigured preset can no longer 501 anonymous traffic. Seeing a v4-named cookie (`next-auth.session-token`) logs one console hint naming the v4 posture.
2. The wire's context resolver wraps principal-resolver throws into a `VendoError` carrying the resolver's own actionable message (`principal resolution failed: …`) instead of the catch-all's generic `Internal Vendo error`; resolver-thrown VendoErrors pass through unchanged.

yousefh409:
[...]

Issue #871, opened 2026-08-06
vendo init wires authJs() on next-auth v4 hosts with no version check or warning


amrib247:
﻿## Summary
`vendo init` auth detection maps **any** `next-auth` / `@auth/*` dependency to the `authJs()` preset without reading the version. The preset is documented v5-only ("Auth.js and NextAuth v5", docs/act-as-presets.md) — but a next-auth **v4** host gets wired silently (`--yes`) or with a confirm prompt that never mentions v5-only. The install then passes `vendo doctor` while sessions, actAs, and connections are all broken for real users (v4 cookie name + crypto are incompatible).

## Affected
vendoai 0.7.0; still present at main `7313b873`.

## Evidence
- `packages/vendo/src/cli/init-auth.ts:38` — `dependency === "next-auth" || dependency.startsWith("@auth/")`
- `init-auth.ts:54` — `Object.keys` only; version strings discarded
- `init-auth.ts:167-174` — silent wiring non-interactive; interactive prompt has no version caveat

## Suggested fix
Parse the manifest's next-auth version range; on v4 either refuse to wire `authJs()` or wire with a loud advisory naming the `NEXTAUTH_SECRET` / `next-auth.session-token` / KDF incompatibilities and what will break.

amrib247:
﻿Fix implemented (PR to follow after local verification). `vendo init` now reads the matched next-auth version range from the manifest: a major-4 host still wires `authJs()` (the composition is correct for a future v5 upgrade), but every wiring path — silent/non-interactive, `--auth authJs`, and the interactive confirm (whose question now names v4) — surfaces an advisory stating that v4 sessions resolve as anonymous, away runs cannot be verified by the host, and `--auth none` is the stay-anonymous alternative. Unparseable ranges (`workspace:*`, tags) get no advisory rather than a wrong one.

yousefh409:
Fixed — init parses the installed next-auth version and attaches an explicit advisory naming what breaks, on both wiring paths.

PR #925, opened 2026-08-06
fix: next-auth v4 hosts fail loud and correct (authJs secret order, lazy sessions, init advisory, honest doctor)

### LINKED ISSUE #870: authJs preset ignores NEXTAUTH_SECRET - next-auth v4 hosts fail despite having a secret
﻿## Summary
The `authJs()` host-identity preset resolves its secret from `AUTH_SECRET` only. next-auth **v4** hosts (a very large population) carry `NEXTAUTH_SECRET` instead — on those hosts every wire request that resolves a principal throws `authJs() has no session secret: set AUTH_SECRET…` even though the host has a perfectly good secret configured.

## Affected
vendoai 0.7.0; still present at main `7313b873`. Found on a real-host install eval (Linkwarden: next-auth v4, `@auth/prisma-adapter`).

## Evidence
- `packages/vendo/src/auth-presets/auth-js.ts:86` — `resolvePresetSecret(secret, "AUTH_SECRET", …)`
- `packages/vendo/src/auth-presets/identity.ts:58-72` — single env name, throws
- `packages/actions/src/presets/auth-js.ts:61` + `packages/actions/src/presets/shared.ts:53-63` — mint half reads only `AUTH_SECRET` (declines to null)
- `grep NEXTAUTH_SECRET packages/` → zero hits

## Suggested fix
Accept `NEXTAUTH_SECRET` as a fallback env name — or, if v4 stays unsupported (the preset mints v5-format JWEs regardless, see the companion issues), detect it and fail with a named, actionable error: "found NEXTAUTH_SECRET — that is next-auth v4, which authJs() does not support."

amrib247:
﻿## What

next-auth v4 hosts currently break silently: authJs() only reads AUTH_SECRET, init wires the v5-only preset onto v4 hosts without a word, a misconfigured preset 501s every request including anonymous ones, and doctor reports a green "host verification" that never touched the host. Found on a real-host install eval (Linkwarden: yarn-4 monorepo, next-auth v4). Fixes #870, #871, #872, #873, #874.

This PR makes v4 fail loud and correct (it does NOT add v4 session support — that design decision is tracked separately):

- **authJs secret**: zero-config resolves `AUTH_SECRET` then `NEXTAUTH_SECRET` (Auth.js's own legacy order), both halves; an explicit secret source that declines never falls through.
- **Lazy-until-cookie sessions**: no Auth.js session cookie → null before any module/secret work, so a misconfigured preset can no longer take down anonymous traffic; a v4-named cookie logs one hint naming the posture.
- **init advisory**: a next-auth major-4 host still wires authJs() but every path (silent, `--auth`, confirm) prints what v4 cannot do and the `--auth none` alternative.
- **Wire**: principal-resolver throws surface their own actionable message instead of a generic `Internal Vendo error` 501.
- **Doctor**: new `act-as-declined` wire answer + `E-AUTH-008` warning distinguish a configured-but-declining seam from an absent one; `E-AUTH-004` carries the wire's failure reason; the act-as pass message states its honest scope (composition-side verification only).

## How

Six commits, one concern each, tests alongside (TDD). Key files: `packages/actions/src/presets/auth-js.ts`, `packages/vendo/src/auth-presets/auth-js.ts`, `packages/vendo/src/cli/init-auth.ts`, `packages/vendo/src/wire/context.ts`, `packages/vendo/src/wire/doctor.ts`, `packages/vendo/src/cli/doctor{,-codes}.ts`. The declined/absent discriminator rides the existing actAs disposition passthrough (the probe executor bypasses guard.bind, so no core contract changes). Docs updated in-change: `docs/act-as-presets.md`, `docs-site/connect/act-as-presets.mdx`, `docs-site/agents/{verify,host-auth}.mdx`. Patch changeset included.

## Tests & review

New/updated suites green: `auth-presets/auth-js` (28), `presets/auth-js` (8), `cli/init-auth` (8, new), `wire/context.principal-errors` (3, new), `cli/doctor` + `doctor-codes` (81), server act-as branches (2 new). Typecheck, dependency-guard, and the workerd portability gate green locally. Field-verified on the originating v4 host: the 973-request 501 storm becomes clean 200s, the one-time v4 hint fires, doctor honest end-to-end. Local Windows caveat: a set of pre-existing environment test failures (POSIX file modes, symlinks, CRLF-sensitive byte comparisons) reproduce identically on unmodified main — CI is authoritative.

## Pending

None blocking. Follow-up design question tracked on the issues: a real v4 mint mode vs a documented hard v5-only gate (#870 discussion).

### DIFF
--- a/.changeset/authjs-v4-fail-loud.md
+++ b/.changeset/authjs-v4-fail-loud.md
@@ -1,0 +1,16 @@
+---
+"@vendoai/actions": patch
+"@vendoai/vendo": patch
+---
+
+next-auth v4 hosts fail loud and correct instead of silently breaking: the
+authJs preset resolves `AUTH_SECRET` then v4's `NEXTAUTH_SECRET` (Auth.js's
+own legacy order), defers module/secret work until an Auth.js session cookie
+is actually present (a misconfigured preset no longer 501s anonymous
+traffic), and names next-auth v4 once when it sees a v4 session cookie.
+`vendo init` prints a v4 advisory when wiring authJs onto a major-4 host.
+The wire surfaces a failed principal resolver's own message instead of a
+generic Internal Vendo error. Doctor distinguishes a declined actAs mint
+(new E-AUTH-008 warning) from an unconfigured seam (E-AUTH-007), passes the
+wire's failure reason through E-AUTH-004, and its act-as pass message stops
+claiming host verification the probe never performs.
--- a/docs-site/agents/host-auth.mdx
+++ b/docs-site/agents/host-auth.mdx
@@ -17,7 +17,7 @@ Init detects from `package.json` (dependencies and devDependencies):
 
 | Dependency matches | Preset | Runtime SDK the preset lazy-loads |
 | --- | --- | --- |
-| `next-auth` or `@auth/*` | `authJs` | `@auth/core` |
+| `next-auth` or `@auth/*` | `authJs` (v5 — a next-auth major-4 host is wired with a printed advisory: v4 sessions resolve as anonymous) | `@auth/core` |
 | `@clerk/*` | `clerk` | `@clerk/backend` |
 | `@supabase/*` | `supabase` | `jose` |
 | `@auth0/*` | `auth0` | `jose` |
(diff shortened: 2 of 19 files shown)


amrib247:
Closing temporarily to batch this with the remaining baseline fix packs into one combined PR (one CI run + review gate instead of six) — the branch stays; the fixes here are field-verified and will return as part of that combined PR shortly.

PR #1010, opened 2026-08-07
fix(store): undo holds the PATH, not the row — the last window in workspace undo

yousefh409:
## The window #995 left open

#995 closed undo's check-then-act window by holding the paths' file rows with
`SELECT … FOR UPDATE` before either leg reads the ledger. **A row lock cannot
lock a row that is not there** — and both undo legs act at a path whose live
row is absent:

- **`created` → `txRows.remove`.** A commit that CREATED a path, undone after a
  **non-ledger** delete (`rm`, `erase`, `moveApp`) took the row away — a
  deleting *commit* is caught by `pathsMovedOn`, these are not. `remove` then
  deletes whatever a racer put back. History is appended before the delete, so
  the content is recoverable: this is a lost live write plus a ledger that
  keeps a commit for a file that is gone.
- **tombstone → `landRow(recordHistory: false)`.** The restore loses its
  `ON CONFLICT DO NOTHING` insert to a racer's create, re-aims at that head and
  swaps it away with no history row behind it. **The content is destroyed.**

And the other writer family: the **SQL façade** — the local/BYO-Postgres
deployment path — ran its commit and its undo with **no transaction at all**,
so an undo landed inside `remove`'s read-then-delete and the delete removed the
revision the undo had just restored, working from the value it read before.
History is append-only (§3.3); that revision is simply gone.

## The change

- **The hold is now on the PATH, not the row**:
  `pg_advisory_xact_lock(hashtextextended(owner || path))`, keyed on the
  identity rather than on a tuple, so it exists before the row does. No DDL, no
  migration. It lives in `workspaceRows` (`land`, `remove`, `undo`) — where the
  ops door and the SQL façade converge — and `ops.workspace.undo` takes the
  same key before its check, so the check and the act are one.
- **The façade's commit and undo are wrapped in a transaction.** The lock is
  transaction-scoped, so outside one it is released at the end of its own
  statement and holds nothing across a pair. `WorkspaceStoreFs.commit` now
  lands its removes and its writes inside one transaction; content is still
  placed before it opens (the saga's one non-transactional leg), and the turn's
  index is applied only after it commits, so a rollback cannot leave the turn
  claiming rows that are gone. `workspaceStore.undo` runs its whole walk in one.
  `WorkspaceRows.transact` is how a caller asks for that.
- **Every hold a transaction needs is taken up front, in one order.**
  `land` and `remove` each take their own path's hold, so a transaction
  touching several paths used to acquire them as its loops reached them —
  removals in the order they were dropped, writes in the order they were
  staged. `WorkspaceRows.hold` takes the whole set at once, sorted by the lock
  key; the commit path hands it the union of its removals and its writes before
  either loop runs, and `ops.workspace.undo` uses the same method rather than a
  second copy of the sort. See **A deadlock this branch introduced** below.

## Scope — what this does NOT cover

**A workspace over a Vendo Cloud-hosted store is not protected by this change.**

A client cannot open a transaction over HTTP, so the hosted backend's
`transact` passes through and the atomicity has to come from whatever serves
the wire. This repo's `createStoreOps` supplies it — each workspace verb runs
in one transaction and now takes the same path hold. **Vendo Cloud's console
does not.** `POST /api/v1/store/workspace/[op]` is a plain Next route into
`storeServiceDeps().withStore` → `withHostedStore`, which opens a fresh
`createStore({ url })` pool per request and closes it: no transaction, no lock,
no queue. Idempotency is not a stand-in — the key is `null` unless the caller
sends the header, and distinct keys both execute.

Closing that is **a separate piece, in `vendo-web`**, where the workspace door
is a `RecordStore` (`atomic.compareAndSwap(record, revision)`,
`insertIfAbsent`) and not SQL — a different mechanism, and not this PR's to
choose. Widening this PR into that repo would pre-empt the approach decision
being made from a scoping brief.

An earlier revision of this PR asserted that the wire was transactional
server-side. That was an assumption, and it is false; the rationale in the
commit and in the code comment on `workspaceOpsRows.transact` now says so
plainly.

## A deadlock this branch introduced (found in review, fixed here)

Flagged by Greptile as a P1, and it was right. Wrapping the façade's commit in
a transaction is what made it reachable: before that, every hold was released
at the end of its own statement and nothing was held across a pair.

Two commits touching the same paths in **opposite orders** — one removing
`left` then writing `right`, the other removing `right` then writing `left` —
each ended up holding what the other was waiting for. Reproduced on real
Postgres, which aborts one of them:

| | pre-fix (pg) | post-fix (pg) |
| --- | --- | --- |
| two commits, same paths, opposite orders | **5/5 `deadlock detected`** | 5/5 pass |

PGlite passes either way: one connection, so it cannot express two transactions
holding anything at once. That is the third time today this engine would have
reported the opposite of the real one.

**Severity, stated plainly:** this is a *loud* failure, not silent data loss.
Postgres detects the cycle and aborts one transaction with an error the caller
sees; nothing is lost or corrupted, unlike the three bugs this PR fixes. It is
still a new failure the branch introduced, so it does not ship unexamined.

**The fix is the standard shape**, not a narrowing of the transaction (the
wrapping is load-bearing and proven so above): take the union of every path the
transaction will lock, sort once by the lock key, acquire in that order, then
do the work. A total order cannot contain a cycle. The per-path holds inside
`land` and `remove` stay — they are what covers a caller touching a single
path, and inside a transaction that already holds them they cost nothing.

## The honest limitation

**An advisory lock is a convention, not a mechanism.** Nothing catches a future
writer that touches these rows without taking it first — the same critique #982
made of the transcript-writer lock, and we now have two of them. The
alternative that *is* enforceable is a soft-delete tombstone row, which
`FOR UPDATE` could lock like any other row; that needs a v8 migration and is
not what this takes. Presenting the lock as airtight would be wrong: it holds
because every writer in this file takes it, and that is a property of today's
code, not of the schema.

## Evidence — real Postgres, not PGlite

Run against a throwaway local Postgres (database and role created for the run
and dropped afterwards). **PGlite serialises and reports the opposite of the
real engine on the first two tests**, which is why they are gated on the real
one.

| test | pre-fix (pg) | post-fix (pg) |
| --- | --- | --- |
| undoing a create whose row is already gone | 5/5 fail | 5/5 pass |
| undoing back through a tombstone | 5/5 fail | 5/5 pass |
| undo inside the façade's own delete | 5/5 fail | 5/5 pass |
| two commits, same paths, opposite orders | 5/5 deadlock | 5/5 pass |

The first two PASS on PGlite pre-fix; the third fails on both, because the
racer completes inside the gap either way.

**The wrapping is load-bearing, not decorative**: with the advisory lock kept
and only the façade's transaction removed, the third test goes back to failing
on both engines while the first two still pass.

**The corrected scope above takes nothing away from this proof.** All three
tests drive the SQL backend — the façade over a local Postgres, and this repo's
own `createStoreOps` — so they never reached Cloud's door and never claimed to.
What was wrong was the inference that Cloud was covered by construction, not
the measurements.

## vendo-web

**No companion PR — written no-op for the code in this diff.** Cloud's
workspace door is `apps/console/lib/api/store-doors.ts`, an independent
implementation over the generic records door: one `vendo_records` row per
(owner, path) in a collection *named* `vendo_workspace_files` (not one of
`@vendoai/store`'s routed/dedicated collections), with each commit's
before-image on the ledger record instead of a `vendo_workspace_history` table.
It never imports `workspaceRows`, `workspaceStore` or `createStoreOps`, and
never touches the two tables this PR locks. Nothing here mirrors.

What it *does* have, and what the separate piece has to address: the same
check-then-act, entirely unheld. `pathsMovedOn` (`store-doors.ts:499`) decides
and the restore loop (`:720`) acts, with no lock of any kind — at an absent row
or a present one — on a connection that lives for one request.


yousefh409:
Closing unmerged — superseded by an owner decision, 2026-08-07.

Yousef has decided to **delete the undo/rollback feature entirely** across both repos: the apps history undo verb, the workspace undo ops in store/core and their conformance cases, the console rollback UI and its API route, and every doc promising rollback. Users lose the ability to roll an app back to a previous version — an explicit product call.

This PR added an advisory lock and a transaction wrapper to the workspace undo path. That path is being removed, so the fix would be thrown away.

**What this PR established, and why it was worth doing** — all of it fed the decision rather than the code:
- Three absent-path variants each failed **5/5 trials on real Postgres** and passed 5/5 after the fix.
- **PGlite reports the opposite of the real engine** on this class of bug — it serialises, so a lock or race test there can never disagree. That was found three separate times today and is now a standing warning for the repo.
- The transaction wrapping was proven load-bearing by removing it alone and watching a test go red.
- Its own rationale had claimed the hosted/Cloud leg was covered "because every op the wire serves is already a transaction server-side". **That was false** — the Cloud door opens a fresh pool per request with no transaction — and the PR was corrected in three places to say so rather than ship an overclaim.

The two genuine data-loss fixes from the same lane (#995) are already merged and stand: undo no longer releases a blob the live row points at, and undo of a superseded commit is refused rather than destroying a later value.

Also closed by this decision: the console workspace-undo race (`store-doors.ts` decide-then-act), which had been scoped to three options with a compare-and-set recommendation. No CAS work will be done — the door is being deleted.

PR #1024, opened 2026-08-07
cut(core): nine dead exports, one assertion vocabulary, three merged suites

yousefh409:
De-slop sweep of `@vendoai/core` outside `src/genui` (already swept). Worst-first slice; 443 net lines removed. Every finding was re-verified across all packages, `examples/`, `fixtures/`, `corpus/`, `docs-site/`, every test dir, and `vendo-web` — the catalog was wrong on five of them.

### Fixed / deleted

**Nine dead public exports.** `inferFieldSemantic`, `humanizeEnumValue`, `semanticAtPointer`, `semanticFormatToken` and `describeSemantic` lost their only caller when the dev-server inference pass was deleted (89153f883, #652). `startSseKeepalive` was justified in its own header by "the `vendo try` dev server writes to a Node `ServerResponse`" — there is no such SSE surface in the repo; `withSseKeepalive` is the live one. `VendoApprovalWirePart`, `VendoConnectWirePart` and `requestNumberValues` (exported for a capability-substitution gate removed in 0b58e3e21) had zero references anywhere, `vendo-web` included. The three money regexes stay — `declaredMoneyUnit` shares them — and its test now pins the "a bare total is a count" rule the deleted inference test carried.

**Comments state constraints, not the changelog that produced them.** Eight files carried dated provenance where the rule should be: `Amendment 2026-08-02:`, `wave-1 live proof E1-5`, `0.4.4 cert defect B`, `AGENT-7 (wave 5, additive — 01 §16 amendment parked)`, `rematch 2026-07-25 rows H12/H14`. Constraint sentences kept verbatim; attributions gone. `index.ts` also loses a five-line tombstone for a `./models.ts` deleted long enough ago that the note is archaeology.

**`src/security/` was a location, not a subject.** Its four suites tested descriptorHash, canonicalJson, the id prefix schemas and the tree caps — each of which already had a module and a sibling suite next door, which is how the same assertion ended up in both. descriptorHash's three files become one, and its local `descriptorCanonical` reimplementation of the preimage is gone: **it had already drifted**, omitting the `title` branch `descriptor-hash.ts:19` added, and passed only because no committed vector carries a title. `contract-coverage.e2e.test.ts`'s independent `node:crypto` oracle proves the canonical↔hash link, so nothing is lost. `id-schema.test.ts` and `tree-dos.test.ts` were unique — only their location was wrong.

**One assertion vocabulary for the conformance kits.** `assert` was written out four times, `assertDeepEqual`/`assertBytesEqual` twice each, `assertParses` twice (once zod-typed, once hand-typed to dodge the import) — `store-ops.ts`'s comment blamed "keep conformance self-contained", true of the barrel and not of a sibling file. They move to a non-exported `conformance/assertions.ts`, so the pinned `/conformance` export inventory is unchanged. The two diverged `assertDeepEqual` copies resolve to the informative one.

**`app-document.test.ts`** loses two byte-equivalent duplicate pairs, a case that re-asserted three earlier ones verbatim, a twin fixture, a twin helper, and a wire-compiler smoke test `packaging.e2e.test.ts` already pins off the packed artifact. **`conformance/knowledge.test.ts`** drops a `runConformance` pass that re-ran the cases it had just mounted one-by-one, plus `expect(suite.seam)` against a hardcoded literal; its posture check kept its job but stopped depending on the capital I in `includeInternal` to pass.

### Rejected — the catalog was wrong

- **"Four micro-suites re-test what `schemas.test.ts` covers"** — false. `schemas.test.ts` is one happy path per schema; the micro-suites hold nearly all the negative and boundary coverage, and `auditContext`, `authMaterialSchema` and `membershipSchema` appear in it not at all. Deleting them drops real coverage for three redundant assertions.
- **"`memory-store-ops.ts`/`memory-app-access.ts` have no outside consumers"** — `memoryStoreOps` is live in `packages/store/src/harness-state.ops.test.ts` and `thread-messages.ops.test.ts`.
- **"Three conformance kits wrap one assertion each"** — all three have external call sites (`packages/store`, `packages/actions`, `fixtures/automations-e2e`) and are pinned by name in `contract-coverage.e2e.test.ts`; `secretsProviderConformance` has two cases, not one.
- **"Getter-object wrapper adds `.it` to forty assertions"** — the wrapper is load-bearing: `packages/store/src/app-access.test.ts:229` passes a real getter, so `options.access` must be read lazily on every one of the 48 uses.
- **"`walkShapePointer` skips the pointer check its wrapper does"** — no caller can reach it. Its one in-repo caller goes through `splitPath`, which returns `""` or a `/`-prefixed pointer by construction, and `shapeAtPointer` guards the public entry. No failing test is writable through a real path, so the change would be pure defensive bloat.
- **`triggers.ts`, `screen.ts`, `make-receipt.ts`** carried no changelog narration — every paragraph states a live constraint.
- **`describeAt` / `describeShapeAt` merge** — `shape.ts` cannot import `semantics.ts` (layering), so the merge means threading `semantics`/`path`/`renderEnum` through the simple walk. Longer and less readable than the 12 duplicated lines.
- **`toolSemanticsSchema`** kept despite being unused in-repo: it is the declared mirror for a live type, and four consumers hand-spell `z.record(fieldSemanticSchema)` instead of importing it. Converting them is a cross-package change.

### Deferred (owner-level or outside this file-set)

- `island-derived-values.ts` deletion (~950 LOC): removing it leaves fabricated-FX-rate islands uncaught at generation time, and the AI reviewer only runs on the validate door. Product call, plus the call site is in `packages/apps`.
- `reshape.ts`'s `asOptions`/`template`/`currencyCents` retirement: dropping them routes stored apps through `fail("provision")`, killing the whole app render, not just those bindings. Needs a stored-app migration story.
- `store-wire.ts`'s 20 test-only request schemas: cutting 20 public exports of a shipped wire protocol is a semver call, not a sweep.
- Conformance case-body facade (14 overlapping record/blob cases across two different seam types), per-case DB lifecycle (fix belongs in `packages/store`'s mount), reserved-collection dedupe with `packages/store/src/validate.ts`, `GuardLike`, the four cross-package type moves, the HTTP status map, `StandardSchema`, `harnessStateKey`, `store.ts`'s two contracts — all need files outside `packages/core/**`.
- The 23 `docs/superpowers/...` citations pointing at a directory deleted in #967: 19 of them are in `src/genui`, so fixing four here would leave the inconsistency worse than one whole-package pass.

### Gates

- `pnpm build --filter=@vendoai/core...` — green
- `packages/core` vitest — **1071 pass, 57 files** (from 1085/60; 14 duplicate cases and 3 files gone)
- `packages/core` tsc, plus tsc with `src/**/*.test.ts` re-included (core excludes tests from typecheck) — 15 pre-existing errors, unchanged, none in touched files
- `pnpm typecheck` for core, apps, store, ui, actions, harnesses — green
- `dependency-guard` + `portability-gate` — green
- **Downstream canaries:** `packages/apps` `src/checking` + `src/generation` + `src/skills` (204 pass), and for the conformance refactor `packages/store`'s `conformance.test.ts`, `ops.conformance.test.ts`, `app-access.test.ts` — **115 tests against real PGlite**, green

### vendo-web

No companion PR needed. Every removed symbol was grepped across the whole console repo: zero hits. `apps/console/tests/store-conformance.test.ts` does not use the conformance kit at all — it imports `STORE_WIRE_PATHS`, `VENDO_STORE_WIRE_FORMAT` and `storeWireStatusSchema`, and resolves core from a pinned `vendor/vendoai-core-0.7.0-store.tgz`, so the kit refactor cannot reach it. The kits it does mount (`knowledge-conformance*.test.ts`) loop `suite.cases` without pinning names, and no case name or failure message changed.

---
## Summary by cubic
Removes nine unused public exports from `@vendoai/core`, unifies conformance assertions, and merges duplicated “security” tests into their modules for a leaner, clearer core with no behavior changes.

- **Refactors**
  - Dropped dead exports: `inferFieldSemantic`, `humanizeEnumValue`, `semanticAtPointer`, `semanticFormatToken`, `describeSemantic` (now internal), `startSseKeepalive`, `VendoApprovalWirePart`, `VendoConnectWirePart`, `requestNumberValues`.
  - Centralized conformance helpers in `conformance/assertions.ts` (`assert`, `assertParses`, `assertDeepEqual`, `assertBytesEqual`); not re-exported, public kit surface unchanged.
  - Removed `src/security/`; folded its suites into adjacent module tests to prevent drift. Comments now state rules, not provenance.

- **Tests**
  - Merged descriptor-hash/title checks into one suite; kept adversarial cases and duplicate-title detection. Vectors are pinned without re-implementing the preimage.
  - Expanded JCS adversarial coverage (nested refusals, Set/class instances, cyclic structures) in `jcs.test.ts`.
  - Relocated id-prefix and tree-cap suites beside their modules; deduped `app-document` cases and asserted exact validator messages where contract text is the API.
  - Simplified knowledge conformance posture checks (tier gating verified without rerunning the suite).

<sup>Written for commit 22f310742c6382683c51b25f1686ce5a42b50748. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1024?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Rebased onto `origin/main` (`6d08afdbb`). New head `cf65c3978`. Scope unchanged: 27 files, 392+/835−, `packages/core/**` + `.changeset/` only.

**One collision, resolved on the merits.** #1023 deleted the `describe("validateTree DoS bound gaps")` block from `packages/core/src/security/tree-dos.test.ts` — the same file this PR *moves* to `src/tree-dos.test.ts`. Rename-vs-edit; both intents are kept and I verified the outcome rather than trusting the auto-resolve: `src/security/` no longer exists, the moved file is 86 lines (was 103), the DoS-gap describe is gone, and all six imports are still used. We agree on the reasoning — an unfalsifiable assertion that goes RED the day someone adds the bound is worse than no test.

**Correction to the rebase brief:** #1023 did *not* touch `conformance/assertions.ts` or merge any conformance helper. Its only `packages/core` change is that 17-line deletion; `conformance/assertions.ts` does not exist on `main` and is introduced here. So there was no convergence to reconcile and no helper where the two sides disagree. The only other core change under this branch was #1017's one-line unused-import removal in `genui/tree-node.ts`, which this PR does not touch.

**Headline claims re-verified post-rebase:** the nine dead exports are still gone (eight at zero references anywhere in `packages/core/src`; `requestNumberValues` shows two — its definition and its one internal caller — because it is now private, not exported). `describeSemantic` is `const`, not `export const`. `startSseKeepalive` gone, `withSseKeepalive` untouched. descriptorHash is one suite file, `descriptor-title-hash.test.ts` and `security/` gone — and **the drifted local `descriptorCanonical` preimage copy is still absent**, which was the real find: it omitted the `title` branch `descriptor-hash.ts:19` added and passed only because no committed vector carries a title.

**Gates after the rebase:**
- `pnpm build --filter=@vendoai/core...` — green (1m15s)
- `packages/core` vitest — **1070 pass / 57 files**. One fewer than the 1071 reported before, which is exactly #1023's deleted test; `main`'s 1097 minus this PR's 27 removals reconciles.
- `packages/core` tsc — clean. Re-run with `src/**/*.test.ts` included (core excludes tests from `typecheck`): 15 errors, the same pre-existing baseline, **none in a file this PR touches**.
- `dependency-guard` — OK, 15 packages. `portability-gate` — all legs green.
- **Downstream canaries:** `packages/apps` `src/checking`+`src/generation`+`src/skills` — 204 pass / 5 skip. `packages/store` `conformance.test.ts` — 21 pass, and `ops.conformance.test.ts` — **53 pass**, both on real PGlite, run one suite at a time.

Two gate failures on the way were build-order artifacts of scoped builds, not regressions: `#store/db` and `packages/vendo/dist/server.js` resolve into `dist`, so `store` and `vendo` each needed their own build first. Both green after building.

The five rejected catalog claims and their reasons are unchanged in the PR body, `walkShapePointer`'s unreachable precondition included.

PR #1042, opened 2026-08-07
chore(packages): the four blocking-lint rules reach zero, 38 findings out

yousefh409:
The four rules the census picked go blocking in a separate, already-frozen PR. That PR cannot land while `packages/*` still reports findings for them — a gate that is red on day one is a gate everybody learns to bypass. This clears the runway and, more usefully, **measures how fast it re-dirties**.

## The count, measured at three heads

Re-measured by running the flip branch's frozen `eslint.blocking.config.mjs` unmodified, because the number moved 33 → 36 → 33 inside a single session before this branch existed:

| Head | Blocking findings | |
| --- | ---: | --- |
| `c4d7c76f3` — original base | **33** → 0 | the census backlog |
| `89d555358` — after +14 PRs | **6** came back → 0 | all six from one PR |
| `261902948` — after +5 more PRs | **0** new → still 0 | current head |

## Which PRs dirty this gate — the part worth keeping

Of the **eighteen** PRs that landed while this was open, **exactly one produced findings**: **#1036**, the undo/rollback deletion. The four UI bug fixes (#1027, #1029, #1040, #1009), the core cut (#1024) and #1050 produced **zero**.

**Deletion PRs are this gate's risk; ordinary work is not.** That is a much narrower claim than "main moves and findings come back", and it means the flip does not need a frozen repo — it needs to not land in the middle of a big deletion.

#1036's three sites: a dead `store` binding in `apps/src/lifecycle.test.ts`, a dead `dbFor` import in `store/src/workspace.test.ts`, and `refusal` in `store/src/workspace.ts`.

### `refusal` was checked, not deleted on sight

It lives in **shipped source**, so a dead store there could have been a refusal that never fires. It wasn't: its only caller was the `undo()` door #1036 removed, and `history()` refuses through the **viewer** predicate and already throws `pathNotFound` directly — which its own comment explains, because a caller who cannot even read is never told the path is there (§9.4). Dead, not swallowed. Deleting it orphaned `pathForbidden` and `VendoError` (both removed) and left a comment pointing at a function that no longer exists, which now says what it means instead.

## What changed

**31 deletions** — that is what "pure-deletion rules" means, the finding *is* the dead thing: three dead SSE/tools-call helpers in mcp's door test; two dead fixtures and two dead engine doubles left behind when the cases that used them went; a dead `DEFAULT_CLOUD_URL`; a `const { ts, sf }` nothing read; five unused imports.

**Two that are not deletions**, each with the reason written down:
- `store/src/workspace-fs.parity.ts` — the assertion alias has no reader **by design**: exporting it would put `import type … from "just-bash"` into every consumer's `.d.ts`, the dependency the file exists to prevent. Deleting it deletes the compile-time guard. It takes the config's own documented `^_` prefix.
- `harnesses/.../claude-code-box.live.test.ts` — the `text` accumulator nothing read is gone; the stream is still drained, because draining it is what drives the turn the assertions read.

**Five complexity findings become the shapes they already were.** No logic reshuffled to game the metric.

### `hosted-store.test-util.ts` — split, not moved, not exempted

It scored **92**. Two causes, neither essential: one `handler` carried five route families inline, and **the records ops were written twice** — once for the Store Wire v1 door (op in path, collection in body), once for the per-collection legacy door — two copies of one protocol, free to drift apart. Now one `recordsOp` both doors call, a named handler per family, and a `route` that reads as the route table it is. #1047's throwing `unserved` is preserved exactly as it landed.

**Moving it out of `src/` was checked and rejected**: `*.test-util.ts` is a repo-wide convention (ten packages), already excluded by every `tsconfig.json` and dropped from every coverage set. It has never shipped, so there was nothing to move it out of. **A narrow exemption was rejected too** — the flip config's opening line forbids per-file disables, and this complexity was incidental.

### The four UI ones

`TriggerStateLine` + `RunHistoryRow` out of the automations panel; `ThreadErrorBlock` (one shape for the two terminal failures that carried it twice), `ToolCallPart` + `AppCardBody` out of thread parts; `embedFlight`, `WorkspaceStage` + `WorkspaceToggle` out of the overlay.

Behaviour-preserving, verified by rendering every view at the base and at this branch and hashing: **14 of 16 byte-identical**, including `overlay`, `automations`, `thread-dark`, `approval`, `build-failed-banner`, `thread-humanized`, `thread-citations`, `tree-jail`, `tree-themed`, `appframe`. The two that differ (`slot-failed`, `slot-states`) **also differ between two runs of byte-identical code**, so the delta is headless-Chromium glyph rasterization, not this diff.

## One finding cleared itself

The unused `catch (reason)` in `connected-accounts-panel.tsx` is gone because **#1009 gave `reason` a real use** (`alreadyGone(reason)`). That file is no longer in this diff.

## Not done here

**No `eslint-disable` added anywhere. No rule weakened. No threshold raised.** The lockfile is byte-identical to `main`. Every test file touched is touched for a dead binding *inside that file* — no test was changed to make source code pass.

> **For the flip worker:** this zero is true at this SHA only. Re-measure at your own head, and check whether a large deletion PR landed in between — that is the one thing shown to reintroduce findings.

### DIFF
--- a/packages/actions/src/sync/static-ts.oracle.test.ts
+++ b/packages/actions/src/sync/static-ts.oracle.test.ts
@@ -19,7 +19,6 @@ import {
  * is safe and intended (see `oracleRead`).
  */
 
-type CaseMode = "match" | "divergent" | "permissive";
 
 interface BaseCase {
   name: string;
--- a/packages/apps/src/interchange.test.ts
+++ b/packages/apps/src/interchange.test.ts
@@ -39,14 +39,6 @@ const document = (overrides: Partial<AppDocument> = {}): AppDocument => ({
   ...overrides,
 });
 
-const codeEdit = JSON.stringify({
-  rung: 2,
-  files: [
-    { path: "/app/server.js", content: "export const ready = true;" },
-    { path: "/app/node_modules/cache/index.js", content: "export const cache = true;" },
-  ],
-});
-
 describe(".vendoapp interchange through createApps", () => {
   it("round-trips a copy with fresh ownership and empty data", async () => {
     const store = memoryStore();
--- a/packages/apps/src/lifecycle.test.ts
+++ b/packages/apps/src/lifecycle.test.ts
@@ -63,7 +63,6 @@ const setup = (withModel = true) => {
 
 describe("apps lifecycle", () => {
   it("disarms changed triggers on edit while preserving unchanged trigger edits", async () => {
-    const store = memoryStore();
     const original: AppDocument = {
       format: VENDO_APP_FORMAT,
       id: "app_trigger_arm",
--- a/packages/apps/src/model-params.test.ts
+++ b/packages/apps/src/model-params.test.ts
@@ -8,7 +8,6 @@
  */
 import type { LanguageModel } from "ai";
 import { describe, expect, it } from "vitest";
-import { askModel } from "./generation/engine.js";
 import {
   UNKNOWN_MODEL_MAX_OUTPUT_TOKENS,
   acceptsSamplingParams,
--- a/packages/harnesses/src/claude-code/claude-code-box.live.test.ts
+++ b/packages/harnesses/src/claude-code/claude-code-box.live.test.ts
@@ -298,13 +298,11 @@ live("claudeCode() — live, in a real e2b box", () => {
       thread: "m_box_env",
       say: "Write the full output of `env | sort` to user/files/env.txt. Then say done.",
     });
-    let text = "";
     try {
-      for await (const event of claudeCode({ sandbox: sandbox(), model: MODEL, maxTurns: 12 })
-        .run(h.turn as never)) {
-        if (event.type === "text") text += event.delta;
-        if (event.type === "error") text += `\n[error] ${event.message}`;
-      }
+      // Drained to drive the turn; the proof is the env dump the agent writes,
+      // not anything the stream says.
+      for await (const _event of claudeCode({ sandbox: sandbox(), model: MODEL, maxTurns: 12 })
+        .run(h.turn as never)) { /* no-op */ }
     } finally {
       delete process.env["VENDO_LANE_E_BOX_CANARY"];
     }
--- a/packages/mcp/src/door.test.ts
+++ b/packages/mcp/src/door.test.ts
@@ -3396,22 +3396,6 @@ function mcpRequest(accessToken: string, sessionId?: string, resource = BASE) {
   });
 }
 
-/** `mcpRequest` covers initialize and tools/list; this is any other method on a
- *  live session, over raw HTTP. */
-function mcpCall(accessToken: string, sessionId: string, name: string, args: Record<string, unknown>) {
-  return new Request(BASE, {
-    method: "POST",
-    headers: {
-      authorization: `Bearer ${accessToken}`,
-      accept: "application/json, text/event-stream",
-      "content-type": "application/json",
-      "mcp-session-id": sessionId,
-      "mcp-protocol-version": "2025-11-25",
-    },
-    body: JSON.stringify({ jsonrpc: "2.0", id: 42, method: "tools/call", params: { name, arguments: args } }),
-  });
-}
-
 /** The standalone SSE stream a client opens for server-initiated notifications.
  *  `accept` is a parameter so a test can send the GET the transport REJECTS. */
 function sseRequest(accessToken: string, sessionId: string, accept = "text/event-stream") {
@@ -3423,39 +3407,6 @@ function sseRequest(accessToken: string, sessionId: string, accept = "text/event
       "mcp-protocol-version": "2025-11-25",
     },
   });
-}
-
-/** The first SSE frame already buffered on a stream. Reading does NOT cancel the
- *  body, which is what lets a test hold a stream the server still believes in. */
-async function firstFrame(response: Response): Promise<string> {
-  const reader = response.body!.getReader();
-  const { value } = await reader.read();
-  reader.releaseLock();
-  return new TextDecoder().decode(value);
-}
-
-/** The next frame on a stream that is supposed to still be REACHABLE, or `""`.
- *  The wait never costs a passing run anything: the frame is already buffered by
- *  the time the assertion reads, so only a deaf stream ever reaches the deadline. */
-async function frameOrNone(response: Response, ms = 5_000): Promise<string> {
-  const reader = response.body!.getReader();
-  let arrived = false;
-  const read = reader.read()
-    .then(({ value }) => { arrived = true; return value === undefined ? "" : new TextDecoder().decode(value); })
-    .catch(() => { arrived = true; return ""; });
-  let timer: ReturnType<typeof setTimeout>;
-  try {
-    return await Promise.race([
-      read,
-      new Promise<string>((resolve) => { timer = setTimeout(() => resolve(""), ms); }),
-    ]);
-  } finally {
-    clearTimeout(timer!);
-    // Released only when the read actually finished, so a test may read the same
-    // stream again: releasing under a PENDING read rejects it, and nothing would
-    // be left to observe that rejection.
-    if (arrived) reader.releaseLock();
-  }
 }
 
 async function pkceChallenge(verifier: string): Promise<string> {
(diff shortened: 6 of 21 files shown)


yousefh409:
> [!WARNING]
> **Superseded.** This records the gates at `aaf02b935`, before two rebases (main moved 19 PRs, including #1036 which reintroduced 6 findings, and five PRs touching `packages/ui`). The `express-host` failure described below was fixed upstream by #1046, and its CI blind spot by #1048. Fresh gate evidence follows in a later comment; do not rely on this one.

---## Local gates (the two CI cannot run)

Both run foreground on this machine, one at a time. Every package this PR touches is green; the four failures below are all proven **not** to belong to this PR.

### 1. `pnpm test:affected` — 41/43 packages green

`Failed: @vendoai-corpus/express-host#test, demo-bank#test`. Neither is a package this PR touches, and both were run down rather than waved off:

| Failure | Verdict | Evidence |
| --- | --- | --- |
| `demo-bank › src/vendo/away-drill.test.ts` — `beforeAll` hook timed out at 240s | **Load-induced, not mine** | Passes alone in **29.16s**. CI's `integration` job runs `pnpm --filter demo-bank test` and it passed green on this exact head. |
| `@vendoai-corpus/express-host › e2e/doctor.e2e.test.ts` — dev-only doctor probes answer 404 | **Pre-existing on `main`** | Reproduced at pristine `c4d7c76f3` in a separate worktree: *byte-identical* assertion and `broken:` diagnostic. |

**Second CI blind spot found.** `@vendoai-corpus/express-host` declares no `test:coverage` script, but CI's `test-rest` runs `turbo run test:coverage` — so that package's suite **never executes in CI at all**, and it is red on `main` today. Same structural shape as the `packages/vendo` typecheck gap: a suite that exists, fails, and cannot be seen.

### 2. `pnpm --filter @vendoai/ui test:browser` — 79 passed / 2 failed, identical to pristine

Real Chromium, 88 specs. The failure set is **byte-identical** before and after:

```
1) [chromium] › e2e/accessibility.spec.ts:26:3 › page has zero WCAG 2.1 A/AA axe violations
2) [chromium] › e2e/screenshots.spec.ts:23:3 › captures page.png
```

Both are the `page` scenario (`VendoPage`'s "Apps" tab never reaches `aria-selected="true"`) — untouched by this PR and already failing at `c4d7c76f3`. A third failure (`extreme-content.spec.ts:48`) appeared in one full-suite rerun and **passes 3/3 in isolation**: flake under load, not a regression.

Every spec that exercises the refactored components passed: `build-failed-banner`, `run-history-voice`, `pass3-consumer-voice`, `slot-states`, `smoke`, `keyboard`, `theme`, `scroll-management`, `jail-and-tree`.

## Screenshots: 14 of 16 views byte-identical to pristine

For a behaviour-preserving refactor, "identical" is stronger evidence than a picture. Rendered at `c4d7c76f3` and at `aaf02b935`, then hashed:

| View | Renders which extraction | sha256 (16) |
| --- | --- | --- |
| `overlay.png` | `WorkspaceToggle`, `WorkspaceStage`, `embedFlight` | `0af1f423ca2e1980` ✅ |
| `automations.png` | `TriggerStateLine`, `RunHistoryRow` | `1d2f37634e8cf8fd` ✅ |
| `thread-dark.png` | `ThreadPart`, `ToolCallPart` | `5f9effe4b24aec36` ✅ |
| `build-failed-banner.png` | `ThreadErrorBlock` | `4880bb74166dcf17` ✅ |
| `approval.png` | `ThreadPart` | `86233e170059c78b` ✅ |
| `thread-humanized.png` | `ThreadPart` | `548bebdd5457217b` ✅ |
| `thread-citations.png` | `ThreadPart` | `341480f2ded0886d` ✅ |
| `tree-jail.png` / `tree-themed.png` / `appframe.png` | `AppCardBody` | `86731b79…` / `778fe7d7…` / `83b53e97…` ✅ |
| `palette.png` / `notice.png` / `slot-picker.png` / `slot-building.png` | overlay + card chrome | all ✅ |

`slot-failed.png` and `slot-states.png` differ — and that is **not** this PR: they also differ **between two runs of byte-identical code** on this branch, so the delta is headless-Chromium glyph rasterization. Both were opened and compared by eye: same layout, same copy ("This view didn't build" / "I couldn't finish building that view — nothing was changed…"), same button positions. Their spec passed in every run.

Two views confirmed visually rather than only by hash:
- **`automations.png`** — `TriggerStateLine`'s middle branch renders live: "● Enabled · waiting on 2 permissions", the exact nested-ternary chain that became three early returns, dot and all.
- **`overlay.png`** — `WorkspaceToggle` renders the collapsed-state corner-bracket glyph, which is the specific place two byte-identical `<svg>`s were merged into one with a conditional `d`.

Both PNG sets are on the build machine at `/tmp/deslop/lint-clear-logs/screenshots-{before,after}/`.

yousefh409:
## Local gates — green at `77250a2ab`

Rebased onto `aeced100e`. Both gates CI cannot run, plus the two it can, foreground on a quiet box.

| Gate | Result |
| --- | --- |
| `pnpm build` (scoped) | ✅ |
| `pnpm typecheck` (scoped) | ✅ — now covers `hosted-store.test-util.ts`, see below |
| `pnpm --filter @vendoai/ui test:browser` | ✅ **81 passed, 0 failed** |
| blocking-rule lint (frozen config) | ✅ **0** |
| `pnpm test:affected` | every package accounted for — detail below |

**Three of my findings landed as their own pieces while this was open**, and each one changed my gate:
- **#1054** typechecks vendo's test files and cleared **96** errors — so `hosted-store.test-util.ts` is now inside CI's typecheck. My rewrite passes under it.
- **#1055** dropped the `/page` scenario, clearing the two browser failures I reported. The suite is now **81/81** rather than 79/2.
- **#1046/#1048** fixed the `express-host` doctor failure and the CI blind spot that hid it.

## Screenshots: 12 of 12 byte-identical to `main`

Re-baselined against **current `main`** in a separate worktree at `aeced100e`, not the stale tree.

| | |
| --- | --- |
| `appframe` `approval` `automations` `notice` `overlay` `palette` | identical |
| `thread-citations` `thread-dark` `thread-humanized` `tree` `tree-jail` `tree-themed` | identical |

**Read the method before the result, because the first comparison was wrong.** A naive hash diff said *six* views differed. Five of those were an artifact of my own harness: my PNGs came from the full 81-test run, the baseline's from an isolated 12-test spec run, and load/animation timing differs enough between those two conditions to change the bytes. Re-running `screenshots.spec.ts` **alone on both trees** — matched conditions — every view matches.

The noise floor was measured, not assumed: running the baseline spec **twice on the same commit** leaves `overlay.png` differing from itself. So `overlay.png` matching one of base's two renders is exactly right, and hash equality alone would have reported five regressions that do not exist.

## `test:affected`

Two fixtures failed in the full local run and both were run down individually rather than waved off:

| Failure | Verdict |
| --- | --- |
| `@vendoai-fixtures/mcp-e2e` — `ECONNREFUSED` | Passes alone: **7 files / 19 tests**. Concurrent `next build` of two example apps was wiping a `distDir` the fixture's server was serving from. |
| `@vendoai-fixtures/automations-e2e` — `ECONNREFUSED` in `resetFixture` | **Stale `dist/` after a rebase.** Passes clean on `main` (17 files / 69 tests) and passes clean here after an explicit rebuild: **17 files / 69 tests**. |
| `@vendoai/vendo` — `harness-system-prompt` timed out at 30s (earlier run) | Load-induced, and now fixed upstream by **#1053**. The test needs 21.8s of its 30s budget alone, so it loses whenever the box is busy; vendo passed on the rebased head. |

**The `automations-e2e` one is worth naming:** a rebase without rebuilding `dist/` makes an e2e fixture boot a host against mismatched packages, and the symptom is `ECONNREFUSED` — indistinguishable from a real server bug. Rebuild before attributing any fixture connection failure to a diff.

## Still true

No `eslint-disable` anywhere, no rule weakened, no threshold raised, lockfile byte-identical to `main`. Every test file touched is touched for a dead binding inside that file.

> This zero holds at `77250a2ab`. `main` moved **26 PRs** while this was open and exactly one — #1036, the undo/rollback deletion — reintroduced findings. #1054 touched ~75 vendo test files and added none.

PR #1056, opened 2026-08-08
fix(demo-bank): standing judgments bind to /api again, so keyless reads run

yousefh409:
## The demo's payoff moment rendered empty

A generated view rendered fine in the thread but the pinned home-hero slot showed
*"Some values below couldn't load — that isn't your data being empty."* plus raw
`MoneyHq*: generated component rendered no content` strings. The server named the
mechanism on every poll:

```
[vendo] query "hostListGoals" (tool "host_listGoals") resolved no data for app
app_demo_moneyhq_vendo-demo — the call answered "pending-approval"; anything
bound to it renders empty
```

## The grades were never missing — they were inert

All 22 grades were sitting in `.vendo/judgments.json` (17 `read`, 4 `write`,
1 `destructive`). Every one was being discarded.

`applyJudgment` ([`packages/actions/src/judgments.ts:130`](../blob/main/packages/actions/src/judgments.ts#L130))
drops a judgment whose recorded binding no longer matches
`bindingIdentity(tool.binding)` — deliberately, so a judgment of a handler that
moved never re-points at whatever now answers under that name.

Maple moved its `/maple` prefix out of the OpenAPI **paths** and into the spec's
**`servers`** url — the shape `src/lib/base-path.ts` documents ("the OpenAPI
spec's server url … is how the prefix reaches the agent's host tool calls") and
`mount-point.test.ts` pins (`expect(spec.servers).toEqual([{ url: BASE_PATH }])`).
`tools.json` regenerated as `GET /api/goals`; `judgments.json`, last written
before that move, still said `GET /maple/api/goals`.

**22 of 22 bindings mismatched, so all 22 tools kept their extraction grade of
`ungraded`.**

From there the reported chain is exactly right:

1. `ungraded` matches none of `policy.json`'s `read`/`write`/`destructive` rules.
2. The guard's no-match default withholds `ungraded` exactly like `destructive`
   ([`core/src/grant-sets.ts:188`](../blob/main/packages/core/src/grant-sets.ts#L188)).
3. Plain reads park as `pending-approval`, and everything bound to them renders empty.

`srcHash` is a red herring — it is never consulted at runtime, only by the judge
deciding whether to re-grade.

## The fix

Strip the stale prefix from the 22 binding strings. Nothing else changes.

`vendo doctor` before → after:

```
warning: catalog: 19/22 tools ungraded — each one asks on every call
ok:      catalog: all 22 tools graded
```

(The 3 that were already graded are exactly the 3 carrying `risk` in
`overrides.json` — that layer is keyed by tool NAME, so it never drifted.)

Effective grades are now 17 `read` / 2 `write` / 3 `destructive`, zero `ungraded`.
Reads match `read → run`; `transferMoney`, `createOrder` and the demo reset stay
`destructive` — nothing loosened.

### Why not the alternatives

- **Put real grades in `tools.json`** — wrong by design. Grades never live there;
  the judge writes them to `judgments.json` and only patches *schema* slots into
  `tools.json`. A `tools.json` full of `ungraded` is correct even after a
  successful grade. Hand-edits would also be overwritten by `predev`'s `vendo sync`.
- **Add an `ungraded` rule to `policy.json`** — `ungraded → ask` is already what
  happens, so it changes nothing; `ungraded → run` would auto-run
  `host_transferMoney`. Not a fix, a hole.
- **Stop committing grades** — backwards. `predev` runs `vendo sync --no-ai`,
  which cannot grade without a model key, so the committed judgments are the
  *only* source of grades on a BYO host. They must be committed **and** correct.

## Keyless browser proof

Keyless = BYO `ANTHROPIC_API_KEY`, **no `VENDO_API_KEY`** (the posture
`README.md` documents). Real browser, real generation, sign in → Ask Maple →
"Build my money HQ" → Pin to dashboard → home.

Controlled A/B on the **same pinned app**, flipping only this file:

| | home hero |
|---|---|
| stale bindings | error copy + 4 `MoneyHq*: rendered no content` strings; 8 `resolved no data … "pending-approval"` in the server log |
| repaired | Money HQ renders live data (Net $142,929.30, real balances); **0** misses in the log |

Restoring the file flips it back, so the causation is pinned to this one artefact.

## Also fixed by this, reported not chased

The standing-access / approval cards labelling **every** permission
*"Not reviewed:"* is the same root cause. `grantRowWord`
([`ui/src/chrome/grant-set-card.tsx:57`](../blob/main/packages/ui/src/chrome/grant-set-card.tsx#L57))
derives the word purely from the grade, and `ungraded → "Not reviewed"`. With
every host tool ungraded, every row said it. Those rows now read
"Reads:" / "Changes:" / "Irreversible:". Its own doc comment prescribes exactly
this fix: *"The honest fix for a mis-graded tool is to fix the GRADE."*

## Gate

`pnpm build --filter=demo-bank...` · demo-bank 115/115 tests · typecheck · lint
(0 errors; 4 pre-existing warnings in untouched files).

---
## Summary by cubic
Restores demo-bank data rendering by re-binding judgments from `/maple/api` to `/api`, so grades apply and keyless reads run. Fixes empty home hero and pending-approval reads without loosening destructive actions.

- **Bug Fixes**
  - Updated 22 bindings in `examples/demo-bank/.vendo/judgments.json` to `/api/*` to match the OpenAPI `servers` base path.
  - Standing grades now apply (reads run; `transferMoney`, `createOrder`, demo reset remain `destructive`).
  - Clears the “Not reviewed” labels in standing-access; rows now show Reads/Changes/Irreversible.

<sup>Written for commit aa2582d50b2a18140649159e5deb84c55636dd4b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1056?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/examples/demo-bank/.vendo/judgments.json
+++ b/examples/demo-bank/.vendo/judgments.json
@@ -2,7 +2,7 @@
   "format": "vendo/judgments@1",
   "tools": {
     "host_auth_create": {
-      "binding": "POST /maple/api/auth/{}",
+      "binding": "POST /api/auth/{}",
       "srcHash": "sha256:48d5e41965cfc171bcd9bbae81d91a8994f7f1aa6ebab60dfba3bfa5bd7a86ec",
       "fields": {
         "title": "Sign in or sign out",
@@ -11,7 +11,7 @@
       "evidence": "      authorize: async (credentials) => {\n        const user = await authenticateMapleUser(\n          typeof credentials?.email === \"string\" ? credentials.email : \"\",\n          typeof credentials?.password === \"string\" ? credentials.password : \"\",\n        );\n        return user ? { id: user.subject, name: user.display, email: user.email } : null;\n      },"
     },
     "host_auth_get": {
-      "binding": "GET /maple/api/auth/{}",
+      "binding": "GET /api/auth/{}",
       "srcHash": "sha256:48d5e41965cfc171bcd9bbae81d91a8994f7f1aa6ebab60dfba3bfa5bd7a86ec",
       "fields": {
         "title": "Run an Auth.js sign-in endpoint",
@@ -20,7 +20,7 @@
       "evidence": "import { handlers } from \"@/auth\";\n\nexport const runtime = \"nodejs\";\nexport const { GET, POST } = handlers;"
     },
     "host_createOrder": {
-      "binding": "POST /maple/api/orders",
+      "binding": "POST /api/orders",
       "srcHash": "sha256:08a7f7a6ad93e85a49032d398199e943a4964320a8a634a606e4203ddd85b64e",
       "fields": {
         "description": "Appends one posted card transaction (defaults: DoorDash, 3184 cents, late-night Pacific timestamp, dining) to the shared demo store and returns it. Despite the \"charges the card\" summary it never debits a balance and places no real order.",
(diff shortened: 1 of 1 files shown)


yousefh409:
### Browser evidence

Screenshots on the verifying machine at `/home/ubuntu/screenshots/pin-fix-0808/`:

| file | what it shows |
|---|---|
| `before-10-home.png` | **bug** — stale bindings: error copy + `MoneyHqAccounts/Spending/Goals/Bills: generated component rendered no content` |
| `after-02-thread-rendered.png` | keyless generation in the thread, live balances |
| `after-10-home.png` | **fixed, keyless** — pinned Money HQ on the home hero with live data |
| `restored-10-home.png` | re-applying the fix flips it back (A/B/A) |
| `keyed-10-home.png` | same, with `VENDO_API_KEY` (Cloud) |
| `reset-02-after.png` | `Reset demo` swept: seed balances, empty invite slot |

Server-log counts of `resolved no data … the call answered "pending-approval"` for the same pinned app:
**stale bindings → 8 · repaired → 0** (keyless and keyed).

`Reset demo` verified: `connections: []`, parked approvals `8 → 0`, home hero back to the
empty invite, total balance back to the seeded `$142,929.30`. No rows left on the Cloud account.

PR #1068, opened 2026-08-08
fix(core): prevent joinUrl paths from climbing above mount

### LINKED ISSUE #988: joinUrl normalises dot-dot segments — a crafted path can climb above the configured mount
Flagged by automated review on #957, verified in code review, currently NOT reachable (every joined path comes from our own generated catalog, not user input) — filing so the sharp edge has an owner before something user-influenced ever reaches the joiner.

`joinUrl` in packages/core/src/url.ts applies WHATWG URL normalisation, so a path like `/api/../../admin` resolves above the configured base path instead of being refused. The one-owner design means a single guard in `joinUrl` (reject or strip `..` segments before joining) fixes every call site at once.

Suggested fix shape: refuse a `pathOrUrl` containing `..` segments (loud error naming the input), or normalise-then-verify the result still sits under the base via the existing `underPathPrefix`. One test per behavior.

Akarshkushwaha:
## What
Fixes `joinUrl` in `packages/core/src/url.ts` to prevent directory traversal paths (e.g., `../`) from escaping the configured base mount prefix. 
Fixes: #988 .

## Why
`joinUrl` relies on WHATWG URL normalization. Before this fix, a crafted path could successfully escape the base URL and resolve higher up in the directory tree. We now normalize the URL first and strictly verify that its final resolved `pathname` remains securely bounded within the original `underPathPrefix()`. 

## Verification
- Added comprehensive test coverage in `packages/core/src/url.test.ts` to verify that variations of directory climbing (e.g. explicit `../`, encoded `%2e%2e`) throw an error.

---
## Summary by cubic
Prevent directory traversal in `joinUrl` so joined URLs cannot escape the base mount in `packages/core`. We now validate the normalized path stays under the base and throw on inputs like `../` or `%2e%2e`; tests added in `packages/core/src/url.test.ts`.

<sup>Written for commit 067e3ba397a46a7afb9a77274e0497e9fa0232fd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1068?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/core/src/url.test.ts
+++ b/packages/core/src/url.test.ts
@@ -93,6 +93,12 @@ describe("joinUrl", () => {
     expect(joinUrl("https://svc:pw@api.site.com/maple", "/api/x").href)
       .toBe("https://svc:pw@api.site.com/maple/api/x");
   });
+
+  it("refuses a path that climbs above the base", () => {
+    expect(() => joinUrl("https://site.com/maple", "/../../admin")).toThrow(/resolves outside base/);
+    expect(() => joinUrl("https://site.com/maple", "../admin")).toThrow(/resolves outside base/);
+    expect(() => joinUrl("https://site.com/maple", "%2e%2e/admin")).toThrow(/resolves outside base/);
+  });
 });
 
 describe("joinPath", () => {
--- a/packages/core/src/url.ts
+++ b/packages/core/src/url.ts
@@ -85,6 +85,9 @@ export function joinUrl(base: string | URL, pathOrUrl: string): URL {
     ? (path === "" ? "/" : path)
     : withPathPrefix(path, pathOrUrl.startsWith("/") ? pathOrUrl : `/${pathOrUrl}`);
   const joined = new URL(suffix, url.origin);
+  if (!underPathPrefix(path, joined.pathname)) {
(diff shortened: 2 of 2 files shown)


yousefh409:
Thanks for taking a run at this. On a closer look the guard checks the deployment mount, but the traversal escapes the *tool's own route* — so on a normal bare-origin API base (`https://api.example.com`) the check is a no-op and the climb still lands (`/admin`, etc.). It also stringifies the base into the error, which would leak a basic-auth `VENDO_HOST_API_URL` into logs, and it breaks relative `VENDO_LOGIN_URL`.

The bigger news: #988 turned out to be reachable today, so we've fixed it at the source (rejecting `..` in path-argument substitution) in #1439. Closing this one — but genuinely appreciate you surfacing the sharp edge.

PR #1075, opened 2026-08-08
genbench: claude-code contender driver (fenced; wiring follows slice 3)

yousefh409:
The strong in-house baseline for the buy-vs-build benchmark: **Claude Code with
hands** — a coding agent in a scratch directory, writing and rewriting one
`index.html` — not a single generation call.

## Fair by construction

The session's brief is assembled from the functions the Vendo column is already
built on, imported rather than copied, so a second description of the world
cannot drift from the first:

- `hostDesignBrief({ theme, designRules })` — the identical theme tokens and
  host rules the screen agent thinks with (`packages/apps`, via `vendo.ts`'s own
  `designRules`)
- the tool schemas `world.ts` derives, and the rows each read answers with
- the page contract the Vendo column gets for free from the product:
  `window.vendo.callTool(name, args)`, `window.__settled`, a `role="dialog"`
  confirmation step, no network. Stating these is what keeps the columns judged
  on the screen instead of on knowing the harness.

## Measurement

Billed by its own session — the SDK spawns its own engine and never touches
`meter.model` — so the SDK's tokens ride the outcome and are priced through the
same `usdFor` table as every other column. A >5% gap between the SDK's own
`total_cost_usd` and that table is warned about, never silently averaged. The
meter is still the run's only clock: snapshot `atMs` and `settledMs` come from
it.

Loadout: `Read`/`Write`/`Edit` in a temp dir that dies with the case, no shell,
no network, `settingSources: []` so a laptop's own CLAUDE.md and skills never
become this column's advantage. A ten-minute wall clock ends a session that will
not end itself, and the page it had written by then is still what it delivered.

## Live smoke (real session, real browser)

`pending-transfers`, sonnet, opened from disk in headless Chromium at 480x900:

| | |
| --- | --- |
| settled | 53,444 ms (first page on disk at 50,709 ms) |
| tokens | 39,632 (4 in · 6,454 out · 18,802 cache read · 14,372 cache write) |
| cost | $0.1564 at genbench prices |
| artifact | 10,616 bytes, one write |
| renders standalone | yes — settled, mounted, **zero console errors** |
| wired | both Cancel controls fire `cancel_transfer` with the row's own id (`tr_1`, `tr_2`) through a `role="dialog"` confirmation |

Screenshots and the artifact live in the run evidence, not in the repo.

## The fence

This branch touches **only** `genbench/src/claude-code.ts`,
`genbench/src/claude-code.test.ts`, and the SDK dependency in
`genbench/package.json` + lockfile. `run.ts`, `report.ts`, `world.ts`,
`floor.ts`, `probe.ts`, `render.ts`, `cases.json` and `world.json` are
untouched, so this column is **not yet in the matrix** and changes no existing
run.

### Wiring that follows slice 3

Three seams in `run.ts`, shared with the `diy` column (also html-format), which
is why they wait for it rather than landing twice:

1. **matrix** — `DRIVERS` gains `"claude-code": (model) => claudeCodeDriver({ model })`,
   its type widens to `(model: ModelAlias) => Contender`, and the call site
   passes `contender.model`.
2. **page** — an outcome with `format: "html"` skips `pageHtml()`; its
   `artifact` *is* `page.html` (as `README.md` already says it should be).
3. **cost** — `cost:` reads the outcome's own `usage`/`usd` when present, since
   this contender is not metered by `meter.model`.

Until then the driver is dead code with a green suite behind it.

## Tests

`claude-code.test.ts` mocks the SDK boundary and nothing else — the double
writes into the workspace the driver really made, so the file watching,
snapshots and cleanup are the driver's own behaviour. Covers world-block
equality against the shared serializer, snapshot order and `atMs` monotonicity,
the timeout path returning a failure instead of throwing, workspace cleanup on
both paths, and SDK usage priced through `usdFor`. Each assertion was proven to
fail against a deliberately broken driver before being trusted. The live smoke
is an eighth test, gated on `GENBENCH_LIVE=1` plus a key, and skips cleanly.

Gate (scoped, foreground): `pnpm build` ✓ · `npx vitest run` in `genbench/` ✓
(38 passed, 1 skipped, run twice) · `pnpm typecheck` ✓ · `pnpm lint` ✓

---
## Summary by cubic
Adds a `claude-code` contender that uses `@anthropic-ai/claude-agent-sdk` to build a single `index.html` in a temp dir, recording writes and pricing usage with genbench rates. Fenced until slice-3 wiring; no existing runs change.

- New Features
  - `claude-code` driver: agent writes `index.html` in a scratch dir with `Read`/`Write`/`Edit`, no shell/network/settings, 10‑min wall‑clock timeout, cleans workspace.
  - Prompt includes the same world as Vendo: `hostDesignBrief(...)` plus derived tool schemas and rows; enforces page contract (`window.vendo.callTool`, `window.__settled`, role="dialog").
  - Captures ordered write snapshots and timing; reports SDK `usage` and computes `usd` via `usdFor` with a warning if SDK bill differs by >5%.

- Dependencies
  - Added `@anthropic-ai/claude-agent-sdk` (>=0.3.0 <1).

<sup>Written for commit d5fc16e23c5e01064aa289d97caf56703c09e431. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1075?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/genbench/package.json
+++ b/genbench/package.json
@@ -24,6 +24,7 @@
   "devDependencies": {
     "@ai-sdk/anthropic": "3.0.12",
     "@ai-sdk/provider": "3.0.2",
+    "@anthropic-ai/claude-agent-sdk": ">=0.3.0 <1",
     "@playwright/test": "^1.49.0",
     "@types/node": "^22.0.0",
     "@types/react": "^19.2.0",
--- a/genbench/src/claude-code.test.ts
+++ b/genbench/src/claude-code.test.ts
@@ -1,0 +1,217 @@
+/**
+ * The claude-code contender, with everything real except the SDK.
+ *
+ * The double is the SDK boundary and nothing else: it writes into the workspace
+ * the driver actually made, on disk, so the file-watching, the snapshots and the
+ * cleanup are the driver's own behaviour and not a fixture's.
+ */
+import { createAnthropic } from "@ai-sdk/anthropic";
+import { hostDesignBrief } from "@vendoai/apps";
+import { existsSync } from "node:fs";
+import { writeFile } from "node:fs/promises";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { beforeAll, describe, expect, it } from "vitest";
+import { claudeCodeDriver, worldBlock, type AgentSdk } from "./claude-code.js";
+import { meteredModel, MODEL_IDS, usdFor, type Meter } from "./meter.js";
+import { designRules } from "./vendo.js";
+import { loadCases, loadWorld, worldForCase, type Case, type World } from "./world.js";
+
+const root = dirname(dirname(fileURLToPath(import.meta.url)));
+let world: World;
+beforeAll(async () => {
+  world = await loadWorld(join(root, "world.json"));
+});
+
+const caseFor = (id: string): Case => ({ id, lane: "screen", prompt: "Show my pending transfers", pass: [] });
+
+/** The run's clock, and nothing else the driver may lean on: this contender is
+ *  billed by the SDK's own session, so a meter it actually used would be a lie. */
+function clock(): Meter {
+  let tick = 0;
+  return {
+    model: "claude-sonnet-5",
+    elapsedMs: () => (tick += 1),
+    totals: () => ({ inputTokens: 0, outputTokens: 0, cacheReadTokens: 0, cacheWriteTokens: 0, calls: 0 }),
+    usd: () => 0,
+  };
+}
+
+const SDK_USAGE = {
+  input_tokens: 1_200,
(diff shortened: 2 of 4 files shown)


yousefh409:
Landed via the stack tip #1080, squash-merged to main (2026-08-09). This PR's commits are contained in that merge; closing as landed-via-tip.

PR #1079, opened 2026-08-08
genbench: judge core (fenced; integration follows slice 4)

yousefh409:
Adds the non-mechanical half of the genbench score: one verdict per rubric line, from a model shown the screenshot, the click trace and the source. The floor already answers what a machine can settle alone (renders / honest data / wired actions); this answers the rest.

**Nothing calls it yet.** Two new files, zero edits to anything existing.

## Fence

Created exactly two files, edited nothing else:

- `genbench/src/judge.ts` (211 lines, ~120 of them code)
- `genbench/src/judge.test.ts`

No new dependencies. `zod` is not declared by `genbench/package.json` and pnpm here is strict, so the schema is built with `jsonSchema()` re-exported from `ai` — same `Schema<T>` that `generateObject` takes.

## Pending integration seams (next slice)

- `run.ts` calling `judge()` per contender/case and adding a `judged` field to `result.json`
- the preview's per-line verdict rows
- deciding whether a `degraded: true` judgement blocks a run's exit code the way the floor does

`judge(input)` is callable as-frozen. The optional second argument (`{ model?, delayMs? }`) exists only so tests can mock the model boundary without sleeping through real backoff; the run passes nothing and gets the pinned model.

## What it refuses to do

**It refuses to know whose screen it is.** Nothing sent names a contender, a model or a run folder, and rubric lines are shuffled before sending and mapped back after.

This turned out to matter more than the contract anticipated, and it is the one thing here that needs your call. **Both** columns name the product in their own artifact, in opposite directions:

- the in-house baseline, because `diy.ts:28` instructs it to call `vendo.callTool(name, args)` — so its document says the name in every wired control
- the product, because its document is stamped `VENDO_APP_FORMAT` = `"vendo/app@1"` and `formatVersion: "vendo-genui/v2"`

So an unblinded artifact hands the judge the answer, and hands it *backwards* on the baseline — the column that says the name loudest is the one that is not the product. The contract said the prompt must contain no harness names *and* said not to strip the artifact; those collide here.

Resolution: strike the **name** out of text evidence, leave the **format** untouched. A tree still reads as a tree (`ui:"tree"`, component names and structure all survive — pinned by a test). Verified the live smoke returns **identical verdicts** before and after, so nothing was lost. Revert is dropping two `blind(...)` wrappers if you disagree.

**It refuses to half-grade.** `jsonSchema` does no runtime validation and no provider enforces an enum, so a short answer or an off-rubric verdict counts as no answer: 3 attempts, backoff only for a provider that is briefly unwell, then every line fails with `degraded: true` and the error recorded. It never throws. Line text and source are always copied from the caller, so a judge that paraphrases a line back cannot rewrite the rubric.

**It refuses to move with the contender.** `JudgeContract` pins `model: "claude-opus-5"`, `rubricVersion: 1`, and `promptHash: sha256(SYSTEM_PROMPT)`.

## THE SYSTEM PROMPT — draft, needs your sign-off

`rubricVersion` bumps on ANY edit; this is verbatim from `genbench/src/judge.ts`.

```text
You are grading one screen of a software product against a fixed checklist. You are not its designer, its author, or a reviewer offering advice: you decide, line by line, what the evidence supports.

THE EVIDENCE, in priority order. Where two sources disagree, the earlier one wins.
1. THE SCREENSHOT — the screen exactly as a person sees it. This is what the user actually gets.
2. THE INTERACTION TRACE — every control on the screen was pressed once, and this records what each press asked the application to do. This is what actually happened when the screen was used.
3. THE SOURCE — what the screen was built from. This is only what was intended. The source may be written in any format, and its format is not evidence: it must never affect a verdict. A line the source promises but the screenshot does not show is not satisfied.

Return exactly one verdict for each numbered checklist line, in the order the lines are numbered — no more, no fewer.
- pass: the evidence clearly shows this line is satisfied.
- fail: the evidence clearly shows this line is violated, OR the line applies to this screen and the evidence does not show it satisfied. Not demonstrated is not a pass.
- na: the line's subject does not occur on this screen at all, so there is nothing here to satisfy or violate — for example, a line about confirming destructive actions on a screen that only displays information. Use na only for an absent subject, never for your own uncertainty: when the subject is present and you are unsure, the verdict is fail.

Every verdict carries a note: one clause naming the specific evidence you used, such as "the header reads Spending" or "pressing Cancel called nothing". No advice, no praise, no summary, and no restating the line back.

Grade only the numbered lines. Anything else you notice about this screen, good or bad, is not yours to grade: it must not change a verdict and must not appear in a note. Judge the screen you were given, not the screen you would have built.
```

## Evidence

**Live smoke** (gated behind `GENBENCH_LIVE=1` + `ANTHROPIC_API_KEY`; skips cleanly otherwise). Grades a real screenshot and a real click trace of a fixture screen built to earn all three verdicts. `claude-opus-5`, one call:

```
[case]  PASS  shows every spending category the tool returned
              six categories Housing through Coffee are each listed as rows
[case]  PASS  shows a total for the month equal to the sum of the categories
              Total $4,243.11 equals the six listed category amounts summed
[case]  PASS  housing is presented as the largest category
              Housing $2,850.00 is listed first and is the largest value shown
[style] FAIL  uses the theme exactly: brand green, Onest, 8px corners — no invented colors or fonts
              button is blue #2563EB with Arial and 2px radius, not green/Onest/8px
[style] PASS  money always shows 2 decimals with currency symbol
              all amounts render as $2,850.00, $612.45, $61.30 etc. with symbol and two decimals
[style] FAIL  destructive actions ask for confirmation
              pressing "Cancel transfer" called cancel_transfer({id:'tr_1'}) immediately with no confirmation step
[style] NA    dates shown as 'Aug 1' style, never ISO
              no dates appear anywhere on the screen
```

Worth noting: it cited the *trace* to fail the confirmation line, and correctly returned `na` rather than `fail` for a rubric line whose subject the screen does not contain.

**Every assertion proven able to fail.** 13 mutations of `judge.ts`, each caught by the suite it was aimed at — including two rounds where a mutation was *missed*, which exposed a vacuous trace-blindness assertion and an untested key check (both fixed; the key check was deleted as defensive code the provider already does better).

```
CAUGHT  blindness: leak the whole input into the prompt
CAUGHT  blindness: stop blinding the artifact (the real leak both columns have)
CAUGHT  blindness: stop blinding the click trace
CAUGHT  blindness: blind so hard the format goes too
CAUGHT  remap: hand back verdicts in the ASKED order
CAUGHT  remap: let the judge overwrite the caller's rubric line
CAUGHT  shuffle: never shuffle at all
CAUGHT  degrade: throw instead of flooring the screen
CAUGHT  retry: one attempt only
CAUGHT  partial grading: accept whatever came back
CAUGHT  validation: stop checking the verdict is on the rubric
CAUGHT  schema: drop na from the allowed verdicts
CAUGHT  contract: pin a stale prompt hash
```

Tests mock only the model SDK boundary (`MockLanguageModelV3`) and assert on the call options the SDK actually produced, so blindness is measured on the bytes that would go over the wire — not on a helper's return value.

**Local gate** (scoped; CI's green check is the gate of record):

| gate | result |
|---|---|
| `pnpm build` | exit 0 |
| `npx vitest run --maxWorkers=2 --minWorkers=2` in `genbench/` | 68 passed, 1 skipped — run twice |
| `pnpm typecheck` | exit 0, 43/43 |
| `pnpm lint` | exit 0 |

---
## Summary by cubic
Adds the judge core for genbench: per-line grading of case and style rubrics using the screenshot, click trace, and source. It’s fenced and not wired into runs yet.

- **New Features**
  - Adds `judge()` (`genbench/src/judge.ts`) to return `pass`/`fail`/`na` with a short evidence note for each line.
  - Enforces blindness: strips contender/model/run names from text evidence while preserving format; shuffles lines and remaps answers.
  - Reliability: JSON result schema via `jsonSchema` from `ai`, retries transient errors up to 3 times, and degrades the whole screen on unusable answers (no partial grades).
  - Pins the grader via `JudgeContract` (model `claude-opus-5`, `rubricVersion: 1`, `promptHash`); includes a draft system prompt.
  - Fence/tests: only `genbench/src/judge.ts` and `genbench/src/judge.test.ts` added; includes a gated live smoke test; no new dependencies.

<sup>Written for commit ffed7924f4ac36c59e2a8846a0dea2528925cde2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1079?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/genbench/src/judge.test.ts
+++ b/genbench/src/judge.test.ts
@@ -1,0 +1,427 @@
+import { MockLanguageModelV3 } from "ai/test";
+import { createHash } from "node:crypto";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { describe, expect, it } from "vitest";
+import { judge, JudgeContract, SYSTEM_PROMPT, VERDICTS, type JudgeInput, type Verdict } from "./judge.js";
+import { probe, type Probed } from "./probe.js";
+import { authoredPage, openBrowser } from "./render.js";
+import { loadWorld } from "./world.js";
+
+// ------------------------------------------------------------------ fixtures
+
+/** A 1x1 PNG. Small and fixed, so the base64 the SDK sends can never
+ *  accidentally spell one of the strings the blindness test forbids. */
+const PNG = Buffer.from(
+  "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg==",
+  "base64",
+);
+
+const TRACE: Probed[] = [
+  { label: "Cancel", confirmed: true, calls: [{ name: "cancel_transfer", args: { id: "tr_1" } }] },
+  { label: "Refresh", confirmed: false, calls: [] },
+];
+
+const CASE_LINES = ["alpha shows every row", "bravo totals the rows", "charlie confirms deletions"];
+const STYLE_LINES = ["delta uses the theme colors", "echo formats money with two decimals"];
+
+const input = (over: Partial<JudgeInput> = {}): JudgeInput => ({
+  screenshot: PNG,
+  artifact: "<section><h1>Spending</h1><p>Housing 2850</p></section>",
+  trace: TRACE,
+  caseLines: CASE_LINES,
+  styleLines: STYLE_LINES,
+  ...over,
+});
+
+const ZERO_USAGE = {
+  inputTokens: { total: 0, noCache: 0, cacheRead: 0, cacheWrite: 0 },
+  outputTokens: { total: 0, text: 0, reasoning: 0 },
+};
+
+type Answer = { verdict: Verdict; note: string };
+
+const replied = (verdicts: Answer[]) => ({
+  content: [{ type: "text" as const, text: JSON.stringify({ verdicts }) }],
+  finishReason: { unified: "stop" as const, raw: undefined },
+  usage: ZERO_USAGE,
+  warnings: [],
+});
+
+/** Every numbered checklist line the model was actually asked about, in the
+ *  order it was asked — parsed back out of the assembled prompt, so the tests
+ *  see exactly what went over the wire and nothing the judge merely intended. */
+const asked = (call: { prompt: unknown }): string[] => {
+  const text = JSON.stringify(call.prompt);
+  const parsed = JSON.parse(text) as Array<{ content: unknown }>;
+  const parts = parsed.flatMap((message) =>
+    Array.isArray(message.content) ? (message.content as Array<{ type: string; text?: string }>) : [],
+  );
+  const checklist = parts.filter((part) => part.type === "text").at(-1)?.text ?? "";
+  return [...checklist.matchAll(/^\s*\d+\.\s+(.+)$/gm)].map((match) => match[1]!);
+};
+
+/** The verdict this line is worth, decided by its own first word — so a remap
+ *  bug shows up as a verdict landing on the wrong line. */
+const owed = (line: string): Verdict =>
+  line.startsWith("alpha") || line.startsWith("delta")
+    ? "pass"
+    : line.startsWith("charlie")
+      ? "na"
+      : "fail";
+
+/** A model that answers each line it was actually asked, in the asked order. */
+const answering = (): MockLanguageModelV3 =>
+  new MockLanguageModelV3({
+    doGenerate: async (call) =>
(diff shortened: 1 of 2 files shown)


yousefh409:
Landed via the stack tip #1080, squash-merged to main (2026-08-09). This PR's commits are contained in that merge; closing as landed-via-tip.

PR #1089, opened 2026-08-09
refactor: one identity model — the host's principal resolver, or nothing (31 ops → 27)

yousefh409:
One identity model, one store. Vendo stops minting identities of its own: the
host's `principal(req)` resolver is now the single source of who a request is,
and everything built to paper over its absence — guest minting, the anonymous
cookie, the session registry, the merge-on-sign-in dance — is deleted rather
than deprecated. Pre-1.0, hard cut, no shim.

## The contract shrinks: 31 ops → 27

`lifecycle.adopt`, `sessionRegister`, `sessionStale`, and `sessionClaim` are
gone from the `StoreOps` type, the wire manifest, the request schemas, and the
conformance kit. `packages/store` drops the session engine, the `vendo_sessions`
table's engine, and `adoptEphemeralSubject`; `ops.status()` reports 27.

## Zero-config identity is gone

`createVendo` now throws at construction if no `principal` resolver (and no
`auth` preset, which supplies one) is composed, naming the exact line to add.
A resolver that returns `null` for a request refuses **that request** with
`forbidden` (403) instead of minting an ephemeral session.

```ts
principal: async () => ({ kind: "user", subject: "dev" })
```

## ⚠ `GET /status` now returns 403 to a caller with no identity

`/status` resolves context like every other route, and the refusal lives in the
shared context resolver. A health check or uptime monitor that hits `/status`
cold — no cookie, no header, nothing the host's resolver can turn into a
principal — now gets **403 instead of 200**.

This is a deliberate consequence of "no modes, no special cases": there is no
route that is exempt from identity, because an exemption is a second identity
model. Hosts that monitor `/status` must either send whatever their resolver
reads, or return a principal for unauthenticated callers.

The portability gate caught this on its first real run — its Worker fixture
was wired `principal: async () => null` and its `/status` probe went 403. The
fixture now returns a principal (`scripts/fixtures/portability-worker/worker.mjs`).

## ⚠ Breaking config: `config.sessions` → `config.sweep`

```ts
// before
sessions: { sweepIntervalMs: 60_000, ttlMs: 1_800_000 }
// after
sweep: { intervalMs: 60_000 }
```

- `sessions` → `sweep` (`{ intervalMs?: number; now?: () => number }`)
- `sessions.sweepIntervalMs` → `sweep.intervalMs`
- `sessions.ttlMs` → **removed outright, no replacement.** It was the idle
  guest-session TTL and there are no guest sessions. Delete the line.

What survives under the new name is the TTL cadence for parked BYO calls and
stranded approvals, which was never about sessions.

## ⚠ Every logged-out `demo-bank` visitor now shares ONE identity

`examples/demo-bank/src/vendo/server.ts:76` resolves logged-out visitors to a
single `maple_guest` principal. Two logged-out browsers therefore see each
other's threads and apps. Per-visitor anonymous isolation is exactly what this
design removes — a host that wants it issues its own per-visitor subject.
demo-bank runs locally, so this is accepted.

## `Principal.ephemeral` SURVIVES

What died is MINTING, not the field. A host is free to mark a resolved
principal `ephemeral: true`, and every door that refuses one still refuses it:
`packages/vendo/src/wire/misc.ts` (dev in-client approval injection, 401),
`connections.ts` (connected accounts), and the MCP door. Only the comments that
described those principals as "Vendo-minted anonymous sessions" changed.

## Deleted tests — the behavior is removed by design

`packages/vendo`: `anon-session-race` (no anon session to race), `hosted-sessions`
(no hosted session registry), `session-lifecycle` (no session lifecycle),
`session-sweep-failure` (mocked `sweepEphemeralSubjects`, which no longer exists
— its surviving posture is re-pinned, see below).
`packages/store`: `ephemeral`, `ephemeral-lifecycle`, `helpers/subjects` (no
ephemeral-subject engine).
`packages/ui`: `anon-session-race` (the race closer is deleted; `request`
collapsed to `send`).
`fixtures/integration`: `anon-merge.e2e` (no merge-on-sign-in),
`anonymous-isolation.e2e` (no anonymous sessions to isolate).

### One test added, to keep a surviving posture pinned

`packages/vendo/tests/sweep-failure-isolation.test.ts` — the deleted
`session-sweep-failure` also pinned something that outlived it: a failing
background sweep must not 500 the innocent request that merely triggered it
(`packages/vendo/src/server.ts`, the catch around `startSweep(false)`). The new
test injects a fault at the real store boundary for `vendo_parked_call` only,
so the surviving parked-BYO-call leg rejects out of `runSweep`, and asserts the
request still answers 200 and the warning is emitted. Verified RED by deleting
that catch, GREEN with it restored.

## Test files edited as SETUP, not assertions

Stated out loud, because the rule is that test changes are their own change.
The full set of surviving test files this branch touches:

Setup only — a principal where the old default supplied one, or the renamed
config key:

- `packages/vendo/tests/byo-approvals-wire.test.ts:96` — `sessions` →
  `sweep` config key rename.
- `packages/vendo/tests/automations-defer.test.ts` ×3 — explicit `principal`
  in harness setup.
- `packages/vendo/tests/review.fixture.test.ts` ×2 — same.
- `packages/vendo/tests/server.test.ts:1694,:3872` — same (the rest of that
  file's −346 is the inline session cases, deleted with the feature).

Renamed, not weakened — the case is now driven by a HOST-resolved ephemeral
principal instead of a Vendo-minted one, and the assertion is unchanged:

- `packages/vendo/tests/situation-abuse.test.ts` — "ANONYMOUS visitor" →
  "EPHEMERAL visitor"; still asserts the situation channel cannot forge a
  `Directions` section.
- `packages/vendo/tests/connections.test.ts` — the ephemeral subject is renamed
  `visitor_abc`; `initiate` still rejects it.
- `packages/vendo/tests/wire/context.memberships.test.ts` — same reframing; an
  ephemeral principal still belongs to no org.
- `packages/vendo/tests/user-facts-seam.test.ts` — "no [User] block for an
  anonymous visitor" becomes "when the seam asserts nothing about the user",
  since a request with no identity is now refused before it gets there.

Cases deleted with the feature they covered:

- `packages/vendo/tests/wire/context.facts.test.ts` — the `sessionStore` /
  `sessions` fixture scaffolding and "never asks the seam for an anonymous
  visitor".
- `packages/vendo/tests/hosted-store.test.ts` (−118), `hosted-store.test-util.ts`,
  `packages/store/src/backends.test-util.ts` — the four removed ops.
- `packages/store/tests/erase.conformance.test.ts`, `effects-erase.test.ts`,
  `workspace-ops.test.ts`, `workspace.erase.test.ts` — `vendo_sessions` rows
  dropped from the cascade fixtures.

Counts moved to match the contract:

- `packages/store/tests/schema.test.ts` — 21 contract tables → **20**.
- `packages/store/tests/thread-messages.ops.test.ts` — 31-op → **27-op**.

Comment only:

- `packages/guard/test/ungraded-default.test.ts:572` — a
  stale "anonymous-subject adoption" cross-reference in a doc block. Zero
  assertion or logic change.
- `fixtures/integration/tests/sync-impact.e2e.test.ts:102` — **comment only**.
  The test asserts a 404 (the route is absent off a development stack) and
  still passes untouched; only the "the anonymous caller" wording changed.
- `fixtures/integration/tests/connections.e2e.test.ts` — removed
  "refuses to initiate for an anonymous (ephemeral) visitor", which drove the
  door by omitting the principal header. Omitting it now means 403 at the
  context resolver, not the connections door, so the test no longer reached
  what it claimed to test. The door's ephemeral refusal is unchanged.
- `fixtures/redteam/tests/oauth-csrf.e2e.test.ts` — removed the `__Host-`
  anon-cookie fixation assertion (there is no anon cookie), plus its two
  header/comment references. The CSRF content-type and import media-type
  probes are untouched and still pass.
- `scripts/fixtures/portability-worker/worker.mjs` — the gate fixture now
  resolves a principal (see the `/status` section).
- `examples/demo-bank/tests/demo-script/console-seed.test.ts:39` — explicit
  `principal` at construction. Caught by CI's `integration` job, whose last
  step is `pnpm --filter demo-bank test`. Its `beforeAll` was throwing, which
  vitest reports as a SKIPPED test, so the suite went from
  116 passed | 1 skipped to 117 passed — the fix un-skips real coverage.

Non-test comment fixes: `packages/vendo/src/wire/misc.ts` (stale "minted
ephemeral session" wording above a 401 whose logic is untouched) and
`packages/store/README.md` (the erase cascade says 20 tables, matching
`ERASE_TABLES`; it read 16 and was already wrong before this project).

## The advertised one-liner now compiles — fixed here

Both refusal messages, the three docs pages, and the changeset told a host to
write `principal: () => ({ subject: "dev" })`. That did not compile, on **two**
counts: the option's type is `(req: Request) => Promise<Principal | null>`, so
the sync form is a `TS2322` — and `Principal.kind` is required
(`packages/core/src/principal.ts:9`), so `{ subject: "dev" }` is a second
`TS2322` even once it is `async`. A host who copy-pasted our own error message
got a red build.

All six sites now read:

```ts
principal: async () => ({ kind: "user", subject: "dev" })
```

`packages/vendo/src/compose-config.ts:105`, `packages/vendo/src/wire/context.ts:35`,
`docs-site/quickstart.mdx:104`, `docs-site/reference/handler-options.mdx:159`,
`docs-site/reference/server-api.mdx:48`, and
`.changeset/vendo-no-longer-mints-a-guest-identity.md:25`.

Widening `principal` to accept a sync return — or defaulting `kind`, which the
wire already constrains to `"user"` for host resolvers
(`packages/vendo/src/wire/context.ts:48`) — was considered and **rejected**. It
reads better, but it widens the public config surface and this project only
removes surface. The snippet now matches the real config surface instead of the
surface bending to the snippet. It is also the form the CLI scaffold
(`packages/vendo/src/cli/init-scaffolds.ts:40`) and `docs-site/agents/host-auth.mdx:60`
already used — the five broken sites were the outliers, not the house style.

The reason this shipped at all is that nothing proved the advertised snippet
compiles. `packages/vendo/tests/advertised-principal.test.ts` now constructs
`createVendo` with exactly that line; `tsconfig.test.json` includes `tests`, so
the snippet IS the assertion and a `typecheck` run is the gate. Verified it can
fail in **both** directions, so neither defect can regress alone — dropping
`async` gives `TS2322: Type '() => { kind: string; subject: string; }' is not
assignable`, and dropping `kind` gives `TS2322: ... Property 'kind' is missing
in type '{ subject: string; }' but required in type 'Principal'`.

## Known, and deliberately deferred to the cleanup stage

- `packages/vendo/src/harness-turn.ts` `evictSubject` is now callerless inside
  the repo but stays — it is public API.
- `packages/store` keeps the `vendo_sessions` `DROP TABLE` for a later
  migration, after existing guest data is erased.


yousefh409:
Re: the P1 on `examples/demo-bank/src/vendo/server.ts:80` — the finding is **factually correct and deliberate**, and it is disclosed under its own heading in the PR body. Recording the reasoning here rather than resolving it silently.

**There is no fix that preserves both properties.** Per-visitor isolation for logged-out visitors *was* Vendo-minted anonymous sessions. Deleting those is the entire point of this change (see the 31→27 op contract shrink). So a host that wants to serve signed-out traffic must now choose an identity for it, and any fixed choice is shared by construction. The only alternatives are (a) require sign-in, or (b) have the host key a subject off something of its own — both are host decisions, and the code comment at that line says exactly that.

**Scope check on the severity.** `demo-bank` is `private: true`, has no deploy pipeline in this repo (it appears only in `ci.yml` as a test target), and no hosting config. It is run locally via `pnpm --filter demo-bank dev`. So "do not deploy this where multiple unauthenticated users can reach it" is the right warning, and it is already in the source comment — but it describes a deployment that does not currently exist.

**Not changing it in this PR.** Making the demo require sign-in is a product decision about the demo's UX, not a defect in this change, and it is outside what this project was scoped to do. Flagged for the owner to decide. It is a one-line change either way and nothing downstream depends on it.

The genuinely load-bearing part of your summary — that identity responsibility moved to the host — is correct and intended.

PR #1105, opened 2026-08-09
Ux/dock and element picker

noah-getpie:
A new sidebar
A Zap tool for removing page elements
A screenshot tool that provides visual context to the AI
A theme carousel
Undo and “Restore All” options for reverting changes
A border around the webpage to clearly indicate when the user is in editing mode
[https://www.loom.com/share/f2f223b49eeb447bb36b0a7dfb0f2ba1](url)

---
## Summary by cubic
Adds an opt-in docked placement for `VendoOverlay` that parks the panel beside the page and reflows the page so users can edit in place. The default remains the centered modal; docked is deliberately non-modal.

- Public API: `VendoOverlay` adds `placement` ("center" default, "dock") and `dockWidth` (px, default 420) in `@vendoai/ui`.
- Docked behavior: no scrim, no scroll-lock, no inert background, no focus trap, no `aria-modal`. Reflow applies to `documentElement` via `data-vendo-dock` and `--vendo-dock-w`, refcounted across multiple panels.
- Mobile: both placements still collapse to the full-bleed takeover.
- UI adjustments: workspace expander is hidden while docked; auto-stage is disabled in docked mode (embeds still land in the rail); launcher shifts left to stay clickable.
- Feedback: an indeterminate editing bar sweeps along the top edge of the framed page while a turn runs.
- Adoption: no change unless you opt in. To use the side panel, set `placement="dock"` and optionally `dockWidth`.

<sup>Written for commit 9a2cea2e8a6800aa6f7d304c906db508089f8cc8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1105?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/docked-panel.md
+++ b/.changeset/docked-panel.md
@@ -1,0 +1,45 @@
+---
+"@vendoai/ui": minor
+---
+
+Dock the conversation panel beside the product — opt-in.
+
+`VendoOverlay` takes a `placement` prop. `"dock"` parks the panel against the
+right edge at full height and reflows the host page into the remaining width,
+so the surface being reshaped stays visible and clickable while the panel is
+open. `dockWidth` (default `420`) sets the panel width and, with it, how far
+the page reflows. Below the mobile breakpoint both still collapse to the
+full-bleed takeover.
+
+`placement` defaults to `"center"`, the centered modal that has always
+shipped, so this release changes nothing for an existing host: the scrim, the
+body scroll-lock, `inertBehind` and `aria-modal` are all still there unless a
+host asks for `placement="dock"`.
+
+Docked is deliberately NON-modal — no scrim, no body scroll-lock, no
+`inertBehind`, no focus trap, and no `aria-modal` — because a modal that
+covers the page is the wrong shape for a tool whose job is editing that page.
+The page reflows via a width reduction on `documentElement` (not `body`,
+whose width is usually author-controlled), torn down on close, unmount, and
+placement flips. Host chrome that is itself `position: fixed` is anchored to
+the viewport rather than to `documentElement`, so it does not reflow with
+this; such elements can read `--vendo-dock-w` to inset themselves.
+
+The reflow is owned centrally and refcounted: `data-vendo-dock` and
+`--vendo-dock-w` live on the one `documentElement` every overlay shares, so
+closing one of two open docked panels no longer hands the page back its full
+width while the other is still open.
+
+The workspace expander stays a centered-placement feature — a full-height rail
+has nowhere to grow a stage into — so it is hidden while docked. For the same
+reason a docked conversation does not auto-stage a built app: staging on the
+user's behalf there would strand an app they could neither see nor collapse.
+The embed still lands in the rail.
+
+**An indeterminate progress bar** sweeps along the top edge of the framed page
+while a turn runs, driven by the existing cross-tree run-activity store. No
+percentage: `RunActivity`'s `done`/`total` count steps already begun, not a
+forecast, and inventing a completion estimate would break the same "no fake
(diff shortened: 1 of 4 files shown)


yousefh409:
Hi Noah, thank you again for this, genuinely good work.

The docked sidebar is merging through this PR, your commits and your authorship intact. We'll handle the rebase onto current main and the three sidebar fixes.

Element picker and region capture I'd love back as two follow-up PRs on fresh main. I'm dropping them from this branch so the sidebar lands clean and fast, not because we don't want them; they're still in your fork history. Findings triaged below so those PRs sail through.

Restyle templates are parked for now: picking a preset sets `selected` and nothing reads it, so it needs real wiring before it ships.

Thanks again. Around for questions on any of it.

Yousef

<details>
<summary>Triage for the follow-ups</summary>

**Element picker**
- [ ] Select auto-arms on open, and the capture-phase click swallower then kills every host-page click (`vendo-overlay.tsx:676`). Fix first — it's why tests here were pinned to `placement="center"`. Arm on an explicit user action.
- [ ] `uniquePath` stops at 10 ancestors, so "Zap this" silently no-ops on anything deeper (`element-picker.tsx:123`).
- [ ] "Zap all related" degrades to `div > div` on classless elements and can hide the dock itself (`element-picker.tsx:113`).
- [ ] The swallower misses `pointerdown`/`pointerup`, so host pointer handlers still fire (`element-picker.tsx:509`).
- [ ] The zap bar sits under an `aria-hidden` ancestor, so its buttons are invisible to assistive tech (`element-picker.tsx:719`).

**Region capture**
- [ ] `html.slice(0, 12_000)` can cut mid-tag, so the model gets malformed markup (`region-capture.tsx:104`).
- [ ] The capture/cancel decision runs inside the `setStart` updater, so it double-invokes in StrictMode (`region-capture.tsx:162`).

**Attachment path (travels with them)**
- [ ] Pasted markup renders twice (`message.tsx:92`); a failed send can drop the chip (`composer.tsx:192`); the `isPastedCode` prefix fallback misreads typed prose (`message-data.ts:431`).
- [ ] No tests on ~1,000 lines. `buildScopeLevels`, `relatedSelectorFor`, `uniquePath` and `markupForRegion` test easily in jsdom. The one I'd insist on: `pastedCodePart` → `readPastedCode` round-tripped through the real write and read paths, no stub on either side.

</details>

yousefh409:
Merged — thanks again, Noah. It's in as `6fd3bfaba`, with your name on the commit.

One change I want you to hear from me rather than spot in the diff: I kept `center` as the default and made the sidebar opt-in via `placement="dock"`. Your version defaulted to dock, which would reflow every existing host's page and drop the scrim, scroll-lock and `aria-modal` on a routine version bump. It was also getting expensive — you'd pinned five tests to `placement="center"` to cope with it, main moved since, and that turned into four more. Opt-in let us delete all nine and leaves nobody surprised on upgrade. The feature itself is exactly as you built it.

Three things I fixed on the way in, all with your commits intact: dock ownership is refcounted now, so closing one of two docked panels no longer strips the reflow from the survivor; the 4px horizontal overflow is gone (`box-sizing: border-box` on the docked `html`); and a docked conversation can't auto-stage an app into a state you can't see. The dead `.fl-overlay-pick` rules went out naturally with the picker.

Element picker and region capture are yours whenever you want to pick them up — the checklist from my earlier comment still stands, and both will land much easier on their own.

Yousef

Issue #914, opened 2026-08-06
path-prefix deployments: host-tool calls double-prefix, so generated views render empty (third seam beyond #866/#867)


yousefh409:
Found during the sprint's cold-walk proof (2026-08-06), driving a real MCP client against demo-bank.

**Under a path prefix, `VENDO_BASE_URL` cannot serve both roles:**
- Set to `http://host/maple` → the door's OAuth/discovery is correct, but every **host-tool call double-prefixes** and 404s. Generated views then render empty behind a "values couldn't load" banner.
- Set to the bare origin → host tools work, but discovery advertises 404 endpoints.

This is a **third seam**, server-side host-tool execution, distinct from #866 (login redirect) and #867 (demo-bank returnTo). No client-side workaround can reach it.

Also: #866 has an unlisted third manifestation — the **consent page's own `<form action>` drops the prefix**.

The cold walk worked around discovery, the login POST, and the consent action in its own harness only (no product code touched) and completed the full chain, so everything past these URL seams is healthy.

Impact: any prefix-mounted self-hosted deployment. Broker-mode deployments (demos.vendo.run/maple) are unaffected for OAuth but the host-tool half applies wherever the prefix shape is used.

PR #1136, opened 2026-08-10
fix(vendo,demo-bank): doctor catches an unset mount base URL; Maple's spec and corpus stop teaching a wrong API

yousefh409:
## What

Four reported host-tool defects. **Three were real and are fixed. One was not a defect** — its *doc comment* was, and only that is corrected. Details per defect below.

---

### 1. `vendo sync` and the relative OpenAPI mount — NOT A DEFECT (comment only)

**Reported:** `openApiMountPath` documents itself as "applied by the caller across ALL extractors — see `runExtractors`", but `runExtractors` never calls it, so a spec with `servers: [{"url": "/maple"}]` yields bindings without the mount and every tool call 404s.

**Found:** the comment is stale; the behavior is deliberate and correct.

`mounted()` was **deleted on purpose** in #957 — commit `2357b2268`, titled *"fix(actions): stored tool paths are prefix-free — `mounted()` deleted (#914)"*, under conductor ruling #14 ("spec law wins"). Baking the prefix into stored paths is exactly what produced `/maple/maple/…` in #914. The law has been pinned by a regression test ever since:

```
packages/actions/tests/sync/openapi.test.ts
  "keeps stored paths prefix-free whatever the spec's relative server declares"
```

The prefix has one home — `VENDO_BASE_URL` — and `core`'s `joinUrl` attaches it once at call time. Implementing the requested change would require **deleting that regression test**, and because `withPathPrefix` is idempotent it would not even change the resolved URL in a correctly configured deployment. I re-applied `mounted()` locally to measure it:

| test | with `mounted()` re-added |
| --- | --- |
| `keeps stored paths prefix-free …` (existing) | ❌ `/cadence/api/dashboard` ≠ `/api/dashboard` |
| `reaches the mounted endpoint … via VENDO_BASE_URL` (new seam) | ✅ **unchanged** — idempotent join |
| `lands one prefix short when the base URL drops the mount` (new seam) | ❌ prefix appears anyway |

So the requested change is a **no-op in the correct posture and a corruption otherwise**. Only the lying comment is fixed.

**The reported 404 symptom is real** — but its cause is defect 2, not extraction. It happens when the base URL lacks the mount, which is precisely what `doctor` was failing to catch.

#### New seam test (no mock on either side)

`packages/actions/tests/sync/openapi.test.ts` now runs a real spec with a relative server url through the **real extractors**, into a **real `createActions` registry**, and reads the URL off the registry's **own outbound fetch** — producer and consumer can no longer agree on a prefix-free path and 404 together:

- `VENDO_BASE_URL=https://host.example/cadence` → `https://host.example/cadence/api/dashboard` (prefixed exactly once)
- `VENDO_BASE_URL=https://host.example` → `https://host.example/api/dashboard` (**one prefix short — the hazard**)

Both proven able to fail: sabotaging `withPathPrefix` to never attach reddens the first; re-adding `mounted()` reddens the second.

---

### 2. `vendo doctor` disarmed itself in the one posture that breaks — FIXED

`checkMountAgreement` returned early when `VENDO_BASE_URL` was unset — the zero-config dev posture, which is **exactly** where the failure occurs:

- stored binding paths are prefix-free by law (spec 2026-08-06 §B1), and
- with no base URL the wire learns the **bare request ORIGIN** (`compose-wire.ts` `onRequestOrigin` → `server.ts:355` passes `url.origin`, no path),
- so a path-mounted host serves every host tool one prefix short. Every page renders; every tool call 404s.

`E-CFG-003` now fires there. The existing disagree/agree branches keep their behavior and the three messages share one fix sentence. Error code and its registry entry are unchanged.

**Proven able to fail:** restoring the early return reddens the new test with `expected undefined to match object { status: 'broken' }` — i.e. the check emitted *nothing at all*.

---

### 3. Maple's OpenAPI example taught a wrong id — FIXED

`openapi.json` said `"Account id, e.g. acct_checking"`; the seeded ids are `acc_*` (`src/server/seed.ts:16`). Live on a real dev server:

```
GET /maple/api/accounts/acct_checking   <- what openapi.json taught -> HTTP 404
     {"error":{"message":"Account not found","code":"not_found"}}
GET /maple/api/accounts/acc_checking    <- the real seeded id       -> HTTP 200
     {"data":{"id":"acc_checking","name":"Maple Checking","kind":"checking",…}}
```

The same `acct_` typo was the fallback account id in `src/server/transfers.ts:26` and `src/server/orders.ts:31` — both fixed.

`.vendo/tools.json` **is** checked in and **is** regenerated by the example's own `predev`/`prebuild` hook (`vendo sync . --no-ai`). Ran it: the corrected description carried through plus the spec's `srcHash` (the spec is every OpenAPI tool's source file). Reported `tools: +0 -0 ~0` — nothing added, removed, or changed.

Worth noting the AI judge had **already caught this** and its note is still in `.vendo/judgments.json`: *"The real seeded ids are acc_checking, … (NOT acct_*)"*.

---

### 4. Maple's knowledge corpus documented an API that doesn't exist — FIXED

`src/vendo/knowledge.ts` claimed `GET /api/accounts` returns `currentBalanceCents` / `availableBalanceCents` / `type` / `currency` and supports `?type=`. Read the handler first (`src/app/api/accounts/route.ts` is literally `GET() { return ok(listAccounts()) }` — it reads no params), then confirmed live:

```
envelope : ["data"]
fields   : id, name, kind, mask, balance, accountNumber, routingNumber, sparkline
ids      : acc_checking, acc_savings, acc_savings_joint, acc_credit, acc_investing, acc_business, acc_moneymarket
kinds    : checking, savings, credit, investing
knowledge.ts claimed fields present? NONE of currentBalanceCents/availableBalanceCents/type/currency
?type=checking honored? -> 7 of 7 accounts -> filter IGNORED (no such param)
```

Rewritten to the real contract, including the `{ data: [...] }` envelope and "narrow by `kind` yourself".

The **"Available balance"** glossary entry described a concept Maple doesn't implement. It now keeps the definition (it's a real banking term a user may ask about) and then says Maple doesn't report one: an account carries a single `balance`, so there is no separate figure to quote.

## Why

Three of these make the agent confidently wrong: two teach it field names and ids that don't exist, and the third lets a path-mounted host ship with a silently dead tool surface. The fourth was a stale comment that made a correct, spec-mandated design read as a bug — worth correcting so the next reader doesn't "fix" it back into #914.

## Verification

- [x] `pnpm build && pnpm test:affected && pnpm typecheck && pnpm lint` on the touched scope — **build 21/21**, **typecheck 42/42**, **lint 0 errors** (4 pre-existing warnings, untouched files)
- Tests on touched packages: `@vendoai/actions` **620 passed**, `@vendoai/vendo` **2068 passed**, `demo-bank` **117 passed**
- Both new tests proven able to fail by reverting their fix (details per defect above)
- Live-verified against a real Maple dev server for defects 3 and 4 (transcripts above)
- Full suite is CI's job — the green `ci` check is the gate of record

**Two pre-existing failures, not from this branch:** `tests/mcp-byo-docs.docs.test.ts` and `tests/your-agent-docs.docs.test.ts` both assert a `"Bring your own agent"` group in `docs-site/docs.json`, which the docs reorg in #1133 (current `main` HEAD) renamed. Confirmed by stashing this branch's changes and re-running on clean `main` — same two failures. Left alone as out of scope.

- [ ] Screenshots attached (if UI-affecting) — n/a, no UI surface touched

## Not in scope

The runtime honesty problem (model asserting facts after failed tool calls), `genbench/`, and `packages/harnesses/src/vendo/` are untouched.

---
## Summary by cubic
Fixes `vendo doctor` so it flags a missing `VENDO_BASE_URL` when the spec uses a relative server mount, preventing 404s on mounted hosts. Also corrects Maple demo docs and examples to match the real API and clarifies the OpenAPI mount behavior comment.

- **Bug Fixes**
  - `@vendoai/vendo`: `E-CFG-003` now fires when `servers[0].url` is relative and `VENDO_BASE_URL` is unset; added a doctor test for this posture.
  - `@vendoai/actions`: clarified `openApiMountPath` comment to reflect prefix-free stored paths and single-prefix joining via `VENDO_BASE_URL`; added an end-to-end seam test that shows URLs are prefixed exactly once and the one-prefix-short hazard when the base URL drops the mount.
  - Maple demo: corrected account id example to `acc_*`; fixed fallback ids in transfers/orders; updated `GET /api/accounts` knowledge to the real contract (envelope, fields, no `?type=`) and noted Maple exposes only a single `balance`; regenerated `.vendo/tools.json`.

<sup>Written for commit ad44b5512026b93736f1221e40563179b3dbdb3a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1136?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/doctor-catches-an-unset-mount-base-url.md
+++ b/.changeset/doctor-catches-an-unset-mount-base-url.md
@@ -1,0 +1,13 @@
+---
+"@vendoai/vendo": patch
+---
+
+`vendo doctor`'s mount-agreement check (E-CFG-003) now fires when the OpenAPI
+spec declares a relative `servers[0].url` and `VENDO_BASE_URL` is unset.
+
+It used to return early in exactly that case, so the check was silent in the one
+posture that breaks. With no base URL the wire learns the bare request ORIGIN
+(`onRequestOrigin`) and stored binding paths are prefix-free by law (spec
+2026-08-06 §B1), so a path-mounted host serves every host tool one prefix short
+of the real endpoint: every page renders and every tool call 404s. The existing
+disagree/agree branches and the error code are unchanged.
--- a/examples/demo-bank/.vendo/tools.json
+++ b/examples/demo-bank/.vendo/tools.json
@@ -207,7 +207,7 @@
       },
       "inputSchemaSource": "declared",
       "outputSchemaSource": "declared",
-      "srcHash": "sha256:b0e1443e377bd2e13d505fb33febab4b1c62285700fa0ccaa3993fc063f265bb"
+      "srcHash": "sha256:040140ea40d1985cef514df49fe6a915c394006acab143a09b0216dbbdc62b65"
     },
     {
       "name": "host_demo_reset_create",
@@ -236,7 +236,7 @@
         "properties": {
           "id": {
             "type": "string",
-            "description": "Account id, e.g. acct_checking"
+            "description": "Account id, e.g. acc_checking"
           }
         },
         "required": [
@@ -311,7 +311,7 @@
       },
       "inputSchemaSource": "declared",
       "outputSchemaSource": "declared",
-      "srcHash": "sha256:b0e1443e377bd2e13d505fb33febab4b1c62285700fa0ccaa3993fc063f265bb"
+      "srcHash": "sha256:040140ea40d1985cef514df49fe6a915c394006acab143a09b0216dbbdc62b65"
     },
     {
       "name": "host_getBudgets",
@@ -373,7 +373,7 @@
       },
       "inputSchemaSource": "declared",
       "outputSchemaSource": "declared",
-      "srcHash": "sha256:b0e1443e377bd2e13d505fb33febab4b1c62285700fa0ccaa3993fc063f265bb"
+      "srcHash": "sha256:040140ea40d1985cef514df49fe6a915c394006acab143a09b0216dbbdc62b65"
     },
     {
       "name": "host_getCashflowInsights",
@@ -425,7 +425,7 @@
       },
       "inputSchemaSource": "declared",
       "outputSchemaSource": "declared",
-      "srcHash": "sha256:b0e1443e377bd2e13d505fb33febab4b1c62285700fa0ccaa3993fc063f265bb"
+      "srcHash": "sha256:040140ea40d1985cef514df49fe6a915c394006acab143a09b0216dbbdc62b65"
     },
     {
       "name": "host_getProfile",
@@ -504,7 +504,7 @@
       },
       "inputSchemaSource": "declared",
       "outputSchemaSource": "declared",
-      "srcHash": "sha256:b0e1443e377bd2e13d505fb33febab4b1c62285700fa0ccaa3993fc063f265bb"
+      "srcHash": "sha256:040140ea40d1985cef514df49fe6a915c394006acab143a09b0216dbbdc62b65"
     },
     {
       "name": "host_getRecurringInsights",
@@ -582,7 +582,7 @@
       },
       "inputSchemaSource": "declared",
       "outputSchemaSource": "declared",
-      "srcHash": "sha256:b0e1443e377bd2e13d505fb33febab4b1c62285700fa0ccaa3993fc063f265bb"
+      "srcHash": "sha256:040140ea40d1985cef514df49fe6a915c394006acab143a09b0216dbbdc62b65"
     },
     {
       "name": "host_getSpendingInsights",
@@ -641,7 +641,7 @@
       },
       "inputSchemaSource": "declared",
       "outputSchemaSource": "declared",
-      "srcHash": "sha256:b0e1443e377bd2e13d505fb33febab4b1c62285700fa0ccaa3993fc063f265bb"
+      "srcHash": "sha256:040140ea40d1985cef514df49fe6a915c394006acab143a09b0216dbbdc62b65"
     },
     {
       "name": "host_getTransaction",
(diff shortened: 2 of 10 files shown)


yousefh409:
### CI: the two failures are `main`'s, not this branch's

`ci` is red on exactly two tests, and **`main` is red on the same two right now**:

| | this PR (run [31371315852](https://github.com/runvendo/vendo/actions/runs/31371315852)) | `main` @ `fb867d903` (run [31371240744](https://github.com/runvendo/vendo/actions/runs/31371240744)) |
| --- | --- | --- |
| `test-shards (vendo-3)` | fail | fail |
| `test-shards (vendo-4)` | fail | fail |

Identical assertion in both, in both runs:

```
AssertionError: the 'Bring your own agent' group must exist: expected undefined to be defined
  tests/mcp-byo-docs.docs.test.ts   (vendo-3)
  tests/your-agent-docs.docs.test.ts (vendo-4)
```

Both read `docs-site/docs.json` and expect a `"Bring your own agent"` nav group that the docs reorg in #1133 renamed. Also reproduced locally by stashing this branch's changes and re-running on a clean `origin/main` checkout — same two failures.

Everything else is green, including both jobs that cover this diff: `test-rest (group-a … @vendoai/actions)` and `test-shards (vendo-1, vendo-2)`, plus `typecheck-lint`, `integration`, `conformance`, `audit`.

Left alone as out of scope — fixing those two stale docs assertions is `main`'s repair, not this branch's.

yousefh409:
### Resolved: those two failures were `main`'s, and `main` is fixed

The two stale docs-nav assertions are repaired on `main` by #1138 (merged as `46cd362`) — they now name the groups #1133 actually shipped. This branch was rebased onto that `main` (`ad44b55`), and CI re-ran fully green: `ci`, `integration`, `conformance`, `audit`, all 9 shards, both `test-rest` groups, `typecheck-lint`. Greptile re-reviewed at 5/5 with no files needing attention. Nothing in this diff changed.

PR #1137, opened 2026-08-10
fix(apps): a remix survives its own edit door

yousefh409:
Three defects in the remix path. Each destroys work the person can see on screen, and each returns **200**.

## 1. The island gate ran over seeded seats (`server/checking/floor.ts`)

The gate's rules are the **generated-island contract** — no imports, ambient Kit only, law 1 "a hand-typed constant feeding displayed math is invented data". Captured **host** source cannot satisfy that by construction. The demo host's own capture blocks on `pad = 6`, which is SVG chart padding:

```
island "PinnedNetWorthView965099a8" computes displayed values from the hand-typed
constant pad = 6 — a constant feeding displayed math is invented data (law 1)
```

`seed.from` writes its row directly with no floor, so **the document is born un-admittable**. Any later edit runs the floor, gets the unsatisfiable blocks handed back verbatim as repair instructions, and the only edit that clears them is to stop rendering the island: the fork is replaced by plain host components, the bundle orphaned, the app renamed. The card still renders and the ✦ pill still says "Remixed", so the person cannot tell.

The filter is restored **by NAME, not by `origin`** — load-bearing, and commented as such. The floor also runs over a compiled `app.vendo`, whose components are bare source strings; `bundleOf` reads those as `origin: "authored"`, so an origin-based skip would silently do nothing on the exact path that matters.

Nothing real is lost: both seed doors already check a capture has a default export themselves (`generation/engine.ts:174`, `remix/seed-surface.ts:180`).

## 2. A fork appended no history entry (`server/remix/seed-surface.ts`)

`seedFrom` calls `deps.apps.put(...)` directly instead of going through `persistEdit`, so it is the only create in the runtime that bypasses the one document write — and a fresh remix arrived with no history at all. Beyond the missing UI row: `serveDocFor` (`remix/review.ts:154`) falls back to `history.documents()` for the newest approved snapshot, so an empty history makes a review-kind remix **fail closed to pending** the moment its current version stops being approved.

## 3. The file save deleted a legacy remix's source (`server/doors/write-surface.ts`)

The seeded-bundle carry-forward was keyed on `origin === "seeded"`. But `componentEntrySchema` still accepts a bare source string, which is how **every remix forked before the seeded bundle existed** is stored — `bundleOf` reads it as `"authored"`, the carry does not fire, and a file-authored save that omits the component **deletes the remix outright**. Same one-token cure: key it on `isSeedComponentName(name)`.

## The old proof test was false, and is rewritten rather than deleted

`packages/apps/tests/seed.test.ts` — *"REFUSES a seeded bundle that violates a component-bundle rule"* — asserted the behaviour this PR removes. It is a proof of a premise that is empirically false, and **its fixture is a three-line component; that toy is the reason the change shipped green**.

It now runs the host's **real** capture (`examples/demo-bank/.vendo/remixable/NetWorthView.json`, 10681 bytes of source) and asserts what is actually true:

- a seeded seat holding that capture is **admitted**;
- the **same source, same `origin: "seeded"` tag**, under an ordinary name is still **gated** — proving the exemption is the name, not the tag;
- a capture with no default export is **refused at the seed door itself**.

## The seam test

`packages/apps/tests/seed-survives-edit.test.ts` drives the real `seed.from` into the real `floorChecks` with nothing stubbed on either side, and asserts the invariant the regression broke: **whatever the seed door mints, the checking floor must admit.**

Red on current main:

```
 ❯ tests/seed-survives-edit.test.ts (2 tests | 2 failed) 716ms
   × mints a document its own checking floor admits 700ms
     → expected [ { severity: 'block', …(2) }, …(1) ] to deeply equal []
   × records the version that says where the app came from 14ms
     → expected [] to have a length of 1 but got +0
```

Green with the fix:

```
 ✓ tests/seed-survives-edit.test.ts (2 tests) 618ms
 Test Files  1 passed (1)
      Tests  2 passed (2)
```

Defect 3 has its own case in `authored.test.ts` ("keeps a LEGACY remix's bare-string source"), verified red against the origin-keyed carry (`Cannot read properties of undefined (reading 'PinnedDashboardHeader22c8fbd4')`) and green with the fix.

## Gate

- `turbo run build typecheck --filter=@vendoai/apps --filter=@vendoai/ui --filter=@vendoai/vendo` — 16/16
- `packages/apps` full suite — 1520 passed, 18 skipped, 113 files
- `packages/vendo` seed/review/in-client/drift fixtures — 10 passed
- **root** `pnpm typecheck` — 42/42
- `pnpm lint` — 0 errors

No UI change, so no screenshots.

---
## Summary by cubic
Fixes three bugs that caused remixed apps to lose work after edits while still returning 200. Seeded seats now skip the island gate by name, `seed.from` writes a history version, and legacy remixes keep their source on save.

- **Bug Fixes**
  - Skip island checks for seeded seats by `isSeedComponentName` (not `origin`) to avoid blocking captured host code; seed doors still reject captures with no default export.
  - `seed.from` appends a history entry so remixes have provenance and review flows don’t fail closed.
  - File-save carry-forward keys on name to preserve remixes stored as bare source strings; prevents accidental deletion when omitted from file edits.

- **Refactors**
  - Added `seed-survives-edit.test.ts`, rewrote `seed.test.ts` to use the real capture, and added a legacy-source case in `authored.test.ts`.

<sup>Written for commit 6069b1e4f7bdfddd8cca8b01472a8486bb58c583. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1137?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/apps-remix-survives-its-own-edit.md
+++ b/.changeset/apps-remix-survives-its-own-edit.md
@@ -1,0 +1,36 @@
+---
+"@vendoai/apps": patch
+---
+
+A remix survives its own edit door. Three defects, all of which destroy work a
+person can see on screen, all of which returned 200.
+
+**The island gate no longer runs over a seeded seat.** Its rules are the
+GENERATED-island contract — no imports, ambient Kit only, no hand-typed constant
+feeding displayed math — and captured host source cannot satisfy them by
+construction (the demo host's own capture blocks on `pad = 6`, SVG chart
+padding). `seed.from` writes its row without the floor, so the document was born
+un-admittable: the next edit ran the floor, got the unsatisfiable blocks back
+verbatim as repair instructions, and the only edit that cleared them was to stop
+rendering the island. The person's fork was replaced by plain host components,
+the bundle orphaned and the app renamed, while the card still rendered and the
+✦ pill still said "Remixed". A capture with no default export is still refused,
+by the seed doors themselves, where it always was.
+
+**A fork records the version that says where it came from.** `seed.from` is the
+one create that does not go through `persistEdit`, so it now appends its own
+history entry. Without it a remix arrived with no history at all — and a
+review-kind remix failed closed to pending the moment its current version
+stopped being approved, because the reviewer's serve path reads the newest
+approved snapshot out of history.
+
+**A file save no longer deletes a legacy remix's source.** The save's
+seeded-bundle carry-forward was keyed on `origin === "seeded"`, but a component
+entry may be a bare source string — the shape every remix forked before the
+seeded bundle existed is stored in — and those read back as `authored`, so the
+carry never fired and a save that omitted the component dropped the remix
+outright.
+
+All three now key on the component's NAME (`isSeedComponentName`). The origin
+cannot carry this: a compiled `app.vendo` prints its components as bare source
+strings, so an origin test reads `authored` on exactly the path that matters.
--- a/packages/apps/src/contract/seed.ts
+++ b/packages/apps/src/contract/seed.ts
@@ -199,9 +199,11 @@ export const seedForkSource = (source: string): string => {
  * components but not its `seed` — a checkout prints seeded sources into
  * `app.vendo` alongside the model's islands.
  *
- * It does NOT buy an admission exemption. Seeded source used to skip the island
- * gate entirely; it now faces the same floor as anything else, because a bundle
- * a host captured is still a bundle this runtime is about to render.
+ * The NAME is the signal every one of those seams has to use — a bare source
+ * string reads back as `origin: "authored"`, so `bundleOf` cannot tell a printed
+ * seeded seat from a model island. The island gate's exemption
+ * (`server/checking/floor.ts`) and the file save's carry-forward
+ * (`server/doors/write-surface.ts`) both key on it for that reason.
  */
 export const isSeedComponentName = (name: string): boolean =>
   /^Pinned[A-Za-z0-9]*[0-9a-f]{8}$/.test(name);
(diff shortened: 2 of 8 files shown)


yousefh409:
**CI note — the two red shards are main's, not this PR's.**

`test-shards (vendo-3)` and `test-shards (vendo-4)` fail on `tests/mcp-byo-docs.docs.test.ts` and `tests/your-agent-docs.docs.test.ts` with the same assertion:

```
AssertionError: the 'Bring your own agent' group must exist: expected undefined to be defined
```

The group is gone from `docs-site/docs.json` — the docs-site reorg (#1133) removed it and left the two nav tests behind. The CI run for `main` at the exact SHA this branch forked from (`fb867d90`) fails on the identical two shards, plus `coverage-merge` and `ci`. Nothing in this PR touches `docs-site` or `packages/vendo/tests`. That breakage needs its own change; `ci` cannot go green here until it lands.

yousefh409:
Resolved — #1138 fixed the docs-nav gates on main. Rebased onto `46cd362ff` (clean, no conflicts; that PR touched only the two docs-nav test files). New head `6069b1e4f`. All checks green on the rebased SHA, including `ci`, `integration`, `conformance`, `audit` and `coverage-merge`. Auto-merge is not enabled.

Issue #867, opened 2026-08-06
demo-bank login: returnTo double-prefixes the base path (/maple/maple/...)


yousefh409:
Found during the MCP door baseline proof (lane D, 2026-08-06).

**Bug:** `examples/demo-bank/src/app/login/route.ts:111-118` — `safeReturnTo` keeps the `/maple` prefix already present in the door's returnTo, then `withBasePath` prepends it again, producing `/maple/maple/...` → 404 after login during the door's OAuth bounce.

**Repro:** same harness as the auth-preset base-path issue (door under /maple locally); the login seam 404s on the post-login redirect. With the double-prefix fixed the full chain passes end to end.

Also noting for whoever picks this up: on a fresh worktree `pnpm build` can hit the turbo cache and skip regenerating the gitignored `packages/ui/src/tree/jail/runtime-bundle.gen.ts`, so demo-bank dev 500s on every page — `pnpm --filter @vendoai/ui build:jail` fixes it. Cache-output question, filed here as a breadcrumb.

yousefh409:
Found again during a fresh-eyes verification walk on `main @ ce98c5460` (2026-08-06) — same bug as the breadcrumb above, with a fuller trace confirming the mechanism.

A `pnpm build` from a warm turbo cache (23/23 tasks, all cache hits) replays the `@vendoai/ui:build` log verbatim, including the `build:jail` step:

```
@vendoai/ui:build: cache hit, replaying logs 4cce0a962e65cbbf
@vendoai/ui:build: $ pnpm run build:jail
@vendoai/ui:build: $ node scripts/build-jail-runtime.mjs
@vendoai/ui:build: vite v6.4.3 building for production...
@vendoai/ui:build: ✓ built in 5.87s
@vendoai/ui:build: [ui] generated jail runtime (942113 bytes)
...
 Tasks:    23 successful, 23 total
Cached:    23 cached, 23 total
  Time:    3.364s >>> FULL TURBO
```

but the gitignored `packages/ui/src/tree/jail/runtime-bundle.gen.ts` is not restored — replaying logs isn't the same as replaying outputs, so this looks like a turbo `outputs` declaration gap for the generated file on the `build:jail`/`build` task. Starting `demo-bank dev` right after this build 500s immediately, on every request that touches the jail:

```
⨯ ./packages/ui/src/tree/jail/JailedComponent.tsx:13:1
Module not found: Can't resolve './runtime-bundle.gen.js'
[...]

Issue #866, opened 2026-08-06
auth presets: /login redirect and resource_metadata URL discard the base path in VENDO_BASE_URL


yousefh409:
Found during the MCP door baseline proof (lane D, 2026-08-06), running demo-bank locally with the door mounted under a path prefix (/maple).

**Bug:** `packages/vendo/src/auth-presets/identity.ts:139` builds the login redirect with `new URL("/login", publicOrigin())`, which discards any path component of `VENDO_BASE_URL`. A host mounted under a prefix (e.g. `https://site.com/maple`) redirects to `https://site.com/login` → 404. Shared by every preset (auth-js, auth0, clerk, jwt, supabase). Same class: the 401's `WWW-Authenticate` `resource_metadata` URL also 404s under a prefix.

**Repro:** `pnpm --filter demo-bank dev`, `VENDO_BASE_URL=http://localhost:3123/maple`, drive `examples/mcp-agent/agent.mjs` → host-login seam 404s. With the URL joined against the full base (not the bare origin), the whole chain passes: DCR → consent → token → tools/list → vendo_make → receipt → wire read-back.

Not sprint-blocking: broker-mode deployments (demos.vendo.run/maple) are unaffected. Related but separate: the deployed demo's protected-resource metadata advertises the internal Railway hostname without /maple — worth checking with it.

PR #1139, opened 2026-08-10
docs(mcp): setup page leads with the agent prompt; reference and playbook split

yousefh409:
## The problem

`/existing-agents/mcp` was a setup page in name only. Someone arriving to *open
the door* had to read past option tables, the `HostOAuthAdapter` contract, wire
formats, and federation claims before reaching anything they could act on. The
reference material and the setup path were interleaved in one page, so neither
job was served: no one could set the door up quickly, and no one could look a
field up quickly either.

## The approved shape

**Setup page (`/existing-agents/mcp`) — five sections, prompt-first.**

1. **What you get** — the one URL, and the experimental status note.
2. **Set up the door (once)** — the `AgentPrompt` card leads. The prompt itself
   is one line that points the agent at `vendo.run/agents.md`, so the actual
   procedure stays centralized in the playbook instead of being duplicated
   here. A short **By hand** block follows for people who want the three things
   themselves (`mcp: true` + auth preset, the `.well-known` route,
   `VENDO_BASE_URL`).
3. **Connect — pick your client** — Claude Code, any MCP client, service key.
4. **See it work** — what `vendo_make` returns and where the screen lands.
5. **Next** — three cards out to reference, guard, and production.

**Playbook (`/agents`) — the procedure lives here.** A new `## MCP door`
section carries the five ordered steps, additive to the install flow, plus a
pointer at the top of the page for agents told to open the MCP door.
`/agents/verify` now links back to it so a doctor failure leads to the wiring.

**Reference (`/capabilities/mcp`) — pure reference.** Setup prose removed; it
links out to the setup page and stays a lookup surface.

## Source-verified corrections made along the way

These are behavior fixes, not just reorganization — each read off the source,
not the old prose:

- **All five auth presets carry the OAuth adapter.** `authJs`, `clerk`,
  `supabase`, `auth0`, and `jwt({ secret })` each fill the door's `oauth` seam
  along with `principal` and `actAs`. A separately passed `oauth` is *ignored*
  whenever `auth` is set — the preset owns the seam.
- **`--auth none` cannot open the door.** It writes a demo `principal` and no
  oauth seam, and `createVendo` throws at composition when `mcp: true` has no
  `HostOAuthAdapter`. The old text implied any auth choice would do.
- **The `serviceAuth` no-op nuance was wrong, and now distinguishes explicit
  from declared.** `mcp.serviceAuth` passed alongside an *explicit*
  `mcp.remoteAs` warns at composition and does nothing (that door serves no
  `/token` of its own). But a *declared* `VENDO_MCP_BROKER_URL` is a default —
  it leaves an explicit `serviceAuth` alone rather than overriding it. Those two
  cases were previously collapsed into one blanket claim.
- **Broker mode details confirmed against `vendo-web` `origin/main`.** Broker
  mode is declared, never discovered (origin becomes issuer, URL becomes
  expected audience, door stops serving `/authorize`, `/token`, `/register`).
  The broker forwards only to a public HTTPS address, so it can never reach
  `localhost`. Its exchange answers field for field like the door's, except
  `access_token` is a signed JWT rather than an opaque string.

## Screenshots

Local Mintlify dev server, real browser.

**Setup page, top to bottom** — the five sections in order:

Setup page top to bottom

**Setup page, top of viewport:**

Setup page top

**The `AgentPrompt` card up close** — it now leads the setup section:

AgentPrompt card

**Reference page (`/capabilities/mcp`), top** — pure reference, links out to setup:

Reference page top

**Playbook MCP-door section (`/agents#mcp-door`)** — the five ordered steps:

Playbook MCP door section

## Orphaned images

Dropping the old step-by-step walkthrough leaves 4 images unreferenced under
`docs-site/images/existing-agents/`:

- `mcp-walk-approval.png`
- `mcp-walk-other-user.png`
- `mcp-walk-pinned.png`
- `mcp-walk-slot-empty.png`

They are **left in place deliberately** — not deleted in this PR. (The fifth,
`mcp-walk-slot-filled.png`, is still used by `/existing-agents/your-agent`.)

## Sequencing

S3 has merged (#1142, plus the five-question init in #1146), this branch is
rebased over it, and the PR is ready. Both surfaces now route through
`vendo init --use-case mcp`, and the by-hand seams are demoted to the no-init
path (Express, `--auth none`, `--auth jwt`) — the stale "init does not write
it" claims are gone. One nuance: that flag documents `main`, not the current
release — npm `latest` is 0.10.0 and ships neither `--use-case` nor `planMcp`,
so the pages are accurate the day the next release cuts, exactly as
`quickstart.mdx` already documents `--use-case` on main today.

## Gates

`pnpm build`, `pnpm test:affected`, `pnpm typecheck`, `pnpm lint` all green, plus
`packages/vendo` `tests/cli/doctor-codes.docs.test.ts` (the test that reads the
playbook) green. `mint broken-links` from `docs-site/`: no broken links.

---
## Summary by cubic
Reorganized MCP docs to make setup fast and clear: the setup page now leads with an agent prompt and concise steps, while reference and playbook content are split for lookup. Also updated the agent playbook install steps to install both `vendoai` and `@vendoai/vendo`, preventing module resolution errors and noisy doctor failures.

- **Refactors**
  - Split docs into three surfaces: setup (`/existing-agents/mcp`, prompt‑first), playbook (`/agents#mcp-door` procedure), and reference (`/capabilities/mcp`, pure lookup).
  - Setup now shows the minimal wiring: `mcp: true` plus an auth preset (or, for a login‑less repo, `principal` + `oauth.session/principal` with a fixed subject and no `actAs`), origin‑root discovery via `wellKnownVendoHandler`, and `VENDO_BASE_URL` requirements. Connect guidance covers Claude Code, any MCP client, and the service‑key path; linked the themed connect page (`{mount}/connect`).
  - Playbook adds the ordered MCP steps, including the login‑less path (leave `actAs` out to avoid doctor failures) and a tighter verify gate; `verify` links back to it. Host auth notes that a principal‑only stub needs an `oauth` seam beside it, not a preset.
  - Install flow: install BOTH `vendoai` and `@vendoai/vendo` (init imports subpaths under `@vendoai/vendo`; the alias alone 500s with “Module not found”). Keep the version check, and note that `vendo init` auto‑uses a `VENDO_API_KEY` already in `.env.local` with no login or key minting.
  - AgentPrompt now accepts a `title`; the setup page leads with a custom title (“Set up with your coding agent”), dropping the “Or install” phrasing.

- **Bug Fixes**
  - Clarified auth behavior: all presets carry the OAuth adapter; mixing an `auth` preset with `oauth`/per‑seam config is ignored or throws at composition. `--auth none` cannot open the door.
  - Documented broker mode as declared, not discovered: set `VENDO_MCP_BROKER_URL` (origin becomes issuer, URL becomes audience); explicit `mcp.remoteAs` wins; `mcp.serviceAuth` is local‑only and ignored when `remoteAs` is explicit. Broker forwards only to public HTTPS; its exchange mirrors the door’s except `access_token` is a JWT.
  - Expanded `VENDO_BASE_URL` duties behind proxies and under path prefixes; noted known prefix issues (#866, #867). Added note that approving a parked call does not resume it — the client must retry on the same MCP session.

<sup>Written for commit 548dcbf2e727ea6b2aade79b59efb48107a47324. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1139?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs-site/agents/host-auth.mdx
+++ b/docs-site/agents/host-auth.mdx
@@ -62,6 +62,8 @@ changes nothing about auth; you add the one line yourself.
   chat routes, so apps and approvals created through your agent loop stay
   visible to the embeds (which call the wire route directly). Replace both
   sides with your real session lookup, or a preset line, before production.
+  Opening the MCP door on top of this principal needs the `oauth` and `actAs`
+  seams beside it, not a preset: [the door section](/agents#mcp-door).
 
 ## Per-preset wiring
 
--- a/docs-site/agents/host-auth.mdx
+++ b/docs-site/agents/host-auth.mdx
@@ -62,8 +62,8 @@ changes nothing about auth; you add the one line yourself.
   chat routes, so apps and approvals created through your agent loop stay
   visible to the embeds (which call the wire route directly). Replace both
   sides with your real session lookup, or a preset line, before production.
-  Opening the MCP door on top of this principal needs the `oauth` and `actAs`
-  seams beside it, not a preset: [the door section](/agents#mcp-door).
+  Opening the MCP door on top of this principal needs an `oauth` seam beside
+  it, not a preset: [the door section](/agents#mcp-door).
 
 ## Per-preset wiring
 
--- a/docs-site/agents/index.mdx
+++ b/docs-site/agents/index.mdx
@@ -23,6 +23,11 @@ minute of latency and a chunk of context you will not use, and it is why
 installs feel slow to start. Read the repo's `package.json` before deep
 directory scans. Start acting after step 1, not after reading everything.
 
+**Asked for the MCP door?** That is outside agents — Claude, ChatGPT, Cursor —
+acting in the host's product through its own guard, riding the `createVendo`
+composition that both step-1 paths write. It needs the same wiring and visible
+surface as any other install: do the install below, then [MCP door](#mcp-door).
+
 ## Install flow
 
 1. **Check for an existing agent — and branch here.** If the host repo already runs its own
@@ -180,6 +185,72 @@ directory scans. Start acting after step 1, not after reading everything.
    from your data; a guarded write will render an approve/deny card inline."
    Name which of their agent's surfaces now renders Vendo output (the chat
    page, the overlay, an embed) so they know where to look.
+
+## MCP door
+
+Additive to the install flow above: take these in order once init has run, and
+skip none of its steps.
+
+1. **`mcp: true` plus the OAuth adapter.** Add `mcp: true` to the
+   `createVendo({...})` init wrote. The door mints its own principals through a
+   `HostOAuthAdapter`, and `createVendo` THROWS at composition without one.
+   Every named preset carries it, so `--auth authJs|clerk|supabase|auth0` (or
+   `jwt({ secret })`) is the whole answer; `--auth none` is NOT — it writes a
+   demo `principal` and no oauth seam, so the door cannot open. An `auth` preset
+   owns the seam: a separately passed `oauth` is ignored whenever `auth` is set.
+   Which preset: [host auth](/agents/host-auth).
+
+   ```ts
+   export const vendo = createVendo({
+     auth: authJs(), // supplies principal, actAs, and the door's oauth half
+     mcp: true,
+   });
+   ```
+
+2. **The discovery route — init does not write it.** The door's discovery
+   documents live at ORIGIN-ROOT paths, outside `/api/vendo`, so the catch-all
+   route never sees them and that directory needs a handler of its own:
+
+   ```ts
+   // app/.well-known/[...vendo]/route.ts (src/app when the repo uses it)
+   import { wellKnownVendoHandler } from "@vendoai/vendo/server";
+   import { vendo } from "@/vendo/server"; // wherever init put the composition
+
+   export const { GET, POST } = wellKnownVendoHandler(vendo);
+   ```
+
+   The handler answers only the door's own paths and 404s everything else under
+   `/.well-known`. Express and custom hosts instead mount the same composition
+   at the origin root as well (`app.use("/.well-known", mountVendo())`),
+   registered after the app's own well-known routes.
+
+3. **`VENDO_BASE_URL`.** Set it to the deployment's FULL public URL, path
+   prefix included. The issuer, every advertised endpoint, the
+   protected-resource identifier, and the RFC 8707 audience all derive from it;
+   left unset they come from the request URL — behind any proxy the
+   proxy-internal origin, so discovery advertises endpoints no client can reach.
+   Forwarded headers are never consulted.
+
+   ```bash
+   VENDO_BASE_URL=https://app.example.com
+   ```
+
+4. **`serviceAuth` — only when asked for.** Skip it unless the HOST's own
+   backend has to act for a user who is not at a browser (a nightly job, a
+   queue worker). `mcp: { serviceAuth: { keys: [...] } }` opens an RFC 8693
+   exchange at the door's own token endpoint: the backend posts a key plus one
+   of the host's user ids and gets back a short-lived token bound to that user.
+   The keys are opaque secrets your human generates — ask before creating one
+   (Rules of engagement below). Third-party agents need none of it; they run
+   the per-user OAuth the door already serves. Recipe:
+   [MCP door](/capabilities/mcp).
+
+5. **Verify — step 6's gate, nothing new.** Doctor reads the door's posture off
+   the running composition, so the dev server must be up for the MCP checks to
+   run at all. The registry checks grade `server.json` and the challenge only
+   when those files already exist — publishing to the MCP registry is a separate
+   job, not part of reaching green. Every `E-MCP` code with its exact fix:
+   [verify](/agents/verify#E-MCP-001).
 
 ## Rules of engagement
 
--- a/docs-site/agents/index.mdx
+++ b/docs-site/agents/index.mdx
@@ -211,14 +211,11 @@ skip none of its steps.
    ```
 
    **No login yet?** No preset applies — there is no session to read, and
-   installing a provider's SDK is not a login. Fill the three seams yourself,
+   installing a provider's SDK is not a login. Fill two seams yourself instead,
    never beside `auth` (one preset or the per-seam trio; mixing throws at
-   composition): keep the demo `principal` `--auth none` wrote, add an `oauth`
-   whose `session` hands the door that same subject instead of bouncing a
-   browser — the door still renders its own consent page — and an `actAs` with
-   empty headers, since a product with no login has no per-user credential to
-   mint. Every client then acts as that one person: a demo or an internal tool,
-   not a multi-user product.
+   composition): keep the demo `principal` `--auth none` wrote, and add an
+   `oauth` whose `session` hands the door that same subject instead of bouncing
+   a browser — the door still renders its own consent page.
 
    ```ts
    export const vendo = createVendo({
@@ -227,10 +224,18 @@ skip none of its steps.
        session: async () => ({ subject: "demo-user" }),
        principal: async (subject) => ({ kind: "user", subject }),
      },
-     actAs: async () => ({ headers: {} }),
      mcp: true,
    });
    ```
+
+   Leave `actAs` out. A stub returning empty headers is worse than an absent
+   seam: doctor then runs its live mint-and-verify round-trip and FAILS
+   [`E-AUTH-004`](/agents/verify#E-AUTH-004), because a fixed demo principal can
+   never echo the probe's own subject. Absent, doctor warns
+   [`E-AUTH-007`](/agents/verify#E-AUTH-007) and step 6's gate still passes — the
+   price is that your own route-bound tools answer `not-implemented` over the
+   door, while `vendo_make` and the apps tools work. Every client acts as that
+   one person: a demo or an internal tool, not a multi-user product.
 
 2. **The discovery route — init does not write it.** The door's discovery
    documents live at ORIGIN-ROOT paths, outside `/api/vendo`, so the catch-all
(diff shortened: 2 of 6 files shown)


yousefh409:
## Cold-walk proof of the new MCP setup page

Fresh-eyes run by an agent that built none of this. Bare `create-next-app` in `/tmp`, real
`claude -p` headless, real Vendo Cloud key, real MCP calls. **Verdict: the promise holds, but
one unstated precondition cost 10 of the 19 minutes.**

**Time to first success: 19 min 4 s** from paste to `vendo doctor --json` exit 0 — including one
human answer turn (the agent stopped at 3 min 35 s to ask about keys, correctly). Independently
re-verified afterwards: **exit 0, 26 checks, 0 failures, 1 warning.**

| Leg | Verdict |
| --- | --- |
| One prompt → MCP door wired | **PASS** |
| `vendo doctor --json` all green | **PASS**, but only via a workaround the docs don't contain |
| Service key → token exchange → real MCP tool call | **PASS**, first try, no surprises |
| Connect page in a real browser | **PASS** |

### Leg 1 — one prompt wired the door

The agent read the playbook, detected Next.js, installed `vendoai@0.9.0`, ran `vendo init --agent`
before the real `vendo init --yes --auth authJs --no-ai`, then added `mcp: true`, the
`app/.well-known/[...vendo]/route.ts` discovery route, and `VENDO_BASE_URL` — exactly the three
things the page's **By hand** section names. It stopped and asked before creating any account, key,
or secret, both times, per **Rules of engagement**. Nothing was invented outside the catalog.

### Leg 2 — green, but the docs walked me into a wall

A bare app has no login. `authJs()` with no login means every wire request is identity-less, so
`/status` returns 403 and four checks break:

```
broken  live/status    /status returned an invalid composition response (403)
broken  auth/present   present credential probe cannot run; start the dev server at … and retry
broken  auth/act-as    cannot probe actAs; start the dev server at … and retry
broken  turn/model     live model turn cannot run; start the dev server at … and retry
```

Three problems, in order of how much time they cost:

1. **The remedy text is wrong.** It says *"start the dev server … and retry"* — the dev server was
   up and answering. The real cause is the 403. The agent chased the wrong thing through
   `node_modules/@vendoai/vendo/dist/*.js` and four extra docs pages before working it out.
2. **No `mcp/*` check runs at all when `live/status` fails.** The failing report contains zero MCP
   checks, yet the setup page says *"gate on doctor, which checks both metadata documents resolve
   and the server card parses."* A healthy door and a broken door look identical here. (Once the
   403 cleared, all three — `mcp/protected-resource`, `mcp/authorization-server`,
   `mcp/server-card` — went green.)
3. **The precondition is never stated.** These two sentences pin you in with no way out:
   - `agents/index.mdx` — *"`--auth authJs|clerk|supabase|auth0` (or `jwt({ secret })`) is the
     whole answer; `--auth none` is NOT — it writes a demo `principal` and no oauth seam, so the
     door cannot open."*
   - `agents/host-auth.mdx` — *"`--auth none` is a valid ship for products without login."*

   Together: a product without login cannot open the door — and neither the setup page nor the
   playbook says so before you start. Installing `@auth/core` is not a login.

The agent unblocked itself by spreading the preset and overriding one seam, which appears on
neither page:

```ts
const base = authJs();
export const vendo = createVendo({
  auth: { ...base, principal: async (r) => (await base.principal(r)) ?? { kind: "user", subject: "demo-guest", ephemeral: true } },
  mcp: true,
});
```

doctor green

### Leg 3 — service key worked exactly as written

Followed only the setup page: `openssl rand -hex 32` into `.env`, `serviceAuth` on the door, then
the page's own curl. The response matched the page's documented JSON field for field
(`vmat_…`, `Bearer`, `600`, `"read write"`). Bearer into an ordinary MCP session → `initialize`
(2025-11-25) → **17 tools**. Then the page's own *See it work* example:

```
tools/call vendo_make {"request":"a simple welcome screen with a short greeting"}
→ {"id":"app_af9ab393…","title":"Welcome","status":"ready","say":"Welcome is on your screen."}
```

The exact four fields the page shows. **Seam checked both ways** — read back through
`vendo_apps_list` and the app is really there, so no side of this is a stub.

service key exchange and tool call

### Leg 4 — connect page

Real browser, no login needed to view: product name, the URL to paste, per-client steps for
Claude / ChatGPT / Cursor, an **Add to Cursor** button, and the JSON snippet.

connect page

### Two smaller docs nits

- **Step 4 has no option for "I already have a key."** It offers `--cloud-key <key>` or `--byo`,
  and never says init auto-detects a `VENDO_API_KEY` already sitting in `.env.local` — which it
  does (`packages/vendo/src/cli/init.ts:1110`). Finding that cost the agent a third page fetch
  (`/reference/cli`), breaking the playbook's own *"this page plus AT MOST ONE more"* rule.
- **"On Vendo Cloud, create the key on the project's keys page … instead"** reads as *"if you have
  a Cloud key."* The actual trigger for the brokered path is `VENDO_MCP_BROKER_URL`
  (`packages/vendo/src/compose-mcp.ts:97-99`). With a Cloud key and no broker URL the local
  `serviceAuth` path is correct — that's what ran here.

### Deviations from a true cold start

1. **The one that matters:** docs were served from local `mint dev`, where `.md` routes 404
   (`/agents.md`, `/llms.txt`, all of them). Production serves `/agents.md` fine — verified 200 —
   so the prompt pointed at `http://localhost:3799/agents` (HTML) instead. The agent stripped the
   text out of the HTML itself and **never fell back to the deployed docs**, so the new MCP section
   is what it actually followed. The playbook's *"if your fetch tool returned a summary, re-fetch it
   raw"* line is untested by this run.
2. My existing Cloud key was already in `.env.local` (as a real user with a key would have it).
   No account created, no key minted, nothing printed.
3. Environment noise, not product: the agent ran on this machine's real global agent config with
   MCP servers off, a local shell guard denied its `.env.local` read, and the app ran on **3080**
   because 3000 was occupied.

Artifacts: `/tmp/mcp-cold-proof/` (agent transcript, `doctor.json`, exchange + tool-call logs,
screenshots); app left in place at `/tmp/mcp-proof-app`.

yousefh409:
`AgentPrompt` now takes an optional `title` (default unchanged), and the MCP setup page uses "Set up with your coding agent" — the door is the install there, not an alternative to one.

**`/existing-agents/mcp` — new title**

MCP page card

**`/quickstart` — unchanged, no edit to that page**

Quickstart card

`npx mint broken-links` → `success no broken links found`.

yousefh409:
The walk, moving: paste → wired → doctor green → service-key exchange → tool call

Vendo MCP door end to end

One continuous `asciinema` recording of a real run — no splicing, no re-takes stitched together. Bare `create-next-app` in `/tmp`, real headless `claude -p`, real Vendo Cloud key, real MCP calls. Idle time capped at 8s for playback; every command and every response is exactly what ran.

**What the recording shows, in order**

| Leg | Result |
| --- | --- |
| Bare app → one pasted prompt → agent installs | **PASS** |
| Agent stops to ask before minting a key | **PASS** — the playbook's rule, honoured |
| `vendo doctor --json` | **PASS** — exit 0, 0 failures, 2 warnings |
| Service key → RFC 8693 exchange | **PASS** — `vmat_…`, `Bearer`, `600`, `"read write"`, field for field as the page documents |
| `initialize` → `tools/list` | **PASS** — 17 tools |
| `tools/call vendo_make` | **PASS** — the exact four fields from *See it work* |

The agent followed this branch's new login-less door path verbatim — demo `principal` plus the `oauth.session`/`oauth.principal` seams, and **`actAs` left out on purpose**, quoting the page's own reason (a stub fails doctor's live mint-and-verify, `E-AUTH-004`). That's the cold-run fix in these commits doing its job.

**Three findings from the run** (all reproduced on a rehearsal run beforehand, so 2/2 or 3/3):

1. **`vendo init` writes imports the playbook never tells you to install.** Init emits `import … from "@vendoai/vendo/server"`, but step 3 says `npm install vendoai` — the alias. Under pnpm's strict linking `@vendoai/vendo` is only in the virtual store, so every route 500s with `Module not found: Can't resolve '@vendoai/vendo/server'` and doctor reports four `live/*` + `auth/*` breakages that name none of it. Both agents diagnosed it and self-healed with `pnpm add @vendoai/vendo`, but it cost several minutes each time. **3/3 runs.** Worth either a playbook line or having init add the direct dep.
2. **The ask-gate can't see a key that's already there.** Step 4 says init auto-detects a `VENDO_API_KEY` sitting in `.env.local`, but nothing tells the agent to *look* before opening the "ask before creating a key" gate — so a one-shot headless run always halts, even when no key needs minting. That's the pause you see in the GIF.
3. **`vendo doctor --help` errors** (`unknown option: --help`) — the agent had to run bare `vendo doctor` to discover `--url`.

**Deviations from a true cold start** — same two as the earlier cold-walk, unchanged:

- Docs served from local `mint dev`, where `.md` routes 404 (`/agents.md`, `/llms.txt`). The prompt pointed at `http://localhost:3796/agents` (HTML); the agent stripped the text out of the SSR'd markup itself and never fell back to deployed docs, so **this branch's content is what it actually followed**. Costs the agent ~2 min of extraction it wouldn't pay against production.
- My existing Cloud key was pre-seeded in `.env.local` (as a real user with a key would have it). No account created, no key minted, nothing printed. `PORT=3210` was pinned because 3000–3002 are held by other dev servers on this box.

The service key is minted by the page's own `openssl rand -hex 32` command (which never prints it) and read into `$SERVICE_KEY` — the recorded curl shows the variable, never the value. Both keys were scrubbed from the machine afterwards; the cast and GIF were scanned for either value and are clean.

Artifacts: `.cast` alongside the GIF in `/tmp/mcp-gif/`.

PR #1161, opened 2026-08-10
feat(apps): the agent's data tools move onto appData — one user stops seeing another's rows

yousefh409:
`vendo_apps_data_list` / `_put` / `_delete` are how the embedded agent saves and reads an app's declared storage on the person's behalf. They landed in the generic `records` family, which has no answer to "whose row is this": every user of an app wrote into one flat collection, and the only thing between them was that nobody had asked.

Every one of those calls now carries `ctx.principal.subject` — the LIVE caller, off the run context, never off the tool args — into the owner-stamped `appData` family that S1 landed. `put` stamps the row, `list` ANDs the stamp into the query, `get` answers `null` for another owner's row and `delete` no-ops on one. A cross-user read is no longer forbidden; it is unexpressible. An id another owner already holds refuses with `conflict` rather than being taken over, and that refusal is surfaced honestly rather than swallowed. Declared file collections get the same treatment through the family's file twins.

`createAppData` takes `{ ops, store }` and its `records`/`blobs` take a required owner, so there is no way to reach a collection without naming whose drawer it is. `backingFor` is the one place that decides which store backs a collection, which leaves the guards exactly where they were and in the order they were: `declaredStorage` (with `state` still reserved) → `validateRecordRefs` → the 256 KB cap → put, and the 5 MB cap for blobs. `getState`/`setState` stay on the `StoreAdapter` façade over `vendo_state`, deliberately and permanently — that is out of scope here.

`AppsConfig` gains an optional `ops` slot the umbrella fills from the same `StoreOps` surface the deployment already selected. Its absence is a real answer, not a failure: `selectStoreOps` returns nothing for a store with neither its own ops nor a SQL handle, and requiring it would crash composition at boot for that store where today it refuses at the op that needed one.

## Tests

The isolation suite drops its adapter-only fake for core's reference `memoryStoreOps()` — the same implementation the conformance suite holds the real backends to — and adds five cases across the seam:

- cross-owner list blindness (both directions, on the id sets)
- a foreign `get` returns `null` while the owner's own `get` lands
- a foreign `delete` is a no-op while the owner's own `delete` works
- a spoofed `refs.subject` is refused by CODE (`validation`). The collection **declares** `subject` on purpose, so the runtime's own undeclared-ref guard is cleared and the refusal proved is the store's, not a shallower one that would pass whether or not the family stamps the owner.
- file-twin isolation across `get` and `list`

The three existing state cases are unchanged apart from the one-line factory call.

**Prove-it-can-fail:** hardcoding the owner in `createAppData`'s `records()` turned `lists only the caller's own rows when two owners share an app collection` RED with `expected [ 'note_grace', 'note_ada' ] to deeply equal [ 'note_ada' ]`, plus the foreign-`get` and foreign-`delete` cases. Restored, green.

## Known scope note

`clear` is owner-scoped, as specified — deleting an app empties the caller's drawer through the appData family. Rows belonging to other owners of the same app are reached by `lifecycle.erase({ appId })`, not by this path.

---
## Summary by cubic
Moves the agent’s data tools to owner-scoped `appData` so one user can’t read or overwrite another user’s rows. Calls now use the live caller from `ctx.principal.subject`, not tool args.

- **New Features**
  - `vendo_apps_data_list/_put/_delete` now go through `appData` with an owner: `put` stamps the subject, `list` scopes by subject, `get` returns null across owners, and `delete` is a no-op across owners.
  - If another owner already holds an id, `put` returns `conflict` instead of overwriting.
  - File collections get the same owner scoping as records.
  - Guards and limits are unchanged; app state stays on `vendo_state`.

- **Migration**
  - `createAppData` now takes `{ ops, store }`; pass your deployment’s `StoreOps` when available.
  - `records(app, name, owner)` and `blobs(app, name, owner)` now require an `owner`.
  - `AppsConfig` gains optional `ops`; composition wires it from the selected `StoreOps`.
  - No changes needed for app authors; runtime tools pass the owner automatically.

<sup>Written for commit 3f1fd1164d97075b56e35a47526677400c5ffe5f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1161?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/apps-data-tools-owner-stamped.md
+++ b/.changeset/apps-data-tools-owner-stamped.md
@@ -1,0 +1,31 @@
+---
+"@vendoai/apps": minor
+"@vendoai/vendo": minor
+---
+
+The agent's data tools move onto `appData` — one user can no longer see another's rows.
+
+`vendo_apps_data_list` / `_put` / `_delete` are how the embedded agent saves and
+reads an app's declared storage on the person's behalf. They landed in the
+generic `records` family, which has no answer to "whose row is this": every
+user of an app wrote into one flat collection, and the only thing between them
+was that nobody had asked.
+
+Now every one of those calls carries `ctx.principal.subject` — the LIVE caller,
+off the run context, never off the tool args — into the owner-stamped `appData`
+family. `put` stamps the row with that subject, `list` ANDs it into the query,
+`get` answers `null` for another owner's row and `delete` no-ops on one. A
+cross-user read is no longer forbidden; it is unexpressible. An id another owner
+already holds refuses with `conflict` rather than being taken over, and that
+refusal is surfaced honestly rather than swallowed. Declared file collections
+get the same treatment through the family's file twins.
+
+Nothing about what an app may declare changed. The guards keep their posts in
+the same order — the declaration check (with `state` still reserved), the
+declared-refs check, the 256 KB record cap, the 5 MB blob cap — and app state
+(`vendo_state`) stays on the `StoreAdapter` façade, deliberately.
+
+`AppsConfig` gains an optional `ops` slot that the umbrella fills with the same
+`StoreOps` surface the deployment already selected. Its absence is a real
+answer, not a failure: a store that offers neither its own ops nor a SQL handle
+keeps exactly today's behavior instead of crashing composition at boot.
--- a/packages/apps/src/server/doors/data-tools.ts
+++ b/packages/apps/src/server/doors/data-tools.ts
@@ -3,7 +3,9 @@
  * declared app data collection, over the same `requireOwned` gate every other
  * door reads.
  *
- * Lifted out of `createAgentTools` unchanged.
+ * Lifted out of `createAgentTools` unchanged. Every row is owned by the LIVE
+ * caller: the subject comes off the run context, never off the tool args, so
+ * generated code has no way to name someone else's drawer.
  */
 import {
   VendoError,
@@ -35,7 +37,9 @@ export const listAppData = async (
   };
   return {
     status: "ok",
-    output: await dependencies.data.records(app, args.collection as string).list(query) as unknown as Json,
+    output: await dependencies.data
+      .records(app, args.collection as string, ctx.principal.subject)
+      .list(query) as unknown as Json,
   };
 };
 
@@ -50,11 +54,13 @@ export const putAppData = async (
   }
   const app = await dependencies.requireOwned(args.appId as string, ctx);
   const refs = optionalRefs(args.refs);
-  const record = await dependencies.data.records(app, args.collection as string).put({
-    id: args.id as string,
-    data: args.data,
-    ...(refs === undefined ? {} : { refs }),
-  });
+  const record = await dependencies.data
+    .records(app, args.collection as string, ctx.principal.subject)
+    .put({
+      id: args.id as string,
+      data: args.data,
+      ...(refs === undefined ? {} : { refs }),
+    });
   return { status: "ok", output: record as unknown as Json };
 };
 
@@ -65,6 +71,8 @@ export const deleteAppData = async (
 ): Promise<ToolOutcome> => {
   const args = input(call.args, ["appId", "collection", "id"]);
   const app = await dependencies.requireOwned(args.appId as string, ctx);
-  await dependencies.data.records(app, args.collection as string).delete(args.id as string);
+  await dependencies.data
+    .records(app, args.collection as string, ctx.principal.subject)
+    .delete(args.id as string);
   return { status: "ok", output: { status: "ok" } };
 };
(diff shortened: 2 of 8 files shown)


yousefh409:
## S4 live two-user proof — BLOCKED, not run

Attempted the real-browser isolation proof for this PR (`eeebbeefe`) on the fleet VM. **The proof did not run: the demo host has no model credential on this machine, so every agent turn dies before a single tool call.** No claim about isolation is made here either way.

### What actually happened

Fresh detached worktree at `eeebbeefe`, `pnpm install` + `pnpm build` (21/21 green), one dev server: `MAPLE_STORE=local PORT=3577 pnpm --filter demo-bank dev`. Real Chromium (Playwright 1.61, chromium-1234), 1440×900, headless.

1. Signed in as **A — yousef@maple.com** at `/maple/login?email=…`, session real. ✅
2. Opened the embedded agent (⌘K) and sent: *"Build me a private notes app that saves notes I can come back to, then save a note that says exactly ADA-SECRET-1786385043"*.
3. Turn failed in ~2s: **"The response didn't finish — I couldn't make that request work; nothing was changed."**
4. Signed in as **B — mia@maple.com**, asked it to list the notes app including Yousef's. Same failure.

Server log (`/tmp/s4-live/devserver.log:70-73`):

```
[vendo] model: no model credential found
[vendo] Vendo found no model key. Set ANTHROPIC_API_KEY / OPENAI_API_KEY / …
Error [VendoError]: … at DevModelController.call (packages/vendo/dist/dev-creds/model.js:292)
```

No `ANTHROPIC_API_KEY`, no `VENDO_API_KEY`, no `.env.local` for `examples/demo-bank` anywhere on the box; the VM's secret guard denies `infisical secrets|export|run`, so I could not supply one. `vendo sync` said the same at boot: *"no Vendo Cloud key in this environment"*.

Screenshots (on the VM, `/tmp/s4-live/`): `A-01-login-yousef.png`, `A-02-signedin-yousef.png`, `A-03-agent-open.png`, `A-04-prompt-typed.png`, `A-05-turn-t*.png`, `B-01-login-mia.png`, `B-02-signedin-mia.png`, `B-03-turn-fails.png`.

### Second blocker, independent of the key

**Maple declares no storage collection anywhere.** No `storage` key in `.vendo/`, `.vendo/remixable/*.json`, or `src/`; the shipped scripted-turn engine (`examples/demo-bank/src/demo-script/engine.ts`) never touches `vendo_apps_data_*`. The only way to get a declared collection in this demo is a live LLM generation via `vendo_make` — so even with a key, the proof is a multi-turn generation + a share/grant (data tools call `requireOwned(appId, ctx)` at `editor`, `packages/apps/src/server/doors/data-tools.ts:27,55,73`), not a two-minute check.

### Secondary (read-only): a host whose subjects contain `/`

Asked: on current `main`, what happens when the agent saves through the data tools if `ctx.principal.subject` is e.g. `org/user`?

**A clean `validation` refusal, on every appData verb, before any DB work.** `appDataOwner` (`packages/store/src/app-data-rows.ts:52-57`) is composed by `appDataRows` at construction (`:91`) and by `appDataFiles` (`:182`), so records and files both refuse; `invalid()` throws `VendoError("validation", …)` (`packages/store/src/validate.ts:35-37`). Verified by running the built store package against `owner: "org/user"`:

```
put:     code=validation  app data owner "org/user" must be non-empty and free of "/"
list:    code=validation  …
get:     code=validation  …
putFile: code=validation  …
```

That reaches the model as a tool error with code `validation` and that exact message (`packages/apps/src/server/doors/agent-tools.ts:247-256`, `:316-318`) — nothing crashes and nothing half-writes. Two caveats for a real host: the wall shows up only at the **first data call** (no boot-time or `vendo sync` warning), and the message says "app data owner", which the host has to map back to its own JWT `sub`.

**One more thing found while reading** (not a fix request, just so it's known): `backingFor` (`packages/apps/src/server/persistence/app-data.ts:62-72`) falls back to the un-owned `resolveAppStorage` façade when `ops === undefined`. `selectStoreOps` (`packages/vendo/src/compose-store.ts:111-114`) returns `undefined` for a store carrying neither its own `ops` nor a SQL handle — i.e. a host that passes a hand-written `VendoStore`/`StoreAdapter`. `createStore()` (incl. BYO Postgres) and the hosted store both have ops, so the shipped paths are stamped; a fully custom adapter silently keeps the pre-S4 shared drawer — no stamp, no scoping, and no `/` check. The comment at `:55-61` says this is deliberate; flagging that it is also silent.

### To actually run this proof

Needs `ANTHROPIC_API_KEY` or `VENDO_API_KEY` materialized into `examples/demo-bank/.env.local` on the runner, plus budget for a real generation turn.

PR #1201, opened 2026-08-11
fix(ui): the citation card is reachable, edge-aware, and its byline stops breaking mid-item

yousefh409:
## What

Three interaction defects in the knowledge citation popover (`TurnCitations` / `CitationChip`), one commit, `packages/ui` only:

1. **Hover gap-death.** The popover opens 8px below its chip and reveal was a pure CSS `:hover` — the pointer dies crossing the gap, so the popover could never actually be reached by hover. Reveal moves to JS hover intent on the `.fl-cite` wrapper (the card is its child, so any approach path re-enters the same element): open on `pointerenter`, close on a 260ms grace timer, re-entry cancels. Click now pins (survives pointer leave); Escape and outside `pointerdown` close; opening one chip closes the incumbent. Touch pointers are excluded from hover handling — tap-to-toggle unchanged. The `@media (hover: hover)` reveal rule is deleted: with a JS-opened card it made an open card undismissable under a resting pointer.
2. **Edge clipping.** The card was hard-anchored `left:0` at fixed width with no collision handling, so a chip at the end of a wrapped row pushed it past a narrow surface's edge (+141px overflow at a 410px surface). A `useLayoutEffect` on open measures against the nearest `.fl-msglist` (the actual `overflow:auto` clipper; falls back to `documentElement`) and clamps inline `left` to an 8px gutter. Bails when the bound has no width, so jsdom tests are untouched.
3. **Byline mid-item wrapping.** `.fl-cite-porigin` ("source · kind · visibility") broke inside items on long source URLs — "Product docs" split across two lines. `flex-wrap: wrap` + inherited `white-space: nowrap` makes each anonymous flex item unbreakable and moves wrapping between items. Zero markup change.

New: `packages/ui/test/chrome/knowledge-citation-hover.test.tsx` (6 tests — grace-timer reachability, pin, Escape, outside-click, single-open, clamp bail). Verified red (5/6) against the unfixed component. No existing test modified; the existing click-driven suites (`knowledge-citations.test.tsx`, both Playwright specs) pass unmodified — `aria-expanded` / `.fl-cite--open` semantics are preserved.

## Why

Found while recreating the shipped component pixel-for-pixel for a knowledge-UI design exploration: the recreation reproduced these against the real CSS. The hover defect means the snippet-on-hover affordance has effectively never worked with a mouse; the clipping hides citations entirely on narrow embedded surfaces (the overlay runs ~360-400px).

## Verification

- [x] `pnpm build && pnpm test:affected && pnpm typecheck && pnpm lint` green (`@vendoai/ui`: 972 tests / 114 files; browser smoke 12/12). Local Playwright `accessibility` + `screenshots` specs: 18 passed, 2 pre-existing fixme skips — includes the axe audit on an open popover.
- [x] Screenshots attached — real Chromium, real driven pointer. Resting chip row is **byte-identical** before/after (`00` pair).

| Defect | Before | After |
|---|---|---|
| Hover travel (straight) |  |  |
| Right-edge chip, 410px surface |  |  |
| Byline, long source |  |  |

<details><summary>All 10 before/after pairs</summary>

| | Before | After |
|---|---|---|
| Resting row (unchanged) |  |  |
| Hover diagonal |  |  |
| Pin survives leave |  |  |
| Escape closes pin |  |  |
| Outside click closes |  |  |
| Byline shipped source |  |  |
| Touch tap opens |  |  |

</details>

Screenshot commit: `e7e70cc` on `assets/fl-cite-fixes` (delete after merge).

---
## Summary by cubic
Fixes the knowledge citation card: it’s now reachable on hover, respects container edges, and its byline keeps items intact while long sources truncate inside the card. Improves mouse/keyboard behavior while keeping touch tap-to-toggle unchanged.

- **Bug Fixes**
  - Hover intent with a 260ms grace: open on pointer enter, close after grace; click pins; Escape or outside click closes; only one card open; removed CSS `:hover` reveal; `aria-expanded` semantics unchanged.
  - Edge-aware positioning: on open, measure against the nearest `.fl-msglist` (fallback `documentElement`) and clamp `left` with an 8px gutter to prevent overflow on narrow surfaces.
  - Byline containment: `.fl-cite-porigin` uses `flex-wrap: wrap` + `white-space: nowrap` to wrap between segments; the source is now `.fl-cite-psource` with `min-width: 0` + ellipsis so long URLs don’t overflow while keeping full text accessible.
  - Tests: added `packages/ui/test/chrome/knowledge-citation-hover.test.tsx` (6 tests) for hover/pin/close behavior and `packages/ui/test/chrome/knowledge-citation-byline.test.tsx` for source truncation; existing tests unchanged.

<sup>Written for commit 4be861dae8ddbb0c82143668d586492c90d01fcb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1201?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/chrome/chrome-css.ts
+++ b/packages/ui/src/chrome/chrome-css.ts
@@ -1614,21 +1614,29 @@ export const CHROME_CSS = ONEST_FONT_CSS + `/* @vendoai/ui chrome — the S1 des
   transition: border-color .12s ease, background .12s ease; }
 .fl-cite-btn svg { flex: none; color: var(--vendo-fg-muted); }
 .fl-cite-btn:hover, .fl-cite--open .fl-cite-btn { border-color: var(--vendo-border-strong); }
-/* The snippet popover: click-toggled everywhere, hover-revealed on pointer
-   devices (the mockup's chip:hover behavior). */
+/* The snippet popover. Open state is the ONE source of truth (turn-citations.tsx
+   drives it from hover intent, click-to-pin, Escape and outside-click): the card
+   sits 8px below its chip, and a :hover rule cannot survive that gap — nor could
+   an open card be dismissed under a pointer still resting on the chip. The left
+   below is only the resting anchor: a chip near a narrow surface's right edge
+   gets an inline left from the same component, so nothing here may shorthand it. */
 .fl-cite-pop { position: absolute; left: 0; top: calc(100% + 8px); width: 292px; z-index: 5;
   background: var(--vendo-surface); border: 1px solid var(--vendo-border); border-radius: 10px;
   box-shadow: var(--vendo-shadow-float); padding: 12px 14px; text-align: left; cursor: default;
   display: none; }
 .fl-cite--open .fl-cite-pop { display: block; }
-@media (hover: hover) { .fl-cite:hover .fl-cite-pop { display: block; } }
 .fl-cite-ptitle { font-size: 12.5px; font-weight: 600; color: var(--vendo-fg);
   display: flex; align-items: center; gap: 6px; }
 .fl-cite-ptitle svg { color: var(--vendo-fg-muted); }
 .fl-cite-psnippet { display: block; font-size: 12px; color: var(--vendo-fg-muted); line-height: 1.6;
   margin: 7px 0 9px; border-left: 2px solid var(--vendo-border); padding-left: 9px; }
+/* nowrap + wrap, together: a nowrap container makes each segment (source, kind,
+   visibility — anonymous flex items, so unstylable on their own) unbreakable,
+   and wrapping then happens only BETWEEN them. Without the pair the row is one
+   nowrap line whose items shrink to min-content, which is what split the kind
+   label "Product docs" down the middle next to a long help.maple.bank source. */
 .fl-cite-porigin { font-size: 11px; color: var(--vendo-fg-muted);
-  display: flex; align-items: center; gap: 5px; }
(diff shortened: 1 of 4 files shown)


yousefh409:
Follow-up `4be861da` from the review pass: the byline fix's own `white-space: nowrap` left a genuinely long single source with no break at all — a 64-char `help.maple.bank` path painted 77px past the card border and handed `.fl-msglist` horizontal scroll. The source is now a real element (`.fl-cite-psource`, anonymous flex items can't be constrained) with `min-width:0` + ellipsis — the cut lands on the tail slug, which restates the card's title; the origin host stays readable. Paint-only: the full source remains in the DOM for screen readers and copy.

| | Before | After |
|---|---|---|
| 64-char source |  |  |
| same at 410px surface |  |  |
| shipped-length source (unchanged) |  |  |

Shipped-length case verified pixel-stable (3 antialiasing px of 53k, no layout shift). New seam test `knowledge-citation-byline.test.tsx` (3 tests, verified red without the fix). Local gate green including both fixture suites force-re-executed serially (their one red in the batched run was the documented full-suite scheduling flake — same task hash passes from CI's cache and on forced local re-run).

PR #1244, opened 2026-08-13
feat(guard)!: an explicit host rule survives the unattended downgrade

yousefh409:
## ⛔ STILL BLOCKED — do not merge, do not approve

Round 2 applied the coordinator's ruling. It **closed the second blocker**
(`mcp-door-parity.e2e.test.ts:249` is green, untouched). Two remain red:

1. `fixtures/integration/tests/away-park-revoke.e2e.test.ts:112` (J5) — the
   mechanism the ruling names does not reach it. Source proof in "Why the
   ruling does not reach J5" below.
2. `packages/vendo/tests/harness-system-prompt.test.ts` — "promises only the
   discovery rail this harness and this deployment actually have" now exceeds
   its 30s budget. NOT load flake: run alone on an idle box it is 15.9s green
   with `packages/guard/src` at `main` and 30.6s timed-out with this branch.
   The likely cause is the feature working as designed — away rule-decided
   calls now EXECUTE instead of parking instantly, so the unattended turn in
   that test does real tool work it never used to do. Whether that is
   acceptable (and the budget moves) or a signal the exemption is too broad is
   a call for the security round, not mine.

## The change

The away/unattended downgrade parked every would-be `run` that was not
grant-authorized, which made `{ presence: "away", action: "run" }` a policy
rule nobody could write. A `run` decided by an explicit host policy **rule**
now survives the downgrade — under two fail-closed limits.

**The exemption's one condition, set in exactly one place** — the host-rules
loop in `#pipeline`, `packages/guard/src/guard.ts:1382-1387`:

```ts
return withInvalidated({
  decision: { action: rule.action, decidedBy: "rule" },
  // Only a rule the host WROTE. A preset is a shorthand they selected, not
  // a rule they authored, so it does not carry the away exemption.
  ...(this.#presetPolicy ? {} : { hostRule: true }),
});
```

Declared at `guard.ts:182` (`hostRule?: true`), read at `guard.ts:1083-1087`.

**Not gated on `decidedBy === "rule"`.** `normalizeCodeDecision`
(`guard.ts:455-467`) deliberately labels `policy.code` decisions `"rule"` too,
so the deploy-time hatch stays away-downgraded. Gating on the string would have
exempted it. The flag is what tells the two apart.

## The resolution, in the ruling's terms

**An explicit user denial always outranks a host rule's run verdict —
fail-closed.** A host rule is standing authority written ahead of time; a
standing denial is this user answering *this* question, and the later, more
specific answer wins. Implemented on the existing `#standingDenial` seam at
`guard.ts:1083-1087`: losing the exemption drops the call back into the away
park, where the same denial turns it into the "you denied this" block, so a
rule cannot reopen a question the user already closed.

**Preset-expanded rules are excluded from the exemption, for the same
reason.** `resolvePolicyConfig` (`policy.ts:46-56`) is the only place a preset
becomes rules, and a string policy is always a preset, so `#presetPolicy`
(`guard.ts:495-499`, set at `guard.ts:540`) reads the provenance with no new
type, no per-rule tagging and no spoofable field. Preset rules are never mixed
with authored ones — `PolicyResolver.rules()` (`policy.ts:137-144`) returns
either the inline/preset set or the file's, never both — so one boolean is
exact rather than approximate.

**Both are deliberately conservative.** They remove capability, never add it,
and can be loosened later if a real use case appears.

## Deliberate contract change

`packages/guard/test/pipeline-conformance.test.ts:372` — "parks an away call
even when a rule says run" — asserted the exact behavior the spec removes. It
is **deliberately updated**, not repaired: it now reads "runs an away call when
an explicit host rule says run" and asserts
`{ action: "run", decidedBy: "rule" }`.

## Untouched and green

`packages/guard/test/security/away-code-hatch.test.ts` (P1-A) — not one
character changed. It is the positive proof the exemption did not leak to the
code hatch:

```
 ✓ test/security/away-code-hatch.test.ts (2 tests) 27ms
```

`packages/vendo/tests/mcp-door-parity.e2e.test.ts:249` — untouched, and fixed
by the preset exclusion:

```
 ✓ §12 — an UNATTENDED turn's door call carries `away`/`automation` into the AUDIT, identically to in process  8671ms
```

## The six proofs (`packages/guard/test/pipeline-conformance.test.ts`)

| # | Test | Proves |
|---|---|---|
| a | `executes rule-decided away reads and writes end to end` | rule-decided read/write tools EXECUTE unattended |
| b | `still refuses a rule-decided away destructive or ungraded call` | destructive and ungraded are STILL blocked unattended — THE LAW is downstream |
| c | `still parks an away run decided by the judge or by the default posture` | judge and default-posture runs STILL downgrade to ask |
| d | `still parks an away run decided by policy.code, which is never an explicit rule` | the code hatch STILL parks |
| e | `lets an explicit user denial outrank a host rule that says run` | a user's no beats a rule added afterwards |
| f | `still parks a preset-decided away call, with its approval row, since a preset is not authored` | a preset does not take the exemption |

## Why the ruling does not reach J5

`fixtures/integration/tests/away-park-revoke.e2e.test.ts:112` still fails:

```
- "code": "needs-permission",   - "tool": "host_invoices_update",
+ "code": "validation",
```

The ruling routes through "an explicit user denial". In J5 there is no denial
row that the guard can see for that call. Proof from source and from base
behavior, not inference:

- `#standingDenial` (`guard.ts:1779-1797`) matches only rows where
  `deniedBy === "human"` **and** `sameParkedCall(...)`, which requires
  `request.call.id === call.id` (`guard.ts:446`).
- The park path at `guard.ts:1158` runs that same check first, and turns a
  match into a **block**. On `main`, J5's sweep **parks** (`pending-approval`,
  `needs-permission`) rather than blocking. Therefore no matching denial row
  exists for that call.

J5's user denies the sweep at automation-**enable capture** time, which is a
different call id. Their "no" is expressed as the **absence of a grant**, not
as a denial row — and parking-on-absence-of-grant is exactly the behavior this
change removes for host-authored rules. The fixture's `.vendo/policy.json`
says `write → run` and is hand-authored, so the preset exclusion does not
reach it either.

The open question is therefore narrower than before, and unresolved: **does a
host-authored `run` rule authorize an away step whose capture-time consent the
user declined?** Today it does, and the step then dies downstream with no
actAs authority — a clean resumable `needs-permission` becomes an opaque
`validation` error. Options are to exclude away/automation venues from the
exemption, or to require capture consent regardless of rule; both are further
narrowings, and neither is mine to choose.

## Gate

```
build      exit=0   Tasks: 21 successful, 21 total
typecheck  exit=0   Tasks: 42 successful, 42 total
lint       exit=0   Tasks:  4 successful,  4 total
guard      exit=0   Test Files 27 passed | 1 skipped (28) · Tests 287 passed | 3 skipped (290)
```

`pnpm test:affected`: 31 of 34 tasks successful. Red: `@vendoai-fixtures/integration`
(J5, above), `@vendoai/vendo` (the discovery-rail budget, above), and
`@vendoai/genbench` (pre-existing on this box — missing `chrome-headless-shell`).

### DIFF
--- a/packages/guard/src/guard.ts
+++ b/packages/guard/src/guard.ts
@@ -492,6 +492,11 @@ class GuardImplementation implements VendoGuard {
   readonly #config: CreateGuardConfig;
   readonly #policyConfig: PolicyConfigObject | undefined;
   readonly #policy: PolicyResolver;
+  /** A string policy IS a preset (`resolvePolicyConfig`), and a preset's rules
+   *  are never mixed with authored ones — `PolicyResolver.rules()` returns
+   *  either the inline/preset set or the file's, never both. So one boolean
+   *  says whether every rule this guard evaluates was authored by the host. */
+  readonly #presetPolicy: boolean;
   readonly #maxCallsPerMinute: number;
   readonly #maxWritesPerRun: number;
   readonly #callWindows = new Map<string, number[]>();
@@ -532,6 +537,7 @@ class GuardImplementation implements VendoGuard {
     // policy misconfiguration `resolvePolicyConfig` catches) must fail loud
     // from `createGuard` itself.
     this.#policyConfig = resolvePolicyConfig(config.policy);
+    this.#presetPolicy = typeof config.policy === "string";
     this.#policy = new PolicyResolver(this.#policyConfig, config.policyCloudFallback);
     this.#maxCallsPerMinute = resolveBreakerLimit(
       config.breakers?.maxCallsPerMinute, "maxCallsPerMinute", DEFAULT_MAX_CALLS_PER_MINUTE,
@@ -1067,9 +1073,17 @@ class GuardImplementation implements VendoGuard {
     // refusal in `bind().execute()` and the `projectableForRun` projection are
     // downstream of every decision here, so a rule can free a read or a write
     // and nothing more.
+    //
+    // And a person's NO outranks the rule. A host rule is standing authority
+    // written ahead of time; a standing denial is this user answering THIS
+    // question, and the later, more specific answer wins. Losing the exemption
+    // drops the call back into the park below, where the same denial turns it
+    // into the "you denied this" block — the rule cannot re-open a question the
+    // user already closed.
     if (
       ctx.presence === "away" && draft.action === "run" && draft.decidedBy !== "grant"
-      && metadata.hostRule !== true
+      && (metadata.hostRule !== true
+        || await this.#standingDenial(call, effectiveDescriptor, ctx))
     ) {
       draft = { action: "ask", decidedBy: "default" };
     }
@@ -1365,7 +1379,12 @@ class GuardImplementation implements VendoGuard {
           decision: { action: "block", reason: rule.note ?? "blocked by policy rule", decidedBy: "rule" },
         });
       }
-      return withInvalidated({ decision: { action: rule.action, decidedBy: "rule" }, hostRule: true });
+      return withInvalidated({
+        decision: { action: rule.action, decidedBy: "rule" },
+        // Only a rule the host WROTE. A preset is a shorthand they selected, not
+        // a rule they authored, so it does not carry the away exemption.
+        ...(this.#presetPolicy ? {} : { hostRule: true }),
+      });
     }
 
     const code = this.#policyConfig?.code;
--- a/packages/guard/src/guard.ts
+++ b/packages/guard/src/guard.ts
@@ -174,6 +174,12 @@ interface DecisionMetadata {
   decision: DraftDecision;
   rationale?: string;
   invalidatedGrants?: PermissionGrant[];
+  /** Set ONLY by the host-rules loop in `#pipeline`, and read only by the away
+   *  downgrade (05 §6), which exempts an explicit host rule. It cannot be
+   *  `decidedBy === "rule"`: `normalizeCodeDecision` deliberately labels
+   *  `policy.code` decisions "rule" too, and that deploy-time hatch must STAY
+   *  away-downgraded (P1-A). The flag is what tells the two apart. */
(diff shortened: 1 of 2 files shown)


yousefh409:
Closing. This branch holds alternative semantics nobody approved, so it can never merge as-is, and a parked draft is clutter. The work is not lost — it is recorded here and in the follow-up issue.

**Why Option A cannot land, in short:** it exempted a would-be `run` verdict decided by an explicit host rule from the unattended downgrade. Three things defeated that, in increasing order of seriousness:

1. `decidedBy === "rule"` is also set by `normalizeCodeDecision` for `policy.code` (`guard.ts:449-461`) — deliberately, so the deploy-time code hatch stays downgraded. Gating on the string would have exempted that hatch. Solved with a flag set only in the rules loop.
2. Presets expand into rules (`policy.ts:7-56`), so a `cautious`-posture read executed ungranted with no approval row. Solved by excluding preset-expanded rules.
3. **Not solvable.** In `fixtures/integration/tests/away-park-revoke.e2e.test.ts:112` the user declines at automation-enable capture time, and that "no" is stored as the *absence of a grant*, not as a denial row. The away downgrade is precisely what enforces it. Ordering "an explicit user denial beats a rule" above the exemption cannot help, because there is no denial row to order. Applied honestly, fail-closed means the exemption can never fire for a call that lacks a grant — which is every case Option A existed to enable. Fail-closed plus Option A is a no-op.

Both guard-level orderings were implemented and green (`guard.ts:1083-1087`, `guard.ts:1386`; guard suite 287 passing) and `test/security/away-code-hatch.test.ts` stayed untouched and green throughout. The blocker was never the implementation.

Follow-up with the real design question: see the linked issue.

Issue #1253, opened 2026-08-13
Next hosts are never told about serverExternalPackages - the local-store default crashes under webpack


amrib247:
## Summary
The local-store default (PGlite) crashes on a real Next.js host unless the host's config carries `serverExternalPackages: ["esbuild", "@electric-sql/pglite"]` — webpack bundles PGlite, and its fs calls then fail with `TypeError: The "path" argument must be of type string … Received an instance of URL` (cross-realm URL). Every Vendo example ships that line, but nothing tells a real host: `vendo init` doesn't mention it, doctor doesn't check it, and no docs page states it. Field case (2026-08-13): switching a Cloud-keyed Linkwarden deployment to an explicit local store took the whole wire down with unhandled rejections until the line was found by diffing against the examples.

## Affected
Every Next.js host using the local store (the zero-config dev default for keyless installs — i.e. the documented BYO path). Docs-site + `vendo init`'s Next scaffold guidance + doctor, at main.

## Suggested fix
Three layers, cheapest first: (1) docs — persistence page + quickstart state the required line; (2) `vendo init` on a Next host prints it as a manual step when no `serverExternalPackages` covers PGlite (init never edits host files, but its tail can name the paste); (3) doctor warns when the composition resolves the local store on a Next host whose config lacks the entry — the failure is deterministic, so the check is too.

amrib247:
Docs half landed on feat/eng379-baseline-fix-packs, commit 33892157a: the persistence page now carries the serverExternalPackages warning with the exact line every example ships. The init-tail manual step and the doctor check remain open here.

YSharara1:
Hit this independently on a Next 16 host and landed on the same missing line, but through a different failure — worth adding because the symptom shares nothing with the one above and the search terms don't overlap.

**Turbopack, production only.** `next dev` is completely clean. After `next build && next start`, every wire request 501s with:

```
TypeError: f.instantiateWasm is not a function
```

Turbopack's production chunking mangles PGlite's Emscripten module where dev's does not, so the break only appears after a build — long past the point where you'd suspect the store. There's no cross-realm `URL`, no `path` argument error, nothing pointing at PGlite or persistence at all. `f.instantiateWasm` is minified and unsearchable.

The fix is the same line, plus a third entry:

```ts
// next.config.ts
serverExternalPackages: ["esbuild", "@electric-sql/pglite", "@vendoai/store"],
```

Two data points I'd add to the case for fix (3), the doctor check:

[...]

Issue #1251, opened 2026-08-13
A client older than the console fails as a silent 501 - 'Unknown store operation' deserves a loud upgrade instruction


amrib247:
## Summary
When the console outgrows an installed client, the failure mode is a **silent 501 on every store-backed route** — no log line, no upgrade hint, doctor's static checks unaffected. Field case (2026-08-13): a 0.8.0 deployment against the current console (after the one-store op refactor, #1089, 31→27 ops): `GET/POST /threads` answered 501 in ~350ms with nothing in the server log, because the console rejects the old client's store ops with `not-implemented: "Unknown store operation: …"` and the wire maps that straight through. The only place the truth appeared was the background sweep's retry line. The operator experience is "chat is broken, no reason given" — we spent an hour ruling out rate limits and spend caps first.

## Affected
`packages/vendo/src/hosted-store.ts` error mapping + the wire's not-implemented handling, at any version older than the live console. This will recur on every breaking console deploy for every not-yet-upgraded install.

## Suggested fix
The hosted store client should recognize the console's `Unknown store operation` rejection as a **version-skew signal** and fail loudly and actionably: one log line and an error message naming the real cause — "this deployment's @vendoai/vendo is older than the Vendo Cloud console — upgrade the package" — instead of a bare 501. Doctor's live checks should surface the same diagnosis (it already makes store round trips). A wire-level version handshake would be the complete fix, but the message upgrade alone removes the silent-mystery failure.

amrib247:
Message-level fix on feat/eng379-baseline-fix-packs, commit 33892157a: an Unknown-store-operation rejection now appends the version-skew diagnosis (this @vendoai/vendo is older than the console - update the package) to the thrown not-implemented error, and logs one loud line per process so a server operator sees the cause without a debugger. A proper wire version handshake remains the complete fix and stays open here.

yousefh409:
Fixed by #1254 — the hosted store now throws a clear 'package is older than the console' error and logs once. The deeper version handshake isn't planned; doctor is static-only now.

Issue #1250, opened 2026-08-13
Dev servers accumulate per-composition background intervals - long-running next dev grinds the hosted store into rate limits


amrib247:
## Summary
A long-running `next dev` host accumulates Vendo compositions — one per route recompile — and every composition keeps its background intervals alive after it is orphaned: the session sweep (60s) and, since dev-local schedule firing, the dev automations ticker (60s). After hours of recompiles the process carries dozens of live interval sets, all hitting the hosted store. Field measurement (2026-08-13, Cloud-keyed Linkwarden host): the sweep was firing **every few seconds** (it should be once a minute), the hosted-store boot notice printed on nearly every request, and the standing Cloud traffic kept the org's rate limit tripped **with no browser open at all** — every chat turn then died mid-stream on "Too many requests". A fresh restart returned the cadence to exactly one sweep per minute and zero 429s.

## Affected
`packages/vendo` composition lifecycle (`compose-sweep` / `startBackgroundSweep`, `startDevAutomationsTicker`) under dev-server module churn; any Cloud-keyed dev host that stays up for hours. Observed at v0.8.0; the mechanism is version-independent.

## Suggested fix
Orphaned compositions must not keep ticking. Options, not exclusive: key the sweep/ticker arming to the store identity (or a process-level registry) so a re-created composition adopts rather than duplicates; or give compositions a disposal path the Next dev module lifecycle can reach; or make the intervals self-check that their composition is still the live one. Worth a regression test that creates N compositions against one store and asserts a single active sweeper. (The dev automations ticker half is being addressed on the ENG-379 baseline branch; the sweep is upstream's.)

amrib247:
The dev-ticker half is fixed on the combined baseline branch feat/eng379-baseline-fix-packs, commit 33892157a: armDevTickerOnce arms the development automations ticker once per PROCESS (Symbol.for flag on globalThis, surviving module re-evaluation), so churned compositions no longer stack tickers. The session-sweep half needs the same treatment but is a composition-lifecycle design call left upstream - the sweep predates the baseline work and a process-level guard would need to consider multi-composition processes.

amrib247:
[...]

Issue #1094, opened 2026-08-09
Local dev deployments have no automation firing authority: Cloud can't reach localhost and nothing drives the local tick


amrib247:
**Symptom (field, Cloud-keyed localhost host, 2026-08-09 — the fifth and final layer of the automations story, after #1076/#1086/#1090/#1093):** an every-5-minutes automation is authored, armed, granted — the full consent flow green — and thirty minutes pass with zero runs.

**Root cause — a local deployment has NO firing authority at all, twice over:**

1. Under the hosted store, `composeAutomations` empties `localTriggerKinds` unconditionally: "Cloud is the firing authority" for schedule/external. That is true for **hosted deploys** — but Cloud's scheduler cannot reach a dev server (a localhost wire has no public URL and is in no deployment inventory), so on local dev the authority never exercises, and the local engine politely refuses to fire "to avoid double-running". Armed schedules go nowhere, forever, and the only mention is a boot-log paragraph.
2. Even self-hosted (keyless), nothing drives the local tick: the engine ships `start()` (a self-ticking interval) but **nothing in the umbrella ever calls it** — the production design is an external cron POSTing `/tick` with `VENDO_TICK_SECRET` (doctor's E-SCHED-001 checks it for machine schedules), which no laptop has.

Net: every local development deployment — the environment every new user starts in — arms schedules that can never fire. This is the tail end of the baseline's F15.

**Fix implemented** on the combined branch `feat/eng379-baseline-fix-packs` (commit 585129333):

- `localFiringKinds` (exported, unit-tested truth table): a **development** composition keeps schedule firing local even under the hosted store. Double-running is structurally impossible — the schedule-cursor claims are atomic in the shared store (`insertIfAbsent`/`compareAndSwap`), so two firers can never win the same tick. Deployed processes are unchanged.
- A development composition arms the engine's own minute ticker from the `ready()` latch — the same Workers-safe arming the background sweep uses (never at construction), once per composition. The engine's `start()` interval is now unref'd (the same rule its own run-delay follows), so it never keeps a dev server from exiting.
- The hosted-store boot notice tells the development story honestly: "This is a DEVELOPMENT process, so schedule automations fire locally…".

Verified live: the dev server's boot log carries the new notice and the ticker arms on first request. Gates: automations 107/107, the new truth-table tests, dependency-guard, and the portability gate (workerd leg) all green.

yousefh409:
[...]

PR #1254, opened 2026-08-13
Baseline fix packs: auth fails loud, Windows installs, honest doctor, automations e2e, docs batch, overlay UX (ENG-379, F1-F17)

amrib247:
## What

The combined baseline-findings branch: every fix from the day-1 new-customer baseline (ENG-378 → ENG-379, findings F1–F17) plus the automations field-testing that followed — six packs and the field fixes they surfaced, developed and field-verified on a real host (Linkwarden: yarn-4 monorepo, Next 15 Pages Router, next-auth v4). 26 commits, rebased onto current main.

**Pack 1 — auth fails loud (ENG-383, #870–#874).** `NEXTAUTH_SECRET` fallback in both authJs halves; lazy-until-cookie session claims with a next-auth v4 hint; init's v4 advisory; the wire names a failed principal resolver (#872); doctor tells a declined actAs mint apart from an unconfigured seam (E-AUTH-008) and its pass message stops claiming host verification the probe never performs.

**Pack 2 — installs work on Windows (ENG-384, #875–#876).** The provider install runner goes through the platform shell with quoted args, captures a bounded stderr tail into the warning, and a resolvable pre-v6 `ai` (a workspace-hoisted copy) is treated as a floor violation by init and doctor both.

**Pack 3 — an honest doctor (ENG-385, #877–#880).** Console-managed deployments stop failing E-CFG-001 (a missing local surface with a Cloud key is a resolution mode, warned not failed); an answering-but-broken `/status` is no longer "unreachable" — the wire's own error body rides the E-LIVE-001 message; non-JSON bodies get their own diagnosis.

**Pack 4 — creates surface their server work (ENG-386, #881–#882).** A create that authors an automation hands the envelope to its caller (`CreateServerWork` through `onServerWork`), the `vendo_make` receipt goes `"partial"` with the reason when required server work could not be built, and the automation card publishes on create exactly as on edit. Unified during the rebase with main's independently-built failure-only signal: one channel per outcome, merged in the bridge.

**Field fixes from automations e2e (#1076, #1078, #1086, #1090, #1093, #1094).** The Cloud gateway model family gets the explicit 64K output cap instead of the stock provider's silent 4096 clamp (the F15 root cause); a BOM'd `package.json` never crashes the CLI; the schedule verb stops teaching the two-step trap and the escalate door names away work; the in-thread automation card is the arming consent surface (decidable pending grants over the durable approvals feed); arming asks carry the approvals listing refs on every StoreAdapter; a development process fires its own scheduled automations (Cloud cannot reach a localhost wire — deployed processes unchanged, pinned by a truth-table test).

**Pack 5 — docs batch (ENG-387, #883–#886).** agents.md/quickstart/CLI/vendo-init gain the monorepo rule and the shared "where commands run" callout (login writes `.env.local` to the CWD); a new `connect/uninstall` page with the full verified init-footprint reversal and the eject naming collision defused; the local-files-shadow-console-publishes warning on dot-vendo and a new Console-managed config section on the Vendo Cloud page. Every claim verified against source before it went into a page.

**Pack 6 — overlay UX (ENG-388).** The overlay remembers conversations: reload resumes the last thread (origin-scoped id + the existing validate/load/mid-stream-resume machinery) and a previous-conversations header button opens an in-place picker. The launcher anchors to any viewport corner and takes host pixel offsets — the whisper and completion toast move with it. Both field-verified.

**Field-outage guards from PR-prep testing (#1250, #1251, #1253).** The dev automations ticker arms once per process (Next dev module churn stacked tickers that ground the hosted store into rate limits); the hosted store names client/console version skew loudly instead of silent 501s; the persistence page documents `serverExternalPackages` for PGlite on Next hosts.

## Test infrastructure hardened along the way

- `wire-server.ts` teardown destroys lingering sockets (a long-lived overlay test parked the next test's close until the 30s hook timeout, silently skipping cleanup).
- Shared jsdom setup clears localStorage per test — chrome persists real state now.
- Fixture thread titles derive from the first message (the real store's rule).
- Two upstream lessons honored in the rebase: main's report-outside-the-try double-callback guard and its `"partial"` receipt status were kept and unified with the envelope.

## Tests & gates

Changesets for every user-facing change ride the branch. Local (Windows): targeted suites for every touched seam green — ui 61/61 (overlay history/placement/consent/steering/export-surface), apps 13/13 (create-server-work seam incl. the ported pack-4 tests), vendo 221/222 + hosted-store 59/59 (the one red is the pre-existing POSIX-path assertion that only fails on Windows); build + eject assembly + typecheck green across ui/apps/vendo; MCP shim regenerated. The full suite is CI's job per the repo rule — this PR's `ci` check is the gate of record.

## Pending

- UI evidence screenshots (consent card, history picker, resumed conversation, launcher offset) — being attached to `docs/verification/` follow-up by @amrib247 from the field-test host.
- The F13 console half (publish-screen copy + ownership indicator) lives in the console repo, tracked on Linear.

Linear: ENG-379 (umbrella), ENG-383…388. Field log: the per-issue fix comments on each GitHub issue above.


amrib247:
Greptile P1 (stale-composition ticker) is real and fixed in the last commit: armDevTicker now ADOPTS - arming stops the predecessor interval via the stop handle the engine start() returns and starts the newcomer, so exactly one dev ticker runs and it always belongs to the composition currently serving requests. Covered by an adopt-contract unit test; #1250 comment updated separately.

Issue #1176, opened 2026-08-10
sync: the impact probe ignores VENDO_BASE_URL and the project's dotenv — ◇ Impact is dead by default on demo-bank


yousefh409:
Found in coverage walk 2 (`vendo sync` on `examples/demo-bank`, `main` @ `f792ce4`). Recordings on [`assets/init-redesign-walk2`](https://github.com/runvendo/vendo/tree/assets/init-redesign-walk2) (`sync-real-impact.txt` is the run with the URL passed by hand; `sync-real-judgment.txt` is the same walk without it).

## The bug

The blast-radius probe hardcodes its own idea of where the host is:

```ts
// packages/vendo/src/cli/sync-flow.ts:536
const wireUrl = (options.url ?? process.env.VENDO_URL ?? "http://localhost:3000/api/vendo").replace(/\/+$/, "");
```

`VENDO_BASE_URL` — the variable that *is* the host's public URL, path prefix included, and the one every other URL Vendo builds hangs off — is not consulted, and neither is the project's `.env` / `.env.local`, even though `runSyncFlow` has already parsed both into `env` two dozen lines earlier (`sync-flow.ts:663`) for the credential ladder.

## Why it matters

Vendo's own flagship demo is the counter-example. `examples/demo-bank` is served under `/maple`, so its wire base is `http://localhost:3000/maple/api/vendo`. A developer running `vendo sync` in demo-bank — with the dev server up and 24 automations in the store — gets:

```
│  impact unknown — dev server not reachable at http://localhost:3000/api/vendo
```

Pass the real URL and the same catalog change answers immediately:

```
◇  Impact
│  host_listAccounts breaks 4 automations
│  host_transferMoney no saved references
```

So the ◇ Impact block — the one part of the redesign that tells a developer whether a tool edit is about to break saved user work — is **dead by default** on any host that is not at exactly `localhost:3000` with no base path. The failure is silent and reads as "server not running", so nobody investigates.

## Suggested fix

Fall back through the env the flow already read, before the hardcoded default:

```
options.url ?? env.VENDO_URL ?? joinUrl(env.VENDO_BASE_URL, "/api/vendo") ?? "http://localhost:3000/api/vendo"
```

`joinUrl` already exists in the wire package and already handles the mount-path case. That single change makes the block work in demo-bank, on a non-3000 port, and behind a tunnel, with no new flags and nothing for the user to configure — the value is already in the file init wrote.

[...]

PR #1258, opened 2026-08-13
fix(sync): impact probe falls back through VENDO_URL / VENDO_BASE_URL…

mud-mos23:
… before the hardcoded localhost default (#1176)

## Verification

- [ ] `pnpm build && pnpm test && pnpm typecheck && pnpm lint` green

---
## Summary by cubic
Resolves the Impact probe URL in `vendo sync` from the merged env before falling back to localhost. Previously it used only `VENDO_URL` or `http://localhost:3000/api/vendo`; now it prefers `VENDO_URL`, else `VENDO_BASE_URL` + `/api/vendo`, fixing base‑path hosts (e.g., `/maple`), bounds the probe to 10s, and only suggests a variable when none was used.

• Resolution order: `--url` > `VENDO_URL` > `VENDO_BASE_URL + /api/vendo` > `http://localhost:3000/api/vendo`.
• Uses the same merged env (dotenv + shell); invalid or blank values fail‑soft. URL assembly via `joinUrl` from `@vendoai/core`.
• Failure note suggests setting `VENDO_URL` or `VENDO_BASE_URL` (or passing `--url`) only when the default was used.
• Adds tests for base‑path resolution, env precedence, `.env.local` seam, and timeout; patch release to `@vendoai/vendo`.

<sup>Written for commit 50aa9d30c76b3efadbe1fc34327c629a31e0aa65. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1258?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Confirmed the bug and the fix is right — it even closes a latent empty-`VENDO_URL` case as a bonus. Two asks before merge:

1. **Add a dotenv seam test.** The `.env` / `.env.local` half of #1176 has no coverage right now — all four new tests set env via `vi.stubEnv`, which is the path main already honored. There's a ready pattern in the same file (~line 753) that writes a real `.env.local` to a tmpdir.
2. **Add `AbortSignal.timeout(...)` to the probe fetch.** This change points that request at arbitrary remote hosts during production `prebuild` (VENDO_BASE_URL is set in deploy environments), and it's currently untimed — a hung host would stall the build. `cli/cloud/client.ts` already threads a signal as precedent.

Then a quick rebase (one 3-line import conflict) and I'll approve the CI run. Thanks — clean, focused change.

PR #1274, opened 2026-08-14
feat(agents): arm() raises the away asks, and run({ identity }) runs on what they minted

yousefh409:
Stacked on #1273.

`agent.arm(subject, { appId, triggerId?, tools? })` raises one approval per wanted tool. On approve, guard mints a standing `source:"automation"` grant bound to (subject, appId, triggerId). `run({ identity: { appId, triggerId? } })` carries that identity into the existing `RunContext` fields, so an armed run proceeds unattended instead of parking.

**Identity is a lookup key, not a credential.** It authenticates nothing. Subject binding is enforced by the guard, not assumed: `#matchingGrant` scopes the query by `ctx.principal.subject` and re-checks `grant.subject`, so a run for B cannot use A's grant (verified adversarially).

**Acceptance:** an agents-minted grant is returned unmodified by automations' `liveAutomationGrants(subject, appId, triggerId)` — one scheme. `packages/vendo/tests/arming-one-scheme.test.ts` proves it as a real seam: real write path, real read path, no stub on either side.

## Stated limit

The raised asks live in this process's memory until answered. A restart between raising an ask and deciding it loses the mint, and the person re-arms. Persisting it would mean a new reserved collection — out of scope.

## ⚠️ An adversarial check round found a critical defect. It is fixed. Read this part.

`createAutomations()` and `arm()` both subscribe `onApprovalDecision`, and both claimed any approved `venue:"automation"` approval carrying an `appId`. Whichever subscriber **registered first** won the approval's one-time `consumed` transition. In the ordinary boot order (engine before agent) automations won — and it derives the trigger from the run row named by `ctx.trigger.runId`, a row `arm()` never wrote. So `triggerId` came out `undefined` and the grant was minted **app-wide**:

- the armed trigger held **nothing** — every armed run parked forever while `arm()` reported success;
- unarmed trigger `main` held **standing unattended authority nobody consented to**.

**Fix:** `arm()` writes a capture row keyed by the approval id, which the existing automations subscriber already reads as its first branch, so the correct `triggerId` is used. Entirely inside `packages/agents/src/` — `consent.ts` and `engine.ts` are untouched.

**Ordering-independence is tested, not claimed.** `packages/vendo/tests/arming-collision.test.ts` runs an automations capture ask and an arming ask raised and decided in the same process, each asserted to be claimed only by its own listener, plus the reverse — under **both** subscriber registration orders. 4/4 green, identical in both.

Revert-to-red on that test reddens the engine-first ordering only. That is correct rather than weak: pre-fix, the agent-first ordering already produced the right grant, so there is no observable difference to assert there. That asymmetry *is* the ordering dependence the fix removes.

Two further findings from the same round, also fixed: `held` omitted the presence-only tools (`vendo_apps_pin`/`unpin`, which the guard refuses away by name), and `held` read the tool's *authored* risk rather than the grade this deployment's `resolveRisk` actually applies. Both meant carding a person for a tool that could never run and minting a grant dead on arrival.

Our own acceptance test was also rearranged: it had been constructing the automations engine *after* the decision — the single ordering that hides the collision. It now boots the engine first and still passes.

## Hard law

`presenceMatches` and the away downgrade in `packages/guard/src/guard.ts` are **byte-identical to main** (SHA-256 compared). This commit touches zero guard source and zero automations source — only `packages/agents/src/{agent,arming,away,index}.ts` and tests.

## Two things for the owner to decide separately (not addressed here)

1. **An armed grant also authorizes attended calls.** `presenceMatches`' present branch is `grant.appId === undefined || grant.appId === ctx.appId` — it ignores `source` and `triggerId`. Pre-existing and byte-identical to main, but arming is the first feature that lets a plain `agent()` mint such grants, so it widens the blast radius. Any fix is inside the frozen `presenceMatches`.
2. **A `no` on an arm() ask now also reaches automations' disarm-on-bare-no branch** for that (app, trigger) — a side effect of the capture row. Scoped to that trigger and guarded by "nothing else pending and no live grant", so it reads as correct (a no disarms what it said no to), but it is new behavior the arming door did not have before.

## Gates

`pnpm build` 0 · `pnpm --filter @vendoai/agents --filter @vendoai/vendo typecheck` 0 · `pnpm lint` 0 · agents 66/66 · vendo `arming-binding` + `arming-one-scheme` + `arming-collision` 8/8.

---
## Summary by cubic
Adds `agent.arm()` to raise away approvals and mint standing automation grants bound to (subject, appId, triggerId), and makes `run({ identity })` execute against those grants. Previously, unattended runs parked and a boot-order race could mint app‑wide grants; now armed runs proceed and minting is trigger‑correct and registration‑order independent.

- API:
  - `agent.arm(subject, { appId, triggerId?, tools? })` raises one approval per reachable tool; tools withheld from unattended or presence‑only are returned in `held` and never asked. On approve, the guard spends the approval and mints a standing `source:"automation"` grant scoped to the trigger. Asks are tracked in‑process memory; a restart loses them.
  - `run({ identity: { appId, triggerId? } })` carries identity into `RunContext` (or uses `agent({ identity })` as default). Identity is a lookup key only; it authenticates nothing.
  - A capture row (`automations:captures`) keyed by approval id ensures the automations subscriber uses the correct trigger; this removes the engine/agent ordering race.
- Safety and rollout:
  - Guard enforces subject binding; a run for B cannot use A’s grant. A “no” on an arming ask also reaches automations’ disarm‑on‑bare‑no for that trigger. Existing `presenceMatches` means armed grants also match attended calls.
  - Required action: callers that want unattended runs must pass `identity` to `run()` or configure `agent({ identity })`.
  - No schema migration. Depends on guard `spendApproval`/`mintGrant`; code feature‑detects and fails closed if absent. Only `@vendoai/agents` is changed.

<sup>Written for commit 85fb1b53f41ef6d9aeff1bf7e0a9bc60ec6747d7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1274?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/agents/src/agent.ts
+++ b/packages/agents/src/agent.ts
@@ -28,6 +28,7 @@ import { resolveDevCredential } from "@vendoai/harnesses/inference/credential";
 import { createStore, storeFiles, type VendoStore } from "@vendoai/store";
 import type { LanguageModel, UIMessage } from "ai";
 import { randomUUID } from "node:crypto";
+import { createArm, type ArmRequest, type ArmResult } from "./arming.js";
 import { startRun, type AgentRun, type RunOptions } from "./away.js";
 import { resolveDoor, type DoorConfig } from "./door.js";
 import { withEgress, type EgressConfig } from "./egress.js";
@@ -87,6 +88,9 @@ export interface AgentConfig {
    * forgery-safe `[User]`/`[Situation]` blocks to the host, to keep or to drop.
    */
   system?: SystemPromptHook;
+  /** Which automation this agent's runs ARE, when it only ever is one — the
+   *  default behind `run({ identity })`, which still wins per run. */
+  identity?: RunOptions["identity"];
 }
 
 export interface VendoAgent {
@@ -106,6 +110,15 @@ export interface VendoAgent {
    *  happened. Venue "automation", presence "away", non-interactive — so a tool
    *  the guard wants a person for parks, and `refs.approvals` is who to ask. */
   run<T = never>(task: string, options?: RunOptions<T>): AgentRun<T>;
+  /**
+   * Ask this person to let an unattended run act as them: one approval per tool
+   * the run could reach, raised on the permission wire for them to answer. A yes
+   * mints a STANDING grant bound to (subject, appId, triggerId) — which is
+   * exactly what `run({ identity })` runs on, so an armed run proceeds instead
+   * of parking. Tools no unattended run may reach at all come back as `held`,
+   * unasked.
+   */
+  arm(subject: string, request: ArmRequest): Promise<ArmResult>;
   session(subject: string, options?: SessionOptions): Promise<AgentSession>;
   /**
    * This agent's MCP door, present exactly when its harness thinks outside this
@@ -367,6 +380,7 @@ export function agent(config: AgentConfig): VendoAgent {
     assertModel: requireModel,
     ...(config.instructions === undefined ? {} : { instructions: config.instructions }),
     ...(config.system === undefined ? {} : { system: config.system }),
+    ...(config.identity === undefined ? {} : { identity: config.identity }),
     // The other half of the door: a credential the harness minted resolves to
     // NOTHING until the turn it points at is published, so without this line a
     // mounted door 401s every tool call the box makes.
@@ -384,6 +398,7 @@ export function agent(config: AgentConfig): VendoAgent {
       return session.stream(message, options.signal === undefined ? {} : { signal: options.signal });
     },
     run: (task, options) => startRun(deps, task, options),
+    arm: createArm({ guard, store, tools: bound }),
     async session(subject, options) {
       await requireModel();
       return createSession(deps, subject, options);
--- a/packages/agents/src/arming.ts
+++ b/packages/agents/src/arming.ts
@@ -1,0 +1,155 @@
+/**
+ * ARMING — the present-time ceremony that turns "let this run while I am away"
+ * into standing authority. The asks are raised through the guard's own away
+ * check, so the question the person answers is exactly the question that firing
+ * would have raised — and the grant a yes mints is the one that firing looks up.
+ *
+ * KNOWN LIMIT — the raised asks live in THIS process's memory until they are
+ * answered. A restart between raising an ask and deciding it loses the mint, and
+ * the person arms again. Persisting them would mean a reserved collection of
+ * this package's own, which arming does not have.
+ */
+import {
+  DEFAULT_TRIGGER_ID,
+  descriptorHash,
+  presenceOnlyCall,
+  withheldFromUnattended,
+  type AppId,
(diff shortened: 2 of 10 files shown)


yousefh409:
Closing this without merging. The branch stays (`yousefh409/p4-arming`, `85fb1b53f`) — everything below should feed the re-spec.

**Why.** `arm()` works, and the critical cross-trigger defect found in review is genuinely fixed. But the fix buys the mint by writing a row into `automations:captures`, a table arming does not own. Three other readers consume that table and none can tell arming's row apart from a real automations consent request, so:

- an arming card left **untapped** keeps a fully denied automation armed;
- a **no** on an arming card switches an unrelated automation off (engine-first subscriber order only — so the ordering-independence the collision test proves for the approve path does not hold for the deny path);
- the guard's one-hour TTL sweep counts as a no, so **nobody has to do anything at all** for that to fire;
- the arming card is counted as an automations permission the person still owes.

Each of the three review rounds on this mechanism uncovered another reader. That pattern is the signal: the real gap is that arming has **no consent record of its own** and borrows automations', which is a design question, not a bug list. Patching the symptoms inside `arming.ts` again would be a fourth round with no reason to think it is the last.

**What is worth keeping for the re-spec**

- The seam contract held: `arm()` was built in parallel against a frozen contract and met `@vendoai/guard`'s real mint with zero conflicts.
- `presenceMatches` and the away downgrade were never touched — verified byte-identical to main at every round.
- The acceptance property is right and was proven for real: a grant minted through the agents path is read back unmodified by automations' `liveAutomationGrants`, real write path and real read path, no stub on either side (`packages/vendo/tests/arming-one-scheme.test.ts`).
- The collision test (`packages/vendo/tests/arming-collision.test.ts`) is a good harness for whatever replaces this — it exercises both subscriber registration orders. It needs extending to the **deny** path, which is where the remaining defects live.
- Two findings that are about arming's honesty rather than its plumbing, and will recur in any design: `held` must account for the presence-only tools the guard refuses by name, and it must read the grade `resolveRisk` actually applies rather than the authored one. Both are fixed on this branch.
- Known limit that shaped the design: raised asks live in process memory, so a restart between raising and deciding loses the ask.

**Separately, and still open** — an armed grant also authorizes *attended* calls, because `presenceMatches`' present branch ignores `source` and `triggerId`. Pre-existing and byte-identical to main, so untouched here, but arming is the first feature that would let a plain `agent()` mint such grants. It wants a deliberate decision before any arming design ships.

PR #1275, opened 2026-08-14
chore(vendo): VENDO_STORE_TRACE — one stderr line per store wire call

yousefh409:
This is the measurement instrument of record for the datastore latency work. With `VENDO_STORE_TRACE` set, every call through the store wire client in `packages/vendo/src/hosted-store.ts` — the 35 named ops and the StoreAdapter façade alike, since both ride that one client — emits a single greppable stderr line naming the op, its wire path, the round trip in milliseconds (the body read included, because that is what the caller actually waits for) and the response size in bytes, e.g. `vendo-store-trace op=transcripts.getThread path=/transcripts/getThread ms=87 bytes=41233`. The change is purely additive and a no-op when the env var is unset: the switch is read once at import, so `call` is then the untouched `wire` with no wrapper to pay for, and the adapter rule still holds — `hostedStore` reads no environment at construction or on any call, and its behavior still comes only from its constructor arguments. No new dependencies, no config surface, no public API change.

---
## Summary by cubic
Adds an opt-in stderr trace for every store wire call to measure datastore latency and failures. Previously no trace; with `VENDO_STORE_TRACE` set, each call logs one line with op, path, elapsed ms including body read, response bytes, and an `outcome`. Unset, behavior is unchanged and overhead is zero.

- Reads `VENDO_STORE_TRACE` once at import; when unset, `call` is the unwrapped client. No new deps or API/config changes, and `hostedStore` still reads no environment.
- Measures size by buffering a clone and returns the original `Response` unchanged (avoids 204 behavior changes). If a 2xx body stream breaks, logs `outcome=unreadable-body` with `bytes=0` and the call still resolves.
- On failures, emits a line with `outcome` set to the server error code or the error name (e.g., `TimeoutError`); `bytes=0`, then rethrows.

<sup>Written for commit 5b5b41b412e04874ddf6e70bb8f0e459de4a0b1d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1275?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/vendo/src/hosted-store.ts
+++ b/packages/vendo/src/hosted-store.ts
@@ -284,11 +284,24 @@ type StoreWireOp = keyof typeof STORE_WIRE_PATHS;
  * call through the store client below — the 35 ops AND the StoreAdapter façade,
  * which rides the same client — emits one greppable stderr line naming the op,
  * its path, the round trip in milliseconds (the body read included, since that
- * is what the caller waits for) and the response size in bytes. Read ONCE at
- * import, like the skew latch below: a debug switch belongs to the process, so
- * the adapter itself still reads nothing at construction or on any call, and
- * unset there is no wrapper to pay for. */
+ * is what the caller waits for), the response size in bytes and the outcome.
+ * Read ONCE at import, like the skew latch below: a debug switch belongs to the
+ * process, so the adapter itself still reads nothing at construction or on any
+ * call, and unset there is no wrapper to pay for. */
 const TRACE = environment("VENDO_STORE_TRACE") !== undefined;
+
+/** The `outcome=` field of a FAILED call's trace line. Failures are the tail of
+ * the latency distribution, not an afterthought: an exhausted 30s budget is the
+ * slowest call this client ever makes, so tracing successes alone would drop
+ * exactly the calls the measurement exists to find and make a turn read as fast
+ * as its lucky requests. The server's code where there is one (VendoError and
+ * the console's mapped errors both carry it), else the error's name —
+ * `TimeoutError` for a spent budget. */
+const traceOutcome = (error: unknown): string => {
+  const code = (error as { code?: unknown } | null | undefined)?.code;
+  if (typeof code === "string") return code;
+  return error instanceof Error ? error.name : "unknown";
+};
 
 /** ONE key per LOGICAL mutation. The retry below replays it verbatim — that
  * replay is the only reason the server can tell "do it again" from "you
@@ -400,12 +413,25 @@ function storeWireClient(
   /** Untraced, `call` IS `wire` — see {@link TRACE}. */
   const call: typeof wire = !TRACE ? wire : async (op, path, init) => {
     const started = performance.now();
-    const response = await wire(op, path, init);
-    const body = await response.arrayBuffer();
-    console.error(
-      `vendo-store-trace op=${op} path=${path} ms=${Math.round(performance.now() - started)} bytes=${body.byteLength}`,
-    );
-    return new Response(body, { status: response.status, headers: response.headers });
+    const line = (bytes: number, outcome: string): void =>
+      console.error(
+        `vendo-store-trace op=${op} path=${path} ms=${Math.round(performance.now() - started)} bytes=${bytes} outcome=${outcome}`,
+      );
+    let response: Response;
+    try {
+      response = await wire(op, path, init);
+    } catch (error) {
+      // A refusal has a body, but `wire` already consumed it to build this
+      // error — the bytes are gone and the failure is what mattered anyway.
+      line(0, traceOutcome(error));
+      throw error;
+    }
+    // Measure a CLONE and hand back the response `wire` returned, untouched.
+    // Rebuilding it instead would be a real behavior change on a null-body
+    // status — `new Response(body, { status: 204 })` throws — and would put the
+    // instrument in the path of every hosted call it is supposed to observe.
+    line((await response.clone().arrayBuffer()).byteLength, "ok");
+    return response;
   };
 
   const json = async (response: Response): Promise<unknown> => response.json().catch(() => ({}));
--- a/packages/vendo/src/hosted-store.ts
+++ b/packages/vendo/src/hosted-store.ts
@@ -20,6 +20,7 @@ import {
 } from "@vendoai/store";
 import { consoleSender, raiseCloudError } from "./cloud-console.js";
 import { keepAliveFetch } from "./keep-alive-fetch.js";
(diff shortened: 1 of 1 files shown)


yousefh409:
Triaged both review findings against a run, not an argument. Pushed `231010b`.

**Finding 1 — no trace line on a failed call: REAL, fixed.** For a general logger this would be a nit; for this instrument it was a defect, because the slowest calls are exactly the ones that fail — an exhausted 30s budget is the slowest call this client ever makes, and it was vanishing from the log, making a turn read as fast as its lucky requests. A failed call now emits its own line with a new `outcome=` field carrying the server's code where there is one (`VendoError` and the console's mapped errors both carry `code`), else the error's name, and `bytes=0` because `wire` has already consumed the refusal body to build the error. Same one-line-per-call greppable shape:

```
vendo-store-trace op=engine.put   path=/engine/put   ms=56 bytes=0   outcome=conflict
vendo-store-trace op=blobs.get    path=/blobs/get    ms=1  bytes=0   outcome=not-found
vendo-store-trace op=engine.get   path=/engine/get   ms=1  bytes=0   outcome=TimeoutError
vendo-store-trace op=engine.get   path=/engine/get   ms=1  bytes=109 outcome=ok
```

Across `hosted-store.test.ts` + `hosted-store-notice.test.ts` that is 238 lines, **33 of them failures that were previously invisible** (`conflict`, `blocked`, `validation`, `not-found`, `not-implemented`, `cloud-required`, `unavailable`).

**Finding 2 — body consumption vs the JSON fallback: the stated concern is a FALSE POSITIVE, but probing it found a different real bug in the same lines, now fixed.**

On the stated concern: the byte measurement lived entirely inside the traced branch, and the untraced path was not merely equivalent but *the same function object* — `const call: typeof wire = !TRACE ? wire : ...`, so with `VENDO_STORE_TRACE` unset there is no wrapper, no `arrayBuffer()` call, and nothing to consume a body twice. A body is never consumed more than once in either mode: a non-2xx throws inside `consoleSender` (`if (!response.ok) await options.raise(response)`) *before* the wrapper ever receives a response, so the traced branch only ever sees a 2xx, and `raiseWireError`'s `JSON.parse(await response.text())` reads a stream the trace has not touched.

I proved it by driving the real code path with the failure and fallback shapes the suite cares about — enveloped 4xx, non-JSON 502, bare 404 HTML, enveloped 404, a **2xx whose body is not JSON** (the `json().catch(() => ({}))` fallback itself), malformed 2xx, an empty 204, a good 200, and a timeout — under both env settings, and diffing what each surfaced. Nine for nine identical.

That probe did catch something Greptile could not, having been blocked before collection: `new Response(body, { status: response.status, ... })` **throws on a null-body status**, so a 204 surfaced `TypeError: Response constructor: Invalid response status code 204` when traced, against the correct service-misbehavior error when untraced. That was a genuine behaviour change introduced by switching the instrument on — unacceptable on a file in the critical path of every hosted call. Fixed by measuring `response.clone()` and returning the response `wire` returned, untouched, so the reconstruction is gone entirely and traced/untraced now return the identical object. After the fix the nine-case diff is clean.

Gate: `pnpm build` 21/21 · `pnpm typecheck` 42/42 · `pnpm lint` clean (4 pre-existing demo-bank warnings, untouched) · `hosted-store.test.ts` + `hosted-store-notice.test.ts` 60/60 with the var unset **and** 60/60 with it set.

yousefh409:
**Correction to my previous comment — Finding 2 was REAL, and I was wrong to call it a false positive.**

I answered a claim Greptile did not make. Its actual words were "when a 2xx response's **body stream fails during reading**". I probed bodies that *parse* badly (non-JSON text, malformed JSON, an empty 204) and concluded the paths matched. A body that parses badly and a body whose stream *errors mid-read* are different things, and my probe never produced the second one. That is on me.

Probed properly — a 200 whose `ReadableStream` enqueues some bytes and then calls `controller.error(...)` — the traced and untraced paths diverged on every case:

| case | untraced | traced (before fix) |
|---|---|---|
| stream breaks after partial JSON | `Error: …invalid record response` | `Error: stream broke mid-read` |
| stream breaks immediately | `Error: …invalid record response` | `Error: stream broke mid-read` |
| stream breaks on a list | `Error: …invalid list response` | `Error: stream broke mid-read` |
| **stream breaks on a void mutation** | **RESOLVED** | **THREW** |

The last row is the one that matters: `engine.delete` never inspects its payload, so untraced a broken body stream is *a successful call*. With tracing on it became a failed one. Turning on an instrument was making calls fail — exactly the wire-path regression class this file cannot afford, and precisely Greptile's point.

Root cause: `await response.clone().arrayBuffer()` rejects when the underlying stream errors, and I let that rejection escape `call`. Untraced, that same stream error is absorbed by the existing `json()` helper's `.catch(() => ({}))`.

Fixed in `5b5b41b`: the measurement can no longer fail the call it measures. A broken read is recorded as a fact *about* the call — `outcome=unreadable-body`, `bytes=0` — and the original response is handed on untouched, so the existing JSON fallback decides the outcome exactly as it always did.

```
vendo-store-trace op=engine.delete path=/engine/delete ms=1 bytes=0 outcome=unreadable-body
```

Re-probed across **11 cases** covering both claims — the four stream-failure shapes above, an intact 200, an enveloped 409, a non-JSON 502, a non-JSON 200, a malformed 200, an empty 204, and a timeout — under both env settings, diffing what each surfaced. Identical on all 11.

What survives from my earlier comment, and I still stand behind because it is mechanical rather than argued: with `VENDO_STORE_TRACE` unset there is genuinely no wrapper at all (`const call: typeof wire = !TRACE ? wire : …` — the same function object), so the untraced production path is unchanged. The defect was real but scoped to running *with* the instrument on, which is the mode the latency work will actually use, so it would have bitten us.

Thanks for the catch — and noting for the record that Greptile's own validation was blocked before collection, so this one was found by review reasoning, not by its run.

Gate on `5b5b41b`: `hosted-store.test.ts` + `hosted-store-notice.test.ts` 60/60 with the var unset **and** 60/60 with it set (238 trace lines, 33 of them failures), plus the 11-case probe above. `build`/`typecheck`/`lint` were green on the previous commit and are re-running now behind a shared machine lock; CI on this PR is the gate of record and I will not consider this landed until it is green.

PR #1278, opened 2026-08-14
perf(store,core,vendo): transcripts.appendMessages — append without downloading the thread

yousefh409:
## What

Appending one message to a hosted thread downloaded the **whole conversation
first**. The wire had no verb carrying an owner, so `threadMessageStore`'s hosted
half had to `getThread` just to read `data.subject` before it could `putMessage`
(`helpers/thread-messages.ts:76-78` names exactly this gap). One agent turn paid
it several times, and the payload grew with the conversation forever. The SQL
half has always done it in one statement.

This adds the missing verb.

- **`transcripts.appendMessages`** — the additive 36th wire op. Body
  `{threadId, subject, messages, title?}`; answer `{revision, count}`, and
  deliberately **not the thread** — echoing the transcript back would reintroduce
  the exact payload the op exists to delete. (Rejected: an optional `subject` on
  `putMessage`; the wire schemas are `.passthrough()`, so an older console would
  silently ignore it.)
- **Engine** — `appendThreadMessages` (`helpers/rows.ts`): a subject-guarded
  create-or-touch of the thread row plus one multi-row insert, in one
  transaction, and it never moves an existing message's `seq`. Both halves of the
  new `ThreadMessageStore.upsertMany` run those same semantics, so they cannot
  disagree.
- **One persistence path per turn** — `Thread.revision` is carried from the read
  that already happened, so `persist`'s first attempt compare-and-swaps on it
  instead of re-reading the row (and its transcript) for a token it was handed.
  `persist` now runs only when the row must be created; every later turn appends,
  title and all. `MAX_PERSIST_ATTEMPTS`, the CAS-miss path, new-thread
  `insertIfAbsent` and the per-row CAS door are untouched.
- **`runtime.ts`** sends a turn's changed messages as ONE `upsertMany`.

## Wire compatibility, both directions

**Old client → new console:** untouched. The op is purely additive; nothing an
0.16.x client sends changes shape.

**New client → old console:** an **explicit capability feature-detect**, not a
catch-and-degrade. `/status` is the wire's own discovery handshake and its op
count is the signal (`STORE_WIRE_APPEND_MESSAGES_OPS = 36`, frozen at the count
this op shipped with). `servesAppendMessages` asks once per `StoreOps` handle,
caches the answer in a `WeakMap`, and routes to the existing
`getThread` + `putMessage` loop when the mount is behind. The shape is the
in-repo precedent at `tool-bridge.ts:380` (`guard.previewCheck`).

This does **not** reintroduce the failure mode behind field incident #1251. That
incident was about reading a *failed mutation* as a signal; `hosted-store.ts`
still logs once and rethrows on an unsupported op, and nothing here catches it.
The fallback is a supported route chosen **before** anything is written, so there
is no half-applied write to retry down a second path. A failed handshake is not
an answer: it is not cached, it does not degrade, and the write fails.

`StoreOps.transcripts.appendMessages` is **optional**, the way `RecordStore.claim`
and `atomic` already are — an implementation that cannot serve it says so by
omitting it, and the conformance suite now asserts that a mount's `/status` count
tracks the capability it advertises.

## Proof

`packages/store/tests/thread-messages.batch.test.ts` — no stub on either side.
Both halves sit on ONE real database, so each writes and the other reads back
through its own real path:
1. the real ops path and the real SQL path land **byte-identical transcripts**,
   read back through both read paths (4 combinations, all equal);
2. a **foreign subject is refused by both** halves, and nothing is written;
3. the **feature-detect fallback** (a stand-in console that reports the 35-op
   count and 501s the route, exactly as a real old console does) lands the **same
   transcript** as the native op;
4. the answer is `{revision, count}` and does not grow with the conversation.

Prove-can-fail: reverting the statement to `SET seq = EXCLUDED.seq` on conflict
turns tests 1 and 3 red.

## Gate

`pnpm build` ✅ · `pnpm typecheck` ✅ (42/42) · `pnpm lint` ✅ (0 errors) ·
store suites 152 passed / 10 files ✅ (batch, message-loss, ask-user-answer,
ops.conformance, conformance, thread-messages, thread-messages.ops,
thread-message-cas, ops, thread-scoping) · vendo 89 passed ✅ (hosted-store,
hosted-store-notice, threads, harness-threads, harness-wire,
harness-refusal-atomicity) · harnesses 78 passed ✅. No UI surface touched.

## Two things the reviewer should know

1. **Op count 35 → 36 lands in more places than the plan named.** Besides
   `packages/core/tests/store-wire.test.ts` (3 coupled literals), the count is
   pinned in `packages/store/tests/ops.test.ts` (2 literals, updated) and
   asserted as `=== 35` in `packages/core/src/conformance/store-ops.ts`, which is
   shared by `createStoreOps` (36) and `memoryStoreOps` (35). That assertion now
   derives the expected count from whether the mount serves the op — which makes
   the handshake a checked capability contract rather than a frozen number.
2. **The conformance case is a gate on the console half.** The new
   `transcripts.appendMessages` case skips only when the op is absent. The hosted
   client always defines it, so vendo-web's `store-ops-conformance` will fail
   until the console's dispatch arm ships — deliberately, and the stacked console
   PR carries both.

---
## Summary by cubic
Append thread messages in one call and enforce turn order under concurrency. Old: clients fetched the whole thread and looped `putMessage`; writers computed `seq` before taking the thread row, causing large payloads and occasional duplicate `seq`. New: `transcripts.appendMessages` accepts `{threadId, subject, messages, title?}`, assigns `seq` in-transaction, returns `{revision, count}`, and all writers (`appendMessages`, `putMessage`, `recordAnswer`) take the thread row before allocating `seq`.

- Wire/API: adds `transcripts.appendMessages`; `/status.ops` is now 36 and clients feature-detect (`>= 36`), falling back to `getThread` + `putMessage`. No thread echo; no schema/UI changes.
- Engine/local backend: subject-guarded upsert + one multi-row insert; unique message-id validation; `title` via COALESCE; `seq` from `max(seq)+1` under the thread row lock; existing messages keep position. `putMessage` and `recordAnswer` now lock the thread row before `seq` allocation.
- Client/runtime: `ThreadMessageStore.upsertMany` batches a turn and uses `appendMessages` when served; otherwise falls back. Threads carry `revision`; `persist` CASes on the carried token; non-creation turns append with a derived title.
- Tests: conformance derives expected `/status.ops` from op presence; seam tests ensure identical transcripts across SQL/wire paths, reject foreign subjects, verify feature-detected fallback; PostgreSQL race tests prove no shared `seq` across `appendMessages` vs `putMessage`/`recordAnswer`. Answers are `{revision, count}`.

**Rollout**
- Servers: implement `transcripts.appendMessages`, report `/status.ops >= 36`, or omit the op and stay `< 36`.
- Clients: no migration required; they auto-fallback until servers update.
- Run the updated conformance suite.

<sup>Written for commit 472312e39e40042e9e22484924dfc0fea229bc64. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1278?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/core/src/conformance/store-ops.ts
+++ b/packages/core/src/conformance/store-ops.ts
@@ -1,7 +1,7 @@
 import { ENGINE_ALLOWLIST_VERSION, engineAppHistory } from "../engine-collections.js";
 import type { VendoErrorCode } from "../errors.js";
 import { isoDateTimeSchema } from "../ids.js";
-import { VENDO_STORE_WIRE_FORMAT } from "../store-wire.js";
+import { STORE_WIRE_APPEND_MESSAGES_OPS, VENDO_STORE_WIRE_FORMAT } from "../store-wire.js";
 import type { StoreOps } from "../store.js";
 import { assert, assertBytesEqual, assertDeepEqual } from "./assertions.js";
 import type { ConformanceCase, ConformanceSuite } from "./index.js";
@@ -1103,6 +1103,37 @@ export function storeOpsConformance(opts: StoreOpsConformanceOptions): Conforman
         await assertThrowsCode(() => ops.lifecycle.promote("app_absent", "org_1"), "not-found", "promoting an unknown app");
       }),
 
+      /** The batch append: ownership is the caller's `subject`, so the mount
+          checks it in its own statement and the client never downloads the
+          thread to check it first. Optional — a mount that omits it is served
+          by putMessage, and this case says so instead of failing it. */
+      opsCase(opts, "transcripts.appendMessages lands a batch under the named subject", async (ops) => {
+        const append = ops.transcripts.appendMessages;
+        if (append === undefined) return;
+        await ops.transcripts.putThread({ id: "thr_am1", subject: "u", messages: [] });
+        const landed = await append("thr_am1", "u", [
+          { id: "msg_a", role: "user", text: "one" },
+          { id: "msg_b", role: "assistant", text: "two" },
+        ], { title: "one" });
+        assert(landed.count === 2, `appendMessages should report 2 rows, got ${landed.count}`);
+        assert(typeof landed.revision === "string", "appendMessages should report the thread's new revision");
+        // The answer is the revision and the count — NOT the thread. Echoing the
+        // transcript back is the payload this op exists to stop paying.
+        assertDeepEqual(Object.keys(landed).sort(), ["count", "revision"], "appendMessages answered with more than {revision, count}");
+
+        const got = await ops.transcripts.getThread("thr_am1");
+        const messages = (got!.data as Record<string, unknown>)["messages"] as unknown[];
+        assert(messages.length === 2, `appendMessages did not land both messages: got ${messages.length}`);
+
+        // A foreign subject is refused by the statement, not by a pre-check.
+        await ops.transcripts.putThread({ id: "thr_am2", subject: "owner", messages: [] });
+        await assertThrowsCode(
+          () => append("thr_am2", "someone_else", [{ id: "msg_x", role: "user", text: "not mine" }]),
+          "conflict",
+          "appending to another subject's thread",
+        );
+      }),
+
       // =====================================================================
       // status
       // =====================================================================
@@ -1111,7 +1142,14 @@ export function storeOpsConformance(opts: StoreOpsConformanceOptions): Conforman
         const status = await ops.status();
         assert(status.format === VENDO_STORE_WIRE_FORMAT, `status.format should be ${VENDO_STORE_WIRE_FORMAT}`);
         assert(typeof status.ops === "number", "status.ops should be a number");
-        assert(status.ops === 35, `status.ops should be 35, got ${status.ops}`);
+        // The count is the handshake's ONLY capability signal, so it must track
+        // what the mount actually serves — a client feature-detects the batch
+        // append on exactly this number (STORE_WIRE_APPEND_MESSAGES_OPS). A
+        // mount that omits the optional op is the 35-op vintage and says so.
+        const expected = ops.transcripts.appendMessages === undefined
+          ? STORE_WIRE_APPEND_MESSAGES_OPS - 1
+          : STORE_WIRE_APPEND_MESSAGES_OPS;
+        assert(status.ops === expected, `status.ops should be ${expected}, got ${status.ops}`);
       }),
     ],
   };
--- a/packages/core/src/store-wire.ts
+++ b/packages/core/src/store-wire.ts
@@ -39,12 +39,13 @@ export const STORE_WIRE_PATHS = {
   "appData.getFile": "/app-data/getFile",
   "appData.listFiles": "/app-data/listFiles",
   "appData.deleteFile": "/app-data/deleteFile",
-  // transcripts (6)
+  // transcripts (7)
   "transcripts.putThread": "/transcripts/putThread",
   "transcripts.getThread": "/transcripts/getThread",
   "transcripts.listThreads": "/transcripts/listThreads",
   "transcripts.deleteThread": "/transcripts/deleteThread",
   "transcripts.putMessage": "/transcripts/putMessage",
+  "transcripts.appendMessages": "/transcripts/appendMessages",
   "transcripts.recordAnswer": "/transcripts/recordAnswer",
   // harness (3)
   "harness.get": "/harness/get",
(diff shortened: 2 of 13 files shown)


yousefh409:
**Confirmed, and fixed.** You were right, and it reproduces on real PostgreSQL 17.11 (throwaway docker, genuine concurrent connections — PGlite is single-connection and can never catch this).

**Before the fix**, 20 rounds of two concurrent `appendMessages` on one thread:

```
duplicate seqs: [0,1,2,2,3,3,4,4,5,5,6,6,7,7,8,8,9,9,10,10,
                 11,11,12,12,13,13,14,14,15,15,16,16,17,17,18,18,19,19,20,20]
expected 21 to be 40
```

40 messages, **21 distinct seqs** — 19 of 20 rounds collided. Your diagnosis of the mechanism was exact: the `max(seq) + 1` read happened *before* `appendThreadMessages` took the thread row, so under READ COMMITTED both racers read the same tail and the row lock serialised only the writes, not the numbering. And the consequence was as you said — `THREAD_MESSAGES_AGGREGATE` orders by `(seq, id)`, so equal seqs fall back to message-id order, which is not turn order.

The SQL half had the same hazard by a different route: two overlapping turns each passed their own array index as `seq`.

**The fix** moves position assignment into the statement that already holds the row:

```sql
INSERT INTO vendo_thread_messages (thread_id, id, seq, message, created_at, updated_at)
SELECT $1, m.id, tail.next + m.n - 1, a.elem, $4, $4
FROM unnest($2::text[]) WITH ORDINALITY AS m(id, n)
JOIN jsonb_array_elements($3::jsonb) WITH ORDINALITY AS a(elem, ordinality)
  ON a.ordinality = m.n
CROSS JOIN (SELECT COALESCE(max(seq) + 1, 0) AS next
            FROM vendo_thread_messages WHERE thread_id = $1) tail
ON CONFLICT (thread_id, id) DO UPDATE ...
```

What serialises it is statement 1, the subject-guarded thread upsert that runs first in the same transaction: the losing racer blocks there — on the row lock, or on the primary key when both are creating the thread — until the winner COMMITs. Statement 2 is then a fresh READ COMMITTED snapshot that already sees the winner's rows, so `tail.next` is correct.

Consequently `upsertMany` and `appendThreadMessages` now take **no caller `seq`**. A position computed outside the transaction cannot be made safe, so the parameter is deleted rather than documented — which also removes the last difference between the two halves' semantics.

**After the fix:** the race test passes, and passes on three consecutive runs. Prove-can-fail is the transcript above.

The blueprint's claim that "overlapping turns writing disjoint message ids never collide" was true of the message *rows* and false of their *order*. It is true of both now.

**Test:** `packages/store/tests/thread-messages.batch.test.ts` — "concurrent appends to one thread never share a seq". Postgres leg only (`it.runIf(backend.name === "postgres")`), asserting every seq distinct **and** that round *k* precedes round *k+1* in the read-back transcript. Round 0 deliberately races on a thread that does not exist yet, so the create path is covered alongside the update path.

**Re-gate:** store 258 passed / 9 files (both legs, pglite + real Postgres) · vendo 89 passed · harnesses 78 passed · `build` + `typecheck` (42/42) + `lint` (0 errors) green.

yousefh409:
**Confirmed again — both pairings collided, and both are fixed.** Real PostgreSQL 17.11, throwaway docker, 20 rounds each:

```
appendMessages × putMessage    duplicate seqs: [0,0,1,1,2,2,3,3,...,19,19]
                               expected 20 to be 40     → 20 of 20 rounds collided

appendMessages × recordAnswer  duplicate seqs: [0,0,1,1,...,9,10,11,11,...,20,20]
                               expected 21 to be 40     → 19 of 20 rounds collided
```

Your reasoning about why my first fix was incomplete was exactly right. Moving the batch path's allocation inside its insert serialised writers that take *that* lock, and nothing else. `putMessage` and `recordAnswer` still computed `max(seq) + 1` in an insert holding no lock:

```sql
INSERT INTO vendo_thread_messages (thread_id, id, seq, message, created_at, updated_at)
SELECT t.id, $2,
       COALESCE((SELECT max(m.seq) + 1 FROM vendo_thread_messages m WHERE m.thread_id = t.id), 0),
       $3::jsonb, $4, $4
FROM vendo_threads t WHERE t.id = $1        -- a plain read: takes no lock
ON CONFLICT (thread_id, id) DO NOTHING
```

…and only touched the thread row *afterwards*. `FROM vendo_threads t` is a plain read, so it is not blocked by the batch path's row lock, and its `max(seq)` comes from its own READ COMMITTED snapshot, which does not contain the other transaction's uncommitted rows. Both write the same number.

**The fix, and it is not a special case for the batch path:** `touchThread` — the `UPDATE vendo_threads … WHERE id = $1` that takes the row — now runs **first** in both `putMessage` and `recordAnswer`. Every transcript writer therefore takes the same two locks in the same order: thread row, then message rows. The loser blocks on the thread row until the winner COMMITs, then allocates on a fresh snapshot that already holds the winner's rows. Uniform lock ordering is also why none of these writers can deadlock against another.

No error path moved: an absent thread updates nothing in `touchThread` and is still reported by the checks below the insert. `ask-user-answer` (which is `recordAnswer`) and `message-loss` are green on both legs.

**After:** all three pairings pass, three consecutive runs each.

**The race test now runs a table of rivals, not a hard-coded second call** — `appendMessages`, `putMessage`, `recordAnswer` — because the lesson of round one is that a race test proves only the pairing it actually runs. `packages/store/tests/thread-messages.batch.test.ts`.

**On the skipped-suite worry: CI is fine, and the repo already knew about this hazard.** `backends()` does silently fall back to PGlite when `POSTGRES_URL` is unset, but `.github/workflows/ci.yml` provisions a `postgres:16` service and sets a job-level `POSTGRES_URL` on all four jobs that need it — `test-shards` (:97, which is where `packages/store` runs, in 3 shards), `test-rest` (:177), `integration` (:290) and `conformance` (:426). The workflow even documents the trap at :386-387 for the MCP token-claim race: *"gates on POSTGRES_URL and otherwise silently skips — the job-level POSTGRES_URL above makes it actually run its forced-overlap races here."* So these new races really do run in CI. One nuance worth recording: CI is Postgres **16**, my proof was **17.11** — the bug is READ COMMITTED snapshot semantics, identical in both, and the assertion is an invariant rather than a version.

**Re-gate:** store 260 passed / 9 files (pglite + real Postgres) · vendo 89 passed · `build` 21/21 · `typecheck` 42/42 · `lint` 0 errors.

PR #1279, opened 2026-08-14
perf(guard,store): client read diet

yousefh409:
Slice 6 of the fast-datastore project — the client stops re-reading, per tool call, what it already knows. Three of the five items landed, one stopped on a load-bearing comment, and one (C) was reverted in review because it opened a permissions window. Net: ~3 fewer datastore calls per tool call.

## A. Freeze flag: cached at check time, fresh at the gate
`packages/guard/src/guard.ts` — `frozen()` gains `{ cached?: boolean }`, and only `#checkWithMetadata`'s read passes it. The pre-execute gate at the dispatch site is byte-identical: it still reads the store, so the invariant its comment spells out ("a freeze that lands during the judge's 15s window must still block execution") is unchanged. Every fresh read — this guard's own `freeze()`/`unfreeze()` included — refreshes the cached value, so an in-process flip is visible at once and only another process's flip can be up to 10s stale, and only for a DECISION.

**Deviation from the blueprint**: the option is `{ cached: true }` opt-in, not `{ fresh: true }` opt-out. A default-cached `frozen()` breaks `freeze.test.ts` -> "obeys a flag row written straight through the store, as the console writes it", which flips the row out of process and expects the public `frozen()` to answer the store. Code wins.

**Calls per tool call: 3 freeze reads -> 1** (2 on the first call of a 10s window).

## B. Grants filtered server-side
`refs: { subject, tool: call.tool }`; the routed door maps the ref to the indexed `tool` column, and the JS `continue` is gone. `invalidated` is unaffected — it only ever collected same-tool grants. The response also shrinks from every grant the subject ever held to that tool's.

## C. The preview pass shares its grant read with the real pass — REMOVED in review
Shipped and then reverted, on the reviewer's finding: a grant revoked (or expired) between the preview and the real, dispatching pass was not seen, so the tool ran on a permission the person had already taken back. A grant is not a decision input the way a rule is — it IS the authority the call executes on, so the preview-to-dispatch gap is exactly the window the kill switch's re-read comment exists to close, and it gets the same answer: read it again. The pipeline reads the grants for the real pass, and `grant-filter.test.ts` now pins that a revoked grant parks instead of running. No revocation-aware cache was attempted: one wire call per tool call does not buy that window.

The replay read is unshared for its own reason — the single-use CAS spend belongs to the real pass.

## D. Workspace: true no-op commit + one batched commit
`commit()` returns early when nothing is staged and nothing is removed (the harness calls it after every tool call). The remove pass and the land pass collapse into a new `commitAll(owner, entries, message?)` on the rows adapter: the hosted backend sends ONE `workspace.commit`, the SQL backend keeps its per-path statements and just wraps them. Entries are grouped by OWNER, because the commit verb addresses one drawer — which is also what keeps an org conflict from taking the same turn's `/user` edit down with it.

Preserved exactly: per-entry `expectedRevision` including the null create-only guard, and a stale one still refusing the WHOLE commit with `conflict`. The wire error carries no path list, so the hosted backend re-reads the heads on the conflict branch and re-runs the server's own comparison to name the paths that moved.

**Calls per turn (hosted): one commit per removal + one per landed file -> 1 per owner touched.**

### What this does NOT do
The ordering fix is ORDERING, not atomicity, and the residual is real: a lost swap can still leave earlier `/user` writes from the same commit applied — that is last-write-wins, which those mounts already promise, and it is not data loss. Genuine atomicity would need the row statements inside one transaction, and the blocker is the blob deletes: `trim` and `remove` release objects through the files adapter inside that scope, so a rollback would leave a row pointing at an object that is already gone. That is a separate lane, deliberately not taken here. The hosted backend is already atomic — one wire commit, refused whole, server-side.

### Pre-existing, not a regression
The deletion loss is NOT introduced by this branch. The same reproduction fails on `origin/main`: the per-path loop that `commitAll` replaces also removed before it landed, and the loss was silent there too. This branch's batching neither caused it nor hid it — it is fixed here because it breaks this slice's own "a stale revision refuses the WHOLE commit" requirement, and because the reproduction was in hand. Attribute it to the old path, not to the batch.

## E. Amortized sweep — STOPPED, not done
`packages/vendo/src/server.ts` is untouched. The ordering comment there calls itself LOAD-BEARING and pins the awaited sweep as step 1 "before anything", and the sweep's own comment gives the reason: "Awaited BEFORE the request is handled, so a request arriving past a TTL sees the swept state rather than racing its own sweep." Detaching the amortized leg is exactly the race that comment closes — an expired parked approval could still be answerable by the request that triggered the sweep. `byo-approvals-wire.test.ts` -> "expires an orphaned parked call via the on-request sweep and serves expired" depends on it. Reporting rather than reordering a documented invariant.

## Tests
New seams (`packages/guard/test/`):
- `freeze-cache.test.ts` — the flag is flipped through the real store write path (the console's own path). A check-time read is served from the cache with the store untouched, AND a freeze landing inside a judge's await still refuses the dispatch. **Prove-it-can-fail**: pointing the pre-execute gate at the cache turns both cases red, the second on `{status:'ok'}` — the tool ran while frozen.
- `grant-filter.test.ts` — grants for several tools written through the real store; the emitted SQL carries `tool = $n`, the selected grant is identical to what the old JS filter would have chosen off the unfiltered list, and a grant revoked between the preview and the real call parks instead of running (finding 1's regression).

Regression test for the deletion loss, in the suite that owns the conflict branch:
- `packages/store/tests/workspace.orgs.test.ts` — "a conflicted commit leaves the deletions in it unapplied and still staged". **Prove-it-can-fail**: restoring the pre-fix ordering turns it red on the surviving file.

Green: whole `@vendoai/guard` suite (288), whole `@vendoai/store` suite (569 — workspace, workspace.erase, flip-guards.conformance included), whole `@vendoai/harnesses` suite (550), `packages/vendo` org-policy + byo-approvals-wire + sweep-failure-isolation (20). Root gate `pnpm build && pnpm typecheck && pnpm lint` — 21/21, 42/42, 4/4.

One unasked-for change, on the record: `commit()`'s permission gate moved verbatim into `assertCommittable()` because the new branch pushed the method past the repo's cognitive-complexity ceiling. Behaviour-preserving — a pure move of the `canCommit`/`canView` block with its comments, no condition altered — and covered by the full store suite re-run after it.

---
## Summary by cubic
Reduces client reads in guard checks and workspace commits while preserving gating and CAS semantics. Fixes two correctness gaps: a revoke between preview and execute now blocks, and a conflicted commit no longer removes files.

- Guard
  - `frozen({ cached: true })` adds a 10s cache for the check-time freeze read; the pre-execute gate still does a fresh read. In-process `freeze()/unfreeze()` refresh the cache. 3 reads → 1 (2 on the first call in a window).
  - Grants list filters server-side with `refs: { subject, tool }`; `invalidated` stays same-tool only.
  - Preview vs. execute: the real pass re-reads grants; replay is also unshared. A revoke/expiry between preview and execute now blocks as expected.
  - Tests: `freeze-cache.test.ts` and `grant-filter.test.ts` cover caching, server-side filtering, and preview-to-execute revocation.

- Store
  - `WorkspaceStoreFs.commit()` returns early when nothing is staged or removed.
  - Per-owner batching: `WorkspaceRows.commitAll(owner, entries, message?)` sends one `workspace.commit` on hosted backends; SQL keeps per-path statements. Per-entry `expectedRevision` is preserved.
  - Conflicts and deletions: writes land first; any CAS conflict returns `conflict` and skips deletions. The façade keeps deletions staged for retry. Changed files only include paths that wrote new bytes. One commit per removal/landed file → 1 per owner.
  - Test: `packages/store/tests/workspace.orgs.test.ts` pins that a conflicted commit leaves deletions unapplied and still staged.

<sup>Written for commit 63dd94caf7565027eda400dfeb47d2130f6530d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1279?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/guard/src/guard.ts
+++ b/packages/guard/src/guard.ts
@@ -125,6 +125,14 @@ const FREEZE_ROW = "freeze";
 /** The block a frozen guard returns — the same words at the check and at the
  *  execute re-read, so the two agree. */
 const FROZEN_REASON = "vendo is frozen — nothing runs until it is unfrozen";
+/** How long a CHECK-TIME freeze answer may be reused (the execute gate never
+ *  reuses one). A freeze flipped in another process therefore starts blocking
+ *  new checks within this window, and blocks every DISPATCH immediately. */
+const FROZEN_CACHE_MS = 10_000;
+/** How long a preview pass's grant read may be handed to the real pass that
+ *  follows it. The two run moments apart for one logical call; anything older
+ *  is a different moment and reads the store again. */
+const PREVIEW_SHARE_MS = 10_000;
 /** Build contract §7 — the effect ledger: one row per completed mutating call,
  *  keyed by (run, tool, exact input). It is what makes fail-and-re-run correct:
  *  a re-run of a run that already sent the payment must not send it again. */
@@ -175,6 +183,14 @@ interface DecisionMetadata {
   decision: DraftDecision;
   rationale?: string;
   invalidatedGrants?: PermissionGrant[];
+}
+
+/** What the subject's standing grants say about one call: the one that
+ *  authorizes it, and the same-tool grants a descriptor change has invalidated
+ *  (which the park offer names). */
+interface MatchedGrant {
+  grant?: PermissionGrant;
+  invalidated: PermissionGrant[];
 }
 
 interface CompletedDecision {
@@ -491,6 +507,12 @@ class GuardImplementation implements VendoGuard {
   readonly #callWindows = new Map<string, number[]>();
   readonly #writeCounts = new Map<string, { count: number; touchedAt: number }>();
   #lastSweepAt = 0;
+  /** The last freeze answer a fresh read produced (see {@link frozen}). */
+  #frozenCache: { at: number; value: boolean } | undefined;
+  /** The last preview pass's grant read, waiting for its real pass. One entry:
+   *  the caller that previews executes moments later, so a second preview means
+   *  the first one's call is gone. */
+  #previewedGrant: { key: string; at: number; matched: MatchedGrant } | undefined;
   readonly #approvalCallbacks = new Set<(id: ApprovalId, approved: boolean) => void>();
   readonly #approvalRequestedCallbacks = new Set<(request: ApprovalRequest) => void>();
 
@@ -874,7 +896,21 @@ class GuardImplementation implements VendoGuard {
     await this.#setFrozen(false, by);
   }
 
-  async frozen(): Promise<boolean> {
+  /** Reads the flag row every time, unless the caller says a slightly stale
+   *  answer will do (`cached`). Only the CHECK-TIME read does: it is one static
+   *  row, read on every tool call, and the gate immediately before dispatch
+   *  (`bind().execute`) is uncached — so the freeze a cached check missed still
+   *  cannot reach a tool. Every fresh read refreshes the cached value, this
+   *  guard's own freeze()/unfreeze() included, so an in-process flip is visible
+   *  at once and only another process's flip can be up to
+   *  {@link FROZEN_CACHE_MS} stale. */
+  async frozen(opts?: { cached?: boolean }): Promise<boolean> {
+    if (
+      opts?.cached === true && this.#frozenCache !== undefined
+      && Date.now() - this.#frozenCache.at < FROZEN_CACHE_MS
+    ) {
+      return this.#frozenCache.value;
+    }
     let record: VendoRecord | null;
     try {
       record = await this.#engine.get(CONTROLS_COLLECTION, FREEZE_ROW);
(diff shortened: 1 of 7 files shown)


yousefh409:
Both findings reproduced with executable tests before any fix, and both were real. Triage push 1 of 2: `b107712af`.

**Finding 1 — a grant revoked after the preview still ran the tool. Confirmed, fixed by reverting the grant half of item C.**

Reproduction (now a regression test in `packages/guard/test/grant-filter.test.ts`): a standing grant on a destructive tool, `previewCheck` → `run` by grant, revoke through the real store, then `bind().execute`. Before: `{status:'ok'}` — the destructive tool executed on a permission the person had already taken back. After: `pending-approval`, `tools.executions` empty.

The review's framing is the right one and the blueprint's was wrong: a grant is not a decision input the way a rule is, it IS the authority the call executes on, so the preview-to-dispatch gap is exactly the window the kill switch's re-read comment exists to close. No revocation-aware cache was attempted — one wire call per tool call does not buy a window in which a revoked permission authorises execution.

The replay read stays unshared, unchanged.

**Finding 2 — a conflicted batch deleted a file and then returned `conflict`. Confirmed, and it predates this branch.**

Reproduction: two `/orgs` files, one deleted and one overwritten against a head a colleague already moved. The commit answers `conflict` — and the deleted file is gone. The same reproduction fails on `origin/main` (the old per-path loop also removed before it landed), so this is not a regression from the batching; but it breaks this slice's own "a stale revision refuses the WHOLE commit" requirement, so it is fixed here.

The fix, plainly: the SQL backend lands every write FIRST and applies the tombstones only if no swap was lost, and the façade keeps deletions staged when the commit was refused so the re-apply still carries them. **This is ordering, not atomicity.** What it costs: a conflict can still leave earlier /user writes applied, exactly as the old per-path loop did — that is last-write-wins, not data loss. Genuine atomicity would need the row statements inside one transaction, and the blocker is that `trim`/`remove` delete BLOBS through the files adapter inside that scope: a rollback would restore a row pointing at an object that is already gone. That is a bigger change than this triage, and worth its own lane. The hosted backend is already atomic — one wire commit, refused whole, server-side.

**Not covered by a committed test:** finding 2's regression lives in `packages/store/tests/workspace.orgs.test.ts`, which is outside this slice's file fence, so the reproduction was run as a throwaway and deleted. Say the word and it goes in as push 2.

**Calls per tool call after both fixes:** freeze 3 → 1 (2 on the first call of a 10s window); grants back to 2 (the item C saving is gone); workspace commit unchanged at 1 per owner touched instead of one per file. Net ≈3 fewer wire calls per tool call, down from ≈4.

Gate: guard 288 ✓ · store 568 ✓ (twice — before and after the complexity refactor) · harnesses 550 ✓ · vendo org-policy + byo-approvals-wire + sweep-failure-isolation 20 ✓ · root `build && typecheck && lint` 21/21, 42/42, 4/4.

PR #1290, opened 2026-08-15
feat(core,store): StoreOps.retention gets an engine

yousefh409:
`StoreOps.retention` was declared in #1288 and served by nothing. The local
backend serves both verbs now, over a quarantine table of its own.

- **`quarantine(collection, olderThan)`** lifts every row whose own age field
  predates the cutoff OUT of the live collection, and answers how many moved.
  Re-running it moves nothing.
- **`purge(collection, quarantinedBefore)`** destroys what was lifted before
  ITS cutoff — measured from the lift, not from the row's age, because the
  grace it honors starts when the row left.

The gap between the two is the whole feature: a retention window that turns
out to be wrong is recoverable right up until the purge. Host SQL on a host's
own cron still works and always did (the table map is public); what this adds
is the recovery window a `DELETE ... WHERE at < ...` cannot express.

## Schema v9 — `vendo_quarantine`

Engine-owned: no caller names it, no collection maps to it, and `purge` is the
only way back out.

- It holds each lifted row **verbatim** (`to_jsonb` of the live row, whichever
  of the thirty-odd tables it came from), not the door's `VendoRecord`
  projection, which is lossy for several of them. A quarantine that cannot put
  back exactly what it took is a delete with a longer name.
- The lift is ONE data-modifying CTE, so a row is never in both places and
  never in neither — without holding a transaction open across however many
  million rows a first sweep moves.
- A table of its own rather than a `quarantined_at` column on every collection:
  the latter means every read in the store grows a `WHERE quarantined_at IS
  NULL` it can never be trusted to remember.
- The key carries `quarantined_at`, so a row re-created after a sweep and swept
  again is a second quarantined copy rather than an overwrite of a first that
  is still inside its own grace.
- The age column each collection is swept on is the door's own `cursorColumn`,
  read off the door's config rather than copied into a second table, so a sweep
  and a reader cannot come to disagree about what "older" means.

## The erase cascade reaches it

A quarantined row is still its owner's data. The lift copies the row's subject
and app id into columns on the way in (from the typed column or from `refs`,
whichever the table has), and both cascades match them — plus the app-history
collection, whose rows carry their app only in the collection NAME and no refs
at all, which is the same blind spot that once hid the version log from
`byApp`. Without this, quarantining would be a way for data to outlive an
erasure, which is the exact hole `vendo_effects` sat in until it was given a
subject. `ERASE_TABLES` gains the table and `retention.ops.test.ts` proves both
legs on real quarantined rows.

## Two collections are refused, not half-swept

`vendo_threads` and `vendo_apps` answer `blocked` on both verbs. A thread's
transcript (`vendo_thread_messages`) and harness state (`vendo_state`) and an
app's whole drawer (records, blobs, state, runs, grants) live in other tables,
so lifting the row alone strands them — and for an app, `createRecordStore`'s
gate then refuses writes to rows nothing can read or erase. Ageing either of
those out is `lifecycle.erase`'s job (or `transcripts.deleteThread`'s), and the
refusal says so. Same shape as `engine.claim`'s existing per-collection
refusal.

## The pending cases

`status().ops` goes 42 → 44, and the two conformance cases that shipped tagged
`pending` lose the tag — one deleted field each, as the contracts lane
designed. `memoryStoreOps()` serves the family too, so the kit runs them
against its own reference.

The harness that proved those cases executable before any engine existed
(`retention-pending.test.ts`) is renamed to `retention-cases.test.ts` and keeps
the half a green suite never shows: each case is still held to FAILING against
an implementation that breaks its rule — now wrapped around the shipped memory
retention rather than a parallel copy of it, so the proofs bind to the code
that runs.

## Gate

`pnpm build` · `pnpm test:affected` (45 tasks green) · `pnpm typecheck` ·
`pnpm lint` — all green locally, plus the full `@vendoai/store` (740) and
`@vendoai/core` (680) suites.

Mutation-checked, not just green: reverting the purge cutoff to a fixed window
turns the purge conformance case red with its own message
("a purge cutoff predating the sweep should destroy nothing, destroyed 2"),
and the three broken implementations in `retention-cases.test.ts` are that same
proof for the other rules.

---
## Summary by cubic
Serves `StoreOps.retention` from the local engine with a recoverable quarantine window. Previously the contract existed but the engine omitted it; now `quarantine` lifts rows older than a cutoff into an engine-owned quarantine and `purge` destroys rows quarantined before its cutoff. `status().ops` rises from 42 to 45.

- Adds schema v9 and `vendo_quarantine` that stores lifted rows verbatim with `subject`/`app_id`; a single DML CTE guarantees rows are never in both places or in neither. Indices support purge and erase selectors.
- Extends the erase cascade to delete quarantined rows by subject and by app (including app-history via collection name).
- Makes sweeps door-accurate: uses each door’s `ageColumnOf`; fences `vendo_state` so only app state is swept; supports reserved and generic collections.
- Blocks `vendo_threads` and `vendo_apps` (`blocked` on both verbs) to avoid stranding dependent data; use `lifecycle.erase` or `transcripts.deleteThread` instead.
- Implements `retention` in the `memoryStoreOps` reference; removes pending tags from the two retention conformance cases; tests updated for schema v9 and the new level.

**Rollout**
- Run `ensureSchema` to migrate to v9 before deploying `@vendoai/store`.
- No caller action is required; the quarantine is engine-owned and unnamed. Use `lifecycle.erase` or `transcripts.deleteThread` for threads/apps.

<sup>Written for commit ae247797a91690092b35f0277fe351f6752dc91d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1290?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Greptile's P1 is real and it landed a fix.

Quarantining `vendo_state` lifted BOTH of that table's tenants: an app's own state (`app_<id>:<subject>`, which the routed door addresses and a caller can name) and thread harness CONTINUITY (`harness_state:<threadId>`, which reaches the table only through `ops.harness`). A caller sweeping the collection cannot see the second one, so lifting it took a live conversation's session away with nothing in the request that asked for it — the invisible half of the same mistake `vendo_threads` is refused for.

The sweep now sees exactly what the collection's own door sees: the predicate is the door's own id grammar (`APP_ID_SEGMENT`, routing.ts), which is the same namespacing that already spares harness rows from app deletion (harness-state.ts).

Regression test added and proven to fail without the fix: against the unfenced sweep it reports `expected 2 to be 1` and the thread's session reads back null.

yousefh409:
Rebased onto #1291 (audit.tally). One conflict, in the local engine's `status()`, and the resolution is the interesting part:

`audit.tally` is declared AFTER `status` precisely so a level that stops short of retention cannot claim it. With retention served, the prefix runs to the end of the list for the first time — so this engine reports **45**, and the assertion is now pinned to `Object.keys(STORE_WIRE_PATHS).length` rather than a literal, since "everything on the manifest" is what it is actually claiming.

Issue #1252, opened 2026-08-13
Connect dock retries failed status polls with no backoff, sustaining any rate limit it meets


amrib247:
## Summary
The connect dock's status poll retries failures immediately, with no backoff, until its 120s deadline. Field measurement (2026-08-13): with the org rate-limited, one open tab polled `GET /connections` at ~2 requests/second; each poll is itself a Cloud broker call, so the failing poll **sustained the very rate limit that was failing it** — a self-feeding outage that outlived the original cause. (First noted as a rider on the F16 verdict; promoting it to its own issue after watching it keep an org throttled in the field.)

## Affected
`packages/ui` connect dock / connections polling, v0.8.0 and main (the poll loop predates the baseline).

## Suggested fix
Exponential backoff with jitter on poll failures (e.g. 1s → 2s → 4s … capped at ~15s within the existing 120s deadline), and treat an `unavailable`/429-mapped answer as a signal to back off harder, not retry sooner. Surfacing the failure to the user earlier than the deadline (the other half of the original F16 note) remains worthwhile but is secondary to not amplifying the outage.

PR #1311, opened 2026-08-15
fix(ui): back off the connect dock's failed status polls

samuelmbabhazi:
## What

`completeConnection`'s status poll backs off exponentially on failures: 1s, 2s, 4s, capped at 15s, with jitter, and floored at the healthy 1.5s cadence so a failing poll never fires faster than a succeeding one. A rate-limited answer (the wire `unavailable` code) jumps straight to the cap instead of climbing to it. A successful answer resets the backoff, every wait clamps to the existing 120s deadline, and waits run in short slices so a cancel is honoured mid-backoff rather than after a full 15s sleep.

Since a capped wait can blind the loop for up to 15s, the dock now checks status one final time before throwing the coded timeout: a connection that went active inside the last backoff window is reported as connected instead of "nothing changed".

## Why

As measured in #1252, a rate-limited org saw one open tab poll at about 2 requests per second for the full 120s deadline. Each poll is itself a Cloud broker call, so the failing poll sustained the very rate limit that was failing it, and the outage outlived its cause.

Refs #1252 (kept open: the issue's numbers were measured on `GET /connections`, which this PR does not touch; this fixes the per-connection status poll half)

## Verification

- [x] `pnpm build && pnpm test:affected && pnpm typecheck && pnpm lint` green
- [ ] Screenshots attached (if UI-affecting) — timing behavior only, nothing rendered changes

New tests in `connect-backoff.test.ts` drive the poll with fake timers and assert the exact upper-bound delay ladder (1.5s, 2s, 4s, 8s, 15s, 15s), the straight-to-cap unjittered jump for an `unavailable` code, the reset to the healthy cadence after a successful answer, the intact coded 120s timeout, the final pre-timeout rescue, and a cancel honoured mid-backoff. The jitter itself is covered at both bounds: with `Math.random` pinned low, the third failure must fire at backoff/2 (2s, not 4s), so deleting the jitter term fails the suite rather than being silently tolerated. The healthy pending cadence is asserted unchanged at 1.5s. Mutating the source (deleting the jitter, or the final check) makes exactly the corresponding test fail while the adjacent connect suites stay green (35 tests across connect, connect-in-thread and client).

Note on `test:affected` locally: `demo-bank#test` fails on any non US Pacific machine independently of this change (the planted charge is seeded at 01:14 machine-local time, so on UTC+2 it lands on the previous UTC day and leaves the pinned same-day window), and `@vendoai/ui#test:ui` needs the Playwright browsers installed. The first is unreachable from this diff (which touches only `packages/ui` and a changeset); the second is a missing local browser install, and the touched suites above run green.

### DIFF
--- a/.changeset/connect-poll-backoff.md
+++ b/.changeset/connect-poll-backoff.md
@@ -1,0 +1,5 @@
+---
+"@vendoai/ui": patch
+---
+
+The connect dock's status poll now backs off exponentially on failures (1.5s floor, 15s cap, with jitter) instead of retrying on the healthy cadence until the 120s deadline, and a rate-limited answer (the wire `unavailable` code) jumps straight to the cap. A failing poll no longer sustains the very rate limit that is failing it. Before giving up at the deadline, the dock checks status one last time, so a connection that went active during the final backoff wait is reported as connected instead of timed out.
--- a/packages/ui/src/chrome/connect-dock.tsx
+++ b/packages/ui/src/chrome/connect-dock.tsx
@@ -17,6 +17,8 @@ import { toolkitDisplayName } from "./humanize.js";
 
 const POLL_INTERVAL_MS = 1_500;
 const POLL_DEADLINE_MS = 120_000;
+const POLL_BACKOFF_BASE_MS = 1_000;
+const POLL_BACKOFF_CAP_MS = 15_000;
 const POPUP_WIDTH = 520;
 const POPUP_HEIGHT = 680;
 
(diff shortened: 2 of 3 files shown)


yousefh409:
The core backoff here is a real win (~5.7× less load on the failing poll). A few things before merge:

1. **Drop the `client-impl.ts` hunk and the `status === 429` check.** Nothing in the repo emits a 429 to the browser — it surfaces as a 501/503 — so that branch is dead, and it makes `throwWireError` inconsistent with the other three wire parsers. Keep `code === "unavailable"` and sequence this behind #1435 (which makes the wire answer 503).
2. **Make the jitter test real.** It pins `Math.random()` to exactly the value where the jitter term cancels to zero, so deleting the jitter entirely keeps the suite green — the thundering-herd defense isn't actually covered.
3. Consider a final status check before the timeout throw — there's an 11–15s window where a connection that actually succeeded gets reported as "nothing changed."

One note: #1252 measures `GET /connections`, which this PR doesn't touch (it backs off `/connections/{id}`) — I'll leave that issue open. Solid direction overall.

samuelmbabhazi:
Thanks for the careful read. All three addressed:

1. Dropped the client-impl hunk and the 429 check. `code === "unavailable"` is now the only rate-limit signal, so this can sequence behind #1435 cleanly.
2. The jitter is now tested at both bounds: with `Math.random` pinned low, the third failure must fire at backoff/2 (2s instead of 4s), so deleting the jitter term fails the suite.
3. Added the final status check before the coded timeout, with a test where the connection goes active inside the last capped wait and is reported as connected.

Also switched the body to a non-closing ref since #1252 stays open for the `GET /connections` half.

PR #1323, opened 2026-08-15
feat(core,store): the meter — storage behind per-user usage limits

yousefh409:
PR1 of a stack. Base: `main`. **This is a review window — the stack merges once, at the tip of PR2 (`yousefh409/limits-feature`), which is based on this branch.**

## What this adds

The counting layer for per-user usage limits: `LimitAction`/`LimitWindow`/`LimitDecision`/`LimitsCallback` in `packages/core/src/limits.ts`, and `StoreOps.usage` — `record`/`count`/`tally` — over a new `vendo_usage` table, local ops, conformance cases, and the hosted-client wire.

This PR is **storage only**. A host will supply a `LimitsCallback`; Vendo will count usage and enforce the host's decision. No policy, no enforcement, no config surface yet — that's PR2.

## Events, not buckets

`vendo_usage` is a table of timestamped events — one row per metered action — never pre-bucketed. A policy authors its own window ("20 messages an hour", "3 generations today", down to the minute), and a pre-aggregated count can only answer the periods whoever bucketed it happened to pick.

## Three deliberate design calls

1. **`vendo_usage` is NOT in `RESERVED_COLLECTIONS`.** That list opens a public records door (list/get by collection name), and a meter row is only ever counted, never listed. Follows the `vendo_quarantine` precedent — also absent from that list for the same reason.
2. **No retention list entry, on purpose.** Retention already refuses to sweep the meter structurally: `assertSweepable` calls `assertEngineCollection` first, which throws before a quarantine/purge on a collection with no door — `vendo_usage` has none. Adding a list entry would be unreachable code. `packages/store/tests/usage.ops.test.ts` has a test proving the refusal (`retention.quarantine`/`retention.purge` both reject with `code: "blocked"`).
3. **`StoreOps.usage` is OPTIONAL**, matching the `retention` precedent in the same interface — making it required would break BYO `StoreOps` implementers. The follow-up PR refuses a configured limits callback against a meterless store rather than failing open.

## Gate

Test evidence is **inherited from the builder's own gate on these four commits, not re-run by me**: `packages/core` 219/219, `packages/store` 280/280, plus `pnpm build`, `pnpm typecheck`, `pnpm lint` all exit 0. I did not re-run the suites myself — the repo's own rule makes the PR's green `ci` check the gate of record, and CI will run the full suite.

---
## Summary by cubic
Adds an events-based meter for per-user usage limits and serves it via `StoreOps.usage` over a new `vendo_usage` table. This enables counting usage windows without changing runtime behavior yet; policy and enforcement come next.

- Schema v10 adds `vendo_usage` with indexes; run your schema migration or `ensureSchema` before deploying.
- The wire protocol appends `usage.record`/`usage.count`/`usage.tally` (status level is now 48). Mounts that don’t implement these return 501; this PR treats `StoreOps.usage` as optional.
- `usage.record` is idempotent-keyed; `count` reads one of subject or pool, with since-inclusive/until-exclusive bounds; `tally` groups by subject and action.
- `vendo_usage` is engine-owned and not listable: retention quarantine/purge refuse it; erase-by-subject deletes a user’s meter rows.
- The hosted client implements the new ops, sending instants as ISO datetimes and preserving idempotency on writes.

<sup>Written for commit ffd507fb8045c7e71a457f7b45d85188283e0357. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1323?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/core/src/conformance/memory-store-ops.ts
+++ b/packages/core/src/conformance/memory-store-ops.ts
@@ -13,6 +13,8 @@ import {
   type RecordInput,
   type RecordQuery,
   type StoreOps,
+  type UsageEvent,
+  type UsageTallyRow,
   type VendoRecord,
 } from "../store.js";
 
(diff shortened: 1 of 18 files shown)


yousefh409:
Landed via #1324. This stack merges at the tip — #1323 was the review window for the meter half, not a merge unit, so all four of its commits (schema v10 `vendo_usage`, `usage.record/count/tally`, the hosted client's meter, and the denial stream part) are inside the squash at 37ed8218e on `main` along with the feature half.

Nothing here is unmerged; closing rather than merging so the same commits don't land twice.

PR #1324, opened 2026-08-15
feat(vendo,core,ui,harnesses): per-user limits — Vendo counts, the host decides

yousefh409:
**Stacked PR — this is PR 2 of 2.** Base is `yousefh409-limits-meter` (#1323), which adds the store's usage meter. Review only the commits above that base; the stack merges at the tip.

## Scope grew, with sign-off

The limits work exposed a **pre-existing** bug in the harness: `packages/harnesses/src/wire.ts` mirrored every denied tool result as the AI SDK's `tool-output-denied`, which is the terminal state of an approval a *person* turned down. The SDK's history converter reads `approval.reason` unguarded, so a refusal nobody was asked about persisted with no approval and **killed every subsequent turn in that thread**. It reproduces with a plain guard block and no limits anywhere.

This PR fixes it at the write site (a blocked outcome now rides the typed-outcome shape `connect-required` already uses) and fixes three sibling sites in `packages/harnesses/src/turn-tools.ts` — unattended park, "check could not run", and "guard asked twice" — which could kill automation threads the same way. A person's own denial deliberately keeps `output-denied`, because it has the approval that state is about.

## What

A host can now cap what each of their users does, in their own logic. Vendo counts; the host decides.

```ts
createVendo({
  limits: async ({ user, action, count }) =>
    user.facts?.plan === "pro" || await count(action, { days: 1 }) < 20,
});
```

That callback is asked once before each metered action — a user message, or an app generation — and gets three things: the resolved `user`, which `action` is about to happen, and a `count` reader already bound to *that* user. Return `false` to refuse, or `{ allow: false, message: "..." }` to refuse and say why in your own words. Anything else allows it, and the meter records it.

`count(action, window?)` is a callback rather than a pre-computed number because nearly every policy reads the meter once, for one window; pre-computing every window a policy *might* ask about would mean a database query per action per call. The `window` takes `days`/`hours`/`minutes` (combined into a single lookback), or a `since` instant, or a `pool` name.

## Pools — capping a team, not just a person

A **pool** is a shared meter a user's usage also counts into: a seat pool, a team, an org. The auth preset grows a `pools` seam beside the existing `facts` seam, resolved off the same session decode, and its answer rides the request context to the policy:

```ts
auth: jwt({
  secret: process.env.SESSION_SECRET,
  user: (subject) => ({ display: "Mia", pools: { workspace: "ws_maple" } }),
})
```

`count(action, { pool: "workspace" })` then counts the whole workspace's usage rather than the one person, and an allowed action accrues to every pool that user belongs to. Counting a pool the user is *not* in throws rather than answering `0` — a zero from a meter that was never resolved would silently under-count every limit written against it, and fails the wrong way.

## The two choke points behave differently, on purpose

**A denied message costs nothing.** The check sits at turn entry, before the conversation thread is even resolved. A refused message performs no database read, no write, and **zero model calls**. The whole response is the limit card. This is the one that protects the bill.

**A denied generation lets the turn carry on.** This check wraps `vendo_make` — the single door an app is built through — and answers the agent with the same "blocked" outcome every other refusal on that registry uses. The agent is told what happened, so **the turn continues** and the agent can explain it in its own words, while the card the person reads is raised on that call's own stream. A refused build is something worth talking about; a refused message is not, because there is nothing yet to talk about.

## What the person sees

Both paths raise the same card in the conversation. It deliberately does **not** wear the failure look — no ✕, no danger colour, and a polite status rather than an alert — because a cap reached is not a breakage. Nothing went wrong; the request simply never ran.

The host's own sentence is what the person reads whenever the policy wrote one. The host set the cap, so the host is the only one who can say what it is or when it lifts. A policy that returned no message gets the chrome's own line, which claims only what it can know.

## The two failure modes, decided deliberately

**A policy that throws DENIES**, and logs `limits.callback_error`. A limits system that fails *open* stops limiting silently: the host keeps believing they have a cap while every user is quietly unlimited. That is strictly worse than a turn that was refused and said so out loud.

**A `limits` policy on a store that cannot count is refused at composition** — it throws where the deployment is built, not at request time. A store with no usage meter reads every user as zero, so no limit would ever be reached and every user would be unlimited. Failing loudly at boot is the only honest answer; silently counting zero is the trap.

## For the operator

`vendo.usage(query)` is the operator's read of the same meter — per subject and action, over one window — for a host's own backend job: an overage sweep, a usage table in an admin screen. A policy never uses this; a policy asks its own bound `count` and can never name another subject. On a store with no meter it refuses rather than returning an empty tally, for the same reason `emit` does: a billing sweep reading "no usage" would bill nobody.

## If you don't use it

With no `limits` key, **nothing is wired**. No limiter is constructed, the tool registry stays the exact same object it always was, and each choke point costs a single `undefined` check. A deployment that never asked for limits does not pay for them.

## Grant & re-run: what this change proves, and where the proof stops

> **Grant & re-run: what this change proves, and where the proof stops.**
>
> Proven by tests, end to end on the real composition: an unattended turn (`presence: "away"`, through `vendo.harness.stream`) whose call parks with nobody to tap it is refused, the turn finishes in words, and **the next turn on that same thread answers normally** — which it did not before this change: it died on rebuilding the thread's history, and so did every turn after it. Proven in the same test: the standing approval is still in the guard's pending queue after the refusal, `GET /approvals` lists exactly that call, and `POST /approvals/decide` with `remember: { scope: tool, duration: standing }` still succeeds against it and empties the queue. So the *ask* the refusal leaves behind is intact, discoverable and answerable, and nothing about the grant is read off the tool part this change touched.
>
> Not exercised by these tests: the full **Grant & re-run** journey as a person performs it — clicking Grant in the automations panel and then Re-run, and the re-run's step actually executing. Two boundaries explain why. First, the button: the tests call the same HTTP endpoint the button calls, but not the button, so the UI wiring itself is covered only by the existing automations-panel tests, not by this change's. Second, and more substantive: a granted permission authorizing a *later away run* is the guard's own law, not this code's — `presenceMatches` requires an automation's app-and-trigger-bound grant, so a grant minted from a bare unattended turn deliberately never authorizes one. That half lives in the automations engine (`run-execution.ts` → `consent.needsPermission` → `runs.rerun`), which reads the raw `ToolOutcome` and never passes through the wire mirror this change alters; its suite (89 tests, including the `needs-permission` run rows) passes unchanged.
>
> Stated plainly: "the refusal no longer kills the thread, and the standing ask survives it and can still be decided" is proven by running the real thing. "Clicking Grant then Re-run makes the step run" is unchanged-by-construction rather than re-proven here — it flows through code this PR does not touch. To close the gap, a verifier should drive the real app: trigger an automation whose step needs a permission nobody has allowed, confirm the run row offers Grant & re-run, click both, and confirm the step executes and the thread keeps answering afterwards.

## Proof

An independent verifier who built none of this walked the real demo host (`demo-bank`/Maple) in headless Chromium.

- **Message cap:** card renders with the host's own sentence; **zero model calls** on the denied turn — 85ms versus 241–357ms for allowed turns, and no model activity in the server log after the last allowed turn.
- **Generation denial:** card renders, the agent is informed, the turn continues to a normal prose reply, and **three further turns in the same thread all answered with 0 errors** (before the fix: 3/3 failed).
- **Regression surface:** a real `host_transferMoney` approval was denied by clicking Deny in the UI — the `output-denied` branch this change deliberately leaves alone. It still reads "you declined it" (correct — the person did decline) and two following turns ran clean.
- **Method limit, stated plainly:** the build machine has no Anthropic or Vendo key, so the real `@ai-sdk/anthropic` client was pointed at a local API-compatible stand-in over real HTTP. Real: HTTP, SSE, tool bridge, agent loop, guard, store, UI, browser. Scripted: only the model's word choice. The message-cap path needed no model at all and was fully real.

Screenshots from this run exist only on the build machine (not attached here — see Notes).

## Known and deferred

Three cosmetic items the verifier raised, decided deliberately:

- The doubled mark (beat ✕ above the card's ⊖) is being fixed in a follow-up commit on this same PR.
- "Regenerate" still appears under the cap card, inviting a retry that can only be refused.
- A denied message leaves no trace after reload.

The last two are existing thread-chrome behavior, not something this PR touches, and were consciously left out of scope.

## Verification

- Gate on the touched scope: `pnpm build` (21/21), `pnpm typecheck` (42/42), `pnpm lint` (0 errors) all pass. Scoped test files pass 80/80 — 47 in `packages/vendo`, 30 in `packages/core`, 3 in `packages/ui`. The full suite is CI's job.
- The pools lane is covered by a **seam test with nothing stubbed on either side** (`packages/vendo/tests/limits-pools.seam.test.ts`): one request driven from a real signed host session, through the real preset, the real composition, the real wire context resolver, the real limiter, into a real PGlite meter, then read back through the real read path. Break any hop and the gate fails closed, so the passing `allow` is what makes every hop load-bearing.
- The message choke's "zero model calls" claim is asserted directly, against a model scripted with exactly two turns: a third call would be exhaustion, not a pass.

## Notes

- One changeset covers the whole feature (`minor` on core, harnesses, ui, vendo) rather than one per commit, which would double-bump the same packages.
- Proof screenshots from the browser verification (`/tmp/limits-proof2-*.png`) live only on the build machine; this repo has no established `gh`-based convention for embedding images in a PR body (checked recent merged PRs), so they are reported in text above rather than attached.

---
## Summary by cubic
Adds host‑defined per‑user limits across `@vendoai/vendo`, `@vendoai/core`, `@vendoai/ui`, and `@vendoai/harnesses`. Previously every message and generation ran; now a `limits` callback gates user messages and app builds. Denied messages make zero model calls; denied builds return a typed `blocked` outcome and stream `data-vendo-limit`.

- API: `createVendo({ limits })` is called once per metered action with `{ user, action, count }`. Return `false` or `{ allow: false, message }` to deny; anything else allows and records usage. Callback errors and unknown pool names deny and log `limits.callback_error`.
- Counting and pools: `count(action, window?)` is bound to the request’s user. Auth presets may return `pools`; `ctx.pools` carries them. Allowed actions accrue to all of a user’s pools. Erasing a subject removes their meter rows, including pooled contributions.
- Choke points: message gate runs at turn entry (no thread write, no model call). Generation gate wraps `vendo_make`, emits `data-vendo-limit`, and settles with `blocked`; the agent continues and `convertToModelMessages` succeeds. Harnesses now mirror refusals as typed `blocked`, not `tool-output-denied`.
- Automation builds: limits skip the automation venue (`ctx.trigger`); automation‑fired builds are never gated or metered.
- UI: `@vendoai/ui` renders `data-vendo-limit` as a quiet status with the host’s sentence when provided, or a default line when not, and drops the duplicate mark under the limit card.
- Ops and composition: `vendo.usage(query)` returns meter tallies; on a meterless store it rejects. Composition refuses when `limits` is set without a meter. Unset `limits` wires nothing and adds only one undefined check.
- Migration: `LimitUser.email` was removed; read from `user.facts.email`. Omit `limits` to keep existing behavior.

<sup>Written for commit c2b33a12794ec3150bf62c0e972e015607c00fa5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1324?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Reopening to force CI to re-run against the new base (`main`) — retargeting alone does not fire a `pull_request` event. No code change; head stays at c2b33a127.

Issue #557, opened 2026-07-25
Wire cloud overrides into actions tool-enablement (defer eager compose-time capability-miss surface)


yousefh409:
Cloud-published overrides.json currently feeds app-generation semantics/domains (lazy, per-generation) but NOT the actions registry's tool ENABLEMENT (disabled/audience). The actions block accepts an optional overrides provider (shipped in Lane 3, additive, tested), but wiring it from cloud is blocked: the registry's first loadHost is driven at COMPOSE by missSurface = actions.descriptors() and memoized, so a cloud read there would fetch at module init (Workers global-scope violation) or lock an empty result before the snapshot warms.

Fix: make the capability-miss surface lazy (defer actions.descriptors() to first use behind a memoized promise, updating the createCapabilityMissCapture contract and the zero-live-tools warning timing), then inject the async cloud overrides provider (already prototyped) so enablement resolves boot-once on first request.

Security note: a cloud-disabled tool must NOT stay live on a cold first load — hence async/awaited resolution.

Boot-once only; live overrides hot-swap remains a separate descoped follow-up (turn-pinned versioned registry snapshot; the in-flight-turn race). Both deferred from the cloud-supported-everything wave (Lane 3).

yousefh409:
Note from Lane 3 quality review: when enablement IS eventually wired from cloud, it carries the same cold-boot-once hazard just fixed for semantics — a provider that resolves to a DEFINED value on a cold snapshot will lock the local-only value for the process lifetime. The async/awaited-first-request resolution described above is what avoids it (don't memoize a defined-on-cold result). Fold this into the fix.

yousefh409:
Fixed and merged (unified-auth / cloud-config backlog batch, 2026-07-26).

PR #1329, opened 2026-08-16
Config resolves in code, forever — the console only hears about it

yousefh409:
Per surface, resolution is now exactly: **a value passed in code → the local `.vendo/<name>` file → not set.** The console is out of resolution entirely. A keyed runtime instead REPORTS what it resolved to, one way and lazily.

## What was deleted

- **`packages/vendo/src/cloud-config.ts`** — the fetch-config-from-console client, and every call site, so console values no longer participate in resolution at all:
  - `compose-prompt.ts` — the brief resolver's cloud leg (`product` is no longer a per-turn cloud resolver)
  - `compose-surfaces.ts` — the theme, design-rules and merged-semantics providers
  - `compose-guard.ts` — the `policyCloudFallback` wiring
  - `compose-actions.ts` — the whole awaited cloud overrides-**enablement** provider (a keyed boot no longer blocks on a console fetch before it can serve a tool)
  - `compose-adapters.ts` / `compose-context.ts` — the `configCloud` adapter slot
  - `server.ts` — the `cloudConfig` / `CloudConfig*` public exports
- **`@vendoai/guard`'s `policyCloudFallback`** (`types.ts`, `guard.ts`, `PolicyResolver`) — nothing can fill it now, so it was dead machinery.
- **`config-surface.ts`** — the `cloud` input leg and the `"cloud"` member of `ConfigSurfaceOwner`.
- **CLI** — `vendo config push`, `vendo config pull` and the draft plane (`/api/v1/config/draft`), with the `--draft`, `--yes`, `--key`, `--api-url` flags. `vendo config status` is now LOCAL-ONLY: it reports `file` / `unset` off disk, makes no network call, needs no credential, and says nothing about a console.
- **`doctor-config-checks.ts`** — the check that downgraded a missing `.vendo` config file to a warning when `VENDO_API_KEY` was set. A missing file is a missing file for every deployment now. `checkSurfaceOwnership` reports `file` / `unset` instead of `file` / `runtime`.
- Docs describing the removed commands and the cloud leg.

## The reporter's trigger conditions

- **Only with a `VENDO_API_KEY`.** Keyless, `reportConfig` is a no-op closure — keyless open-source users report nothing, ever.
- **Boot's first resolution** sends the initial report (`createComposition`, right after `emitDeploymentBoot` — the first moment all five surfaces have resolved values).
- **Per resolution cycle** thereafter: the lazy surface providers in `compose-surfaces.ts` re-resolve all five surfaces and hash them (`sha256Hex(canonicalJson(report))`). A report is pushed **only when the hash differs from the last pushed hash**.
- **No heartbeat, no timer, no new transport** — it rides the existing `createBatchedUploader`, which gained one field (`method`) so it can PUT. `cloudKeyFetch` gained one line so a `204` is read as delivery rather than a parse failure. The deployment identity headers (`x-vendo-deployment-*`) already ride every keyed call and are reused, not reinvented.
- A batch collapses to its newest report — the console wants the config running now, not a history of it.

## The wire contract (frozen)

`PUT /api/v1/config/report`, key-authed, **204 No Content**.

```json
{"surfaces": {"<name>": {"source": "file"|"code"|"unset", "content": string|null}}}
```

All five surface keys (`design-rules.md`, `brief.md`, `theme.json`, `policy.json`, `overrides.json`) are ALWAYS present; `content` is null **iff** `source` is `"unset"`.

The fixture pair was committed **first** (`f503289`), at `packages/vendo/tests/fixtures/config-wire/` — the console repo commits the identical bytes. Per the repo's seam law, `tests/config-report.seam.test.ts` writes through the REAL path (real `.vendo` files + a real code-set knob → real resolution → real reporter → real batched uploader → the captured HTTP request) and asserts the captured request equals the request fixture. Nothing is stubbed between resolution and the wire.

## Tests deleted, and why

These covered exactly the behavior this change REMOVES:

- `packages/vendo/tests/cloud-config.test.ts` — the entire file (265 lines): the console config client's fetch, ETag/304 revalidation, and stale-while-revalidate snapshot. The client is gone.
- `packages/vendo/tests/config-surface.test.ts` — the cloud-fallback cases and the whole `two-path resolution` describe. Kept explicit/file/unset, and converted the "published-version flip" case into a file-edit-per-resolution case, which is the surviving invariant.
- `packages/vendo/tests/cli/config.test.ts` — all `config push`, `config pull`, `--draft`, and `.env.local` key-resolution cases, plus the cloud/unknown columns of `status`. Replaced with local-only `status` coverage.
- `packages/vendo/tests/cli/doctor.test.ts` — "degrades missing cloud-resolvable config files to warnings when VENDO_API_KEY is set". Its sibling was rewritten into "keeps missing config files fatal with VENDO_API_KEY set — a key is not a config source", which is the new law.
- `packages/vendo/tests/server.test.ts` — the whole `#557 cloud overrides tool ENABLEMENT` describe (5 tests). The per-generation semantics test was KEPT and converted to a local `.vendo/overrides.json`, since "re-resolves live per generation, not memoized" still holds.
- `packages/guard/test/policy.test.ts` — the `cloud policy fallback` describe (5 tests).

**Two test edits that are not deletions, stated out loud:** `server.test.ts`'s "wires the Cloud share/publish client" and "selects the store with the adapter-rule precedence" capture *every* fetch and assert on call counts. The boot report is new traffic on a keyed compose, so both now skip the report path when recording — the same isolation those tests already buy for the connectors seam with `connectors: []`. Their actual assertions are unchanged.

## Deviation from the spec, for review

The spec named `config push` and "the draft commands" for deletion and was **silent on plain `config pull`**. I deleted `pull` too: with `push` gone and console editing out of scope, nothing can ever populate the console's published config, so `pull` could only reach its own "not published — nothing to pull" error. Keeping it would leave a command that authenticates to the console for a value that can no longer exist. **Approved as an amendment, not an unflagged deviation** — decided by Yousef, recorded at `~/handoff/decisions/config-pull.md`: option A, no stub, since nothing exists for it to download and a production data check confirmed zero customer usage.

The spec's "existing 512KB/surface and 2MB/doc limits" do not exist in this repo (they live in the console repo, alongside the old `apps/console/fixtures/config-wire/`). The reporter sends the resolved bytes and the console enforces; client-side truncation would put bytes on the wire that are not what the deployment resolved.

## Gate

`pnpm build` ✅ · `pnpm test:affected` — `@vendoai/vendo` **2347 passed / 45 skipped (216 files)**, `@vendoai/guard` **323 passed**, run twice ✅ · `pnpm typecheck` **42/42** ✅ · `pnpm lint` **0 errors** ✅

`@vendoai/genbench` fails on this machine with `browserType.launch: Executable doesn't exist … chrome-headless-shell` — Playwright browsers are not installed here, unrelated to this diff.

Net **−1,506 lines**.

---
## Summary by cubic
Config now resolves only from code and local `.vendo` files; the console never participates. Previously: code → file → console; now: code → file → unset. Keyed runtimes only report what resolved; they never read from the console. A code brief of "" means no brief and does not fall back to disk.

- Deletes hosted-config and all readers, so boot no longer blocks on a cloud fetch and console publishes cannot change running deployments.

**Breaking changes and migration**
- Remove `cloudConfig` and its types from `@vendoai/vendo`; delete any `configCloud` adapter wiring.
- Remove `policyCloudFallback` from `@vendoai/guard`; policy resolves only from code or `.vendo/policy.json`.
- Delete `vendo config push/pull`; `vendo config status` is local-only and makes no network calls.
- Treat missing `.vendo/*` files as fatal; run `npx vendo init` or add the files.
- If you relied on console-owned surfaces, add programmatic values or check in the corresponding `.vendo/*` files.

**Reviewer notes**
- Reporter: keyed only (`VENDO_API_KEY`), sends at first resolution and on hash change, rides the existing batched uploader with `method: "PUT"`, no heartbeat.
- Wire: `PUT /api/v1/config/report` → 204; body includes all five surfaces with `source: "code"|"file"|"unset"` and `content` null iff unset. Defined code values remain `source: "code"` (e.g., `brief: ""` does not fall back).
- Transport: uploader stops retrying on 4xx; retries 5xx/transport failures. `cloudKeyFetch` returns `undefined` on 204 and throws errors carrying `status`.
- Docs/tests updated to remove hosted-config paths; new seam tests freeze the report wire. Keyless users do not report.

<sup>Written for commit c2403831555fbc08b2336e3d338aa78e7fee2a48. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1329?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
**Triage: REAL — accepted, with a correction to the proposed fix.**

The finding is right and it is this PR's miss. The diff rewrote `docs-site/reference/cli.mdx:633-652` and stopped exactly at the `Options:` heading (the diff's last context line), so the stale block below survived the rewrite.

Verified against source, not the docs:

- `runConfig` (`packages/vendo/src/cli/config.ts:52-62`) dispatches on `status` and nothing else; `runStatus` (`:38-50`) reads only `options.targetDir` / the first positional. No flag is parsed anywhere in the command, so `--draft`, `--yes`, `--key` and `--api-url` are silently ignored — matching your runtime artifact. The only surviving `--api-url` parsers are `vendo cloud`'s (`packages/vendo/src/cli/cloud/command.ts:37`, `cloud/auth.ts:89,128`), a different command tree.
- `--draft` and `--yes` referenced the push/pull draft plane, deleted in this PR along with `cloud-config.ts`.

One correction to the fix: **654-665 is the wrong range, in both directions.**

- It under-deletes. Line 666 (`generation picks it up per generation.`) is the tail of the same paragraph and would be left dangling under the heading.
- It over-deletes. The paragraph at `654-666` is not uniformly stale. `.vendo/overrides.json` still gates both app generation and tool enablement, and enablement is still boot-once — that is `OVERRIDES_ENABLEMENT_NOTE` (`packages/vendo/src/config-surface.ts:31-34`), which this PR keeps and which `config status` still prints to the user at `cli/config.ts:48`. Deleting the paragraph would leave the CLI printing a behaviour the reference no longer documents. What is actually false is the word *hosted* and the clause *"so a console publish applies on the next restart"* — a console publish no longer exists.

So the fix we will apply:

- Delete `cli.mdx:654-662` — the whole `Options:` block plus its trailing blank line. The command has no options.
- Rewrite `cli.mdx:663-666` to track the note the command prints: `.vendo/overrides.json` (not "hosted") gates both app generation (field semantics) and tool enablement (`disabled` / `audience`); enablement resolves boot-once on the first request, so **an edit** applies on the next restart, while generation picks it up per generation.

No option validation is being added: `vendo config status` ignores unknown flags the same way every other verb in this CLI does, and making one command uniquely strict is a CLI-wide decision, not a docs fix.

Not applied on this branch yet — batched into the next push with the rest of the review round. Thanks; this was a genuine catch that both the author and the local gate missed, since docs prose is not typechecked.

yousefh409:
Greptile re-reviewed `415b71d3` (the uploader fix) and returned the **same single finding**, escalated P2 → P1, with the range corrected to `cli.mdx:654-666` — which matches the correction above. No new findings on the uploader change. Verdict stands: one REAL docs defect, fix as described, batched into the next push.

yousefh409:
**Triage of Greptile review 3 (`3a62a221`): the `cli.mdx` finding is resolved and not re-raised. One new finding — REAL.**

**REAL — the config report can disagree with the running brief.**

Verified from source, not the artifact:

- `packages/vendo/src/compose-prompt.ts:31-36` — the runtime brief. When `config.profile.brief !== undefined`, the profile branch is taken and `"".trim() || undefined` yields `undefined`; it never reaches `selectConfigSurface("brief.md", …)`. The comment at `compose-prompt.ts:28-30` states the intent outright: an explicitly empty profile brief means "no brief" and never falls through to disk.
- `packages/vendo/src/compose-adapters.ts:89` — the reported value is `config.instructions ?? composed?.instructions ?? config.profile?.brief`, so `profile.brief: ""` passes `""` through, since `"" !== undefined`.
- `packages/vendo/src/config-report.ts:67` — `""` is `!== undefined`, so it is handed to `selectConfigSurface` as `explicit: ""`.
- `packages/vendo/src/config-surface.ts:59-63` — `input.explicit?.trim()` is `""`, which is falsy, so resolution falls through to the file and returns `{ value: file, owner: "file" }`.

Net: with `profile.brief: ""`, a `.vendo/brief.md` on disk, and a key set, the runtime prompt carries **no** brief while the report says `brief.md` is `source: "file"` with the file's bytes. Report-only — no runtime behavior changes — but the report is this PR's entire product, so a reachable wrong answer counts.

`profile.brief` is public surface (`packages/vendo/src/types.ts:285`), so this is reachable config, not a synthetic case.

**Scope: brief only.** I checked the other four before accepting the general form of the claim, and the blanket fix would be wrong:

- `design-rules.md` — `packages/vendo/src/compose-surfaces.ts:73` resolves `config.apps?.designRules?.trim() || config.profile?.designRules?.trim()`, so a blank explicit value **does** fall through to the file at runtime. The report already agrees. Making `config-report.ts` honor every defined code value verbatim would fix brief and break this one.
- `theme.json` / `policy.json` / `overrides.json` — objects rendered via `JSON.stringify` (`compose-adapters.ts:100-103`), so they can never produce a blank string.
- `instructions: ""` — `compose-prompt.ts:31` trims it to falsy and falls through to the file, and the report falls through identically. Already agrees.

So the rule genuinely differs per surface, and the fix belongs in the per-surface table, not in the shared seam.

**Fix** — move each surface's blank rule into `codeSurface` (`compose-adapters.ts:87-93`) where the per-surface knowledge already lives, and let a defined code value be authoritative in `config-report.ts:63-73`:

```ts
// compose-adapters.ts
const codeSurface: Record<ConfigSurfaceName, unknown> = {
  // Blank falls through to the file at runtime (compose-surfaces.ts:73).
  "design-rules.md": config.apps?.designRules?.trim() || config.profile?.designRules?.trim() || undefined,
  // An explicitly empty profile.brief means "no brief" and never reaches disk
  // (compose-prompt.ts:31-36), so it must report as code-owned-and-empty
  // rather than falling through to the file.
  "brief.md": (config.instructions ?? composed?.instructions)?.trim()
    || (config.profile?.brief === undefined ? undefined : config.profile.brief.trim()),
  "theme.json": config.theme ?? config.profile?.theme,
  "policy.json": isGuardInstance(config.guard) ? undefined : config.guard?.policy ?? config.profile?.policy,
  "overrides.json": config.profile?.overrides,
};
```

```ts
// config-report.ts — a defined code value already means code owns the surface;
// the blank rule is the caller's, because it differs per surface.
for (const name of CONFIG_SURFACES) {
  const code = options.codeValue(name);
  const { owner, value } = code === undefined
    ? selectConfigSurface(name, { readFile: options.readFile })
    : { owner: "explicit" as const, value: code };
  surfaces[name] = value === undefined
    ? { source: "unset", content: null }
    : { source: owner === "explicit" ? "code" : "file", content: value };
}
```

`profile.brief: ""` then reports `{ source: "code", content: "" }`, which is what the runtime is actually doing, and stays inside the wire contract (`content` is null iff `source` is `"unset"`). Worth a seam test pinning empty-profile-brief-with-a-file-present to `source: "code"`.

Greptile's proposed direction was right; the correction is that it must not be applied at the shared `selectConfigSurface` seam, or `design-rules.md` regresses in the opposite direction.

PR #1355, opened 2026-08-16
feat(core): turn envelope wire ops

yousefh409:
A hosted agent turn currently pays a round trip per store op at each end of the turn. This PR adds the core-package contract for two batched cross-family wire ops that collapse those into one call each: `turn.load` bundles a turn's opening reads (`transcripts.getThread` + `workspace.index` + `workspace.read`, plus the optional `harness.get` and the `usage.count` a limits policy needs), and `turn.commit` bundles its closing writes (`transcripts.appendMessages`, plus the optional `harness.set` and the run's audit `engine.put`). Both bodies are composition of the per-op schemas that already existed, and every answer is exactly what the individual op returns — a part the caller does not ask for is absent from the request and absent from the answer, so asking for less costs less. The family is OPTIONAL on `StoreOps` (`usage`'s rule) and its two paths are APPENDED to the tail of `STORE_WIRE_PATHS`, so `/status`'s monotone level keeps meaning what it meant; clients feature-detect with `>=` against the frozen `STORE_WIRE_TURN_OPS = 50`, the way the batch append is detected. The in-memory conformance double implements both as a fan-out over the ops they bundle, and the new conformance cases pin the one contract that matters: each part equals what its own op answers, and each part lands where its own op lands it.

Two existing test files move with the manifest, stated rather than smuggled: the wire inventory count and tail assertions (48 -> 50 paths), and the conformance kit's level-35 mutant mount, which now omits the turn family too — `turn.commit` IS a batch append, so a mount stopping short of op 36 serves neither.

**Known red, and owned by PR 2:** `packages/store/tests/ops.test.ts` ("status() reports the 48 ops this wire serves") fails on this branch alone, because the local backend hardcodes `ops: 48` in `packages/store/src/ops.ts` while that test compares the reported level against `STORE_WIRE_PATHS.length`. The level must not run ahead of the ops, so that number rises only when the local backend actually serves the envelopes — which is PR 2.

This is PR 1 of a 3-PR stack (core contract -> store implementation -> adoption) and merges only at the tip.

---
## Summary by cubic
Batches an agent turn’s store calls into two optional wire ops to cut round trips: previously each turn opened and closed with multiple calls; now `turn.load` and `turn.commit` make one call each without changing semantics. `turn.load.read` is optional so a turn can open with just the thread and index.

- Adds optional `turn` family to `StoreOps`; appends `turn.load`/`turn.commit` to the tail of `STORE_WIRE_PATHS` and freezes `STORE_WIRE_TURN_OPS = 50` for `/status.ops >= 50` feature detection.
- `turn.load` composes `transcripts.getThread` + `workspace.index`, and optionally `workspace.read`, `harness.get`, `usage.count`. Omitted request parts are absent from the response.
- `turn.commit` composes `transcripts.appendMessages`, optionally `harness.set`, and optionally audit `engine.put`. Answers `{messages: {revision, count}}` plus optional `audit`.
- In-memory conformance implements both as fan-outs; new cases pin part-for-part parity and assert `/status.ops >= STORE_WIRE_TURN_OPS` iff `ops.turn` is served.
- Updates wire path inventory and tail assertions from 48 to 50 and keeps the level‑35 mutant mount honest by omitting `turn` alongside the batch append.

- No breaking changes. Clients may batch turns when `/status.ops >= 50`; otherwise call individual ops. Mounts must only bump `/status.ops` to 50 when they actually serve `/turn/*`.

<sup>Written for commit cb0f86e5d35ec5af79c0d14f02fb046a78817a85. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1355?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/turn-envelope-wire-ops.md
+++ b/.changeset/turn-envelope-wire-ops.md
@@ -1,0 +1,5 @@
+---
+"@vendoai/core": minor
+---
+
+Two batched store-wire ops so one agent turn costs one call at each end instead of many. `turn.load` bundles a turn's opening reads — `transcripts.getThread`, `workspace.index`, `workspace.read`, and optionally `harness.get` and the `usage.count` a limits policy needs — and `turn.commit` bundles its closing writes: `transcripts.appendMessages`, optionally `harness.set`, and optionally the run's audit `engine.put`. Both bodies are pure composition of the per-op schemas that already existed, and every answer is exactly what the individual op returns, so nothing here is new semantics; the only thing that changes is the number of round trips. The family is OPTIONAL on `StoreOps` (`usage`'s rule) and rides the tail of `STORE_WIRE_PATHS`, so a mount that omits it reports the level it always did and a caller that finds it absent makes the calls it always made. Clients feature-detect on `/status` with the frozen `STORE_WIRE_TURN_OPS = 50`, read with `>=`, exactly as the batch append is detected.
--- a/packages/core/src/conformance/memory-store-ops.ts
+++ b/packages/core/src/conformance/memory-store-ops.ts
@@ -944,6 +944,32 @@ export function memoryStoreOps(): StoreOps {
   };
 
   // ---------------------------------------------------------------------------
+  // turn — a fan-out over the ops it bundles and nothing more. Written as calls
+  // through the members above so the envelope cannot drift from the individual
+  // ops: a rule added to one is a rule the envelope already obeys.
+  // ---------------------------------------------------------------------------
+
+  const turn: NonNullable<StoreOps["turn"]> = {
+    async load(request) {
+      return {
+        thread: await transcripts.getThread(request.thread.id),
+        index: await workspace.index(request.index),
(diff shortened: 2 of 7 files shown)


yousefh409:
Landed via the stack tip #1364, which squash-merged as aa1c8db3ba94a12bbc6b6d746180ecac495d5380 and carried this branch's commits. `STORE_WIRE_TURN_OPS = 50` and the `turn.load`/`turn.commit` wire ops are on `main` in `packages/core/src/store-wire.ts`. Closing as already-merged — the branch is left in place.

PR #1358, opened 2026-08-16
feat(store): turn envelope implementations

yousefh409:
**PR 2 of a 3-PR stack** — base is `feat/turn-envelope-core` (#1355), not `main`. The stack merges only at the tip.

## What

The `@vendoai/store` half of the turn envelopes: the local backend serves both ops, and the hosted client sends them.

- **`turn.load`** — a fan-out over the very ops it bundles (`transcripts.getThread`, `workspace.index`, `workspace.read`, plus `harness.get`/`usage.count` only when the request asks for them), through a shared `turnLoadOverOps`. Shared rather than copied because the hosted client's fallback is the same fan-out over its own ops.
- **`turn.commit`** — a MUTATION envelope, so its three writes (the batch append, the harness slot, the run's audit row) share **ONE `db.transaction()`**, following `workspace.commit`'s template minus the ceremony it does not need: the wire body carries no idempotency key, so there is no ledger row to claim and no replay to detect, and there is no content-staging leg to compensate. What is left is one transaction that takes the whole commit back on any failure. The thread row is taken FIRST — the order every transcript writer here takes its locks in — and the audit leg goes through the same allowlist gate `engine.put` does.
- **`ops: 48` → `50`.** The level moves WITH the object and never ahead of it: it rises here because this is where the backend genuinely starts serving both ops. This is also what fixes `packages/store/tests/ops.test.ts`, which fails on the base branch precisely because the manifest reached 50 and the backend had not.
- **`appendBatch`** now holds the batch append's body, shared by `transcripts.appendMessages` and the envelope so the two cannot key or position a message differently. The row-id rule they both used is one module-level `rowIdOf`; `putMessage`'s third copy of it is deleted.

## Feature detection

Lives in the hosted client, copied verbatim from `servesAppendMessages` (`src/helpers/thread-messages.ts`): **ONE cached `/status` read BEFORE the first send**, compared against `STORE_WIRE_TURN_OPS` with `>=`. It routes rather than reports — a mount below the level is served by the individual calls the caller always made, which is the cheaper fallback `STORE_WIRE_TURN_OPS` exists to reach. Never a `catch` around a failed mutation (#1251), and a failed handshake is not cached.

## Why

One agent turn currently costs many HTTPS round trips. These two calls collapse the opening reads and the closing writes into one each. No new semantics — every part is exactly what its own op returns.

## Verification

- `pnpm build`, `pnpm typecheck`, `pnpm lint` — all exit 0.
- `tests/ops.test.ts` — 12 passed (was failing on the base branch).
- `tests/ops.conformance.test.ts` — **97 passed**, including PR 1's two `turn.*` conformance cases run against the real Postgres-backed local backend, no stub on either side.
- `tests/hosted-store.test.ts` (69), `thread-messages.batch`, `thread-messages.ops`, `message-loss`, `hosted-store.atomic-parity` — all pass, covering the append path this PR refactored.

**One test file moves with the surface, stated out loud:** the hosted client's family inventory in `tests/hosted-store.test.ts` gains `turn` (eleven families → twelve). Nothing else in a test file changed.

---
## Summary by cubic
Serves `turn.load` and `turn.commit` in `@vendoai/store` to collapse a turn’s opening reads and closing writes into one call each. `/status` now reports `ops: 50` (was `48`); clients feature‑detect and route to individual ops where unavailable.

- `turn.load` fans out via `turnLoadOverOps` over `transcripts.getThread`, `workspace.index`, `workspace.read`, and optional `harness.get`/`usage.count`. Adds prefetch seams: `workspaceStore.open({ index })` and `harnessStateStore.resume(...)`. `workspaceIndexPage` converts the index page and returns `undefined` when a cursor remains. Exports `turnLoadOverOps`, `workspaceIndexPage`, `harnessStateRow`, and `harnessStateKey`.
- `turn.commit` lands the batch append, harness state, and audit in one `db.transaction()`. Locks the thread first; audits go through the same allowlist as `engine.put`. Unifies message keying via shared `rowIdOf` and `appendBatch`.
- Hosted client caches ONE `/status` read at the wire client and shares it across capability checks; routes to individual ops when `STORE_WIRE_TURN_OPS` is not met. Normalizes `usage` bounds to ISO datetimes. Exposes `turn` on the ops surface.
- `@vendoai/guard`: adds `guard.reportThrough(event, place)` and optional `VendoGuard.reportThrough?`; `report` now calls through it so normalization and metrics are unchanged.
- Exports mirrored in `./postgres` entry to match the main entry: `workspaceIndexPage`, `turnLoadOverOps`, `harnessStateKey`, `harnessStateRow`.
- Tests: adds `turn-seams.test` for prefetch seams and pins a single `/status` handshake; updates hosted client family inventory to include `turn`. No migration required.

<sup>Written for commit 93334bb865dd3112ba51fa0768872f227b3f5be7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1358?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Landed via the stack tip #1364, which squash-merged as aa1c8db3ba94a12bbc6b6d746180ecac495d5380 and carried this branch's commits. The store-side envelope implementation is on `main` (`packages/store/src/helpers/turn.ts`, `hosted-store.ts`, `ops.ts`, `postgres.ts`). Closing as already-merged — the branch is left in place.

Issue #1373, opened 2026-08-16
Approval honesty: asks that never rendered (0.20 field), and timeouts narrated as declines (main)


amrib247:
﻿Two related approval-honesty issues from field-testing an ungraded catalog (structural-only judgment, so every tool call asks):

**1. (Undiagnosed, 0.20) The asks never rendered.** Three present-mode tool calls each raised an approval ask; no card or control ever appeared in the thread; each timed out at APPROVAL_WAIT_MS and the turn dead-ended. The ask-render pipeline was fixed before v0.20.0 and is test-pinned on main — so this field repro is not explained by known history. Filing for visibility; we will attempt a re-repro on current npm and attach the transcript either way.

**2. (Confirmed on main) A timed-out ask narrates as "you declined it".** The APPROVAL_WAIT_MS settle rides the same `tool-output-denied` path as an explicit no, so the thread blames the user for a question they never saw, and the persisted transcript carries no decline-vs-timeout distinction. #1324 fixed the adjacent blocked-outcome copy ("wasn't allowed") but not the timeout path. Proposed: a typed timeout outcome, "the approval request expired unanswered" in the beat, and the distinction persisted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

amrib247:
Field re-test on published 0.26.0 (expense.fyi checkout, 2026-08-17), both halves:

**1. The never-rendered asks (0.20 field case): does NOT reproduce — closing this half as fixed-by-unknown.** Same host, same account, same class of prompt ("Add an expense: 120 for coffee today"). On 0.26.0 the pipeline renders completely: the beat appears ("Expenses add create — waiting for your approval", risk=ungraded), and the full approval card renders with argument details (Description/Amount/Category/Date), the ungraded-judgment honesty line ("This hasn't been checked, so we can't say what it changes — it runs as you."), the grant-remember controls (scope/duration), and a bottom-sheet presentation at narrow widths. Whatever broke rendering on 0.20 was fixed somewhere in 0.21-0.26 without being tracked to this issue. Versions: field repro @0.20.0, re-test @0.26.0.

**2. Timeout-as-decline: CONFIRMED live on 0.26.0, verbatim.** Letting that same ask expire at APPROVAL_WAIT_MS (90s, nobody clicked anything): the sheet dismissed itself and the beat settled as

> Expenses add create — you declined it

[...]

Issue #1371, opened 2026-08-16
doctor's live checks can never run on a preset-authed host: /status 403s the unauthenticated probe


amrib247:
﻿On any preset-authed host (authJs/clerk/supabase/auth0), `vendo doctor`'s live checks are structurally dead: the unauthenticated GET /status probe hits the identity wall (presets return null for a sessionless request -> 403), E-LIVE-001 reports the failure, and the present/actAs/model-turn checks chain into "start the dev server and retry" — a false cause, the server is up.

/status resolves a principal solely for the per-caller `memberships` echo; everything else in its body is a deployment fact, and zero-config dev hosts already answer it unauthenticated. Proposed direction: make /status identity-free (memberships echo only when a principal resolves; null -> omit, not 403) — restores most of doctor's live surface in one change and stays inside the resolver-or-nothing doctrine — plus 403-aware copy for the remaining probes ("the wire is up; the present probe needs a signed-in session") instead of "start the dev server". A doctor dev-credential is the wrong fix: it contradicts the doctrine.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

amrib247:
Resolved upstream by #1385 (merged 2026-08-17), which chose the opposite door from the direction proposed here: instead of making /status identity-free so the probes could run, doctor's entire live surface was deleted - it now grades what is on disk, starts no server, makes no HTTP request. The false "start the dev server" chain this issue describes cannot happen anymore because the probes that printed it no longer exist; the probe-only error codes are retired append-only. The upstream PR cites the same field pain on the Auth.js preset that this issue measured on supabase().

Residual question this issue leaves behind, parked deliberately: the wire's /status still 403s a sessionless request on preset hosts (the per-caller memberships echo forces principal resolution), while zero-config dev hosts answer it unauthenticated. With doctor no longer a consumer, the case for an identity-free /status rests only on manual curl verification and external monitoring - a thinner motivation that should be argued fresh if someone hits it, not ride this issue's now-dead premise. Closing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1370, opened 2026-08-16
init's ONE-STEP paste mounts nothing visible, and supabase() wires without naming its server-side env


amrib247:
﻿Two init gaps that compound on a fresh host, found field-installing into a real product:

**1. The "ONE STEP LEFT" paste mounts nothing visible.** The frame prints only `<VendoProvider>` — deliberate per the source comment ("No overlay line") — but the frame's own copy ("Until this lands, Vendo is wired but nothing on the page can reach it") implies the paste completes the install, and `vendo doctor` HARD-FAILS exactly this state as E-WIRE-006 ("no visible agent surface is mounted"). A verbatim install renders nothing and reads as broken. The paste should include `<VendoOverlay />` (annotated: remove if you render your own surface).

**2. `auth: supabase()` wires silently without naming its server-side env.** Init detects `@supabase/*` and wires the preset; a typical host carries only the `NEXT_PUBLIC_*` pair, while the preset verifies sessions with server-side `SUPABASE_JWT_SECRET` and/or `SUPABASE_URL`. Nothing advises this at init or doctor time — the first signed-in turn fails (with good wire copy, but a turn too late). The `AuthMatch.advisory` mechanism already exists (next-auth v4 uses it); supabase should attach one when neither env name is present, and doctor deserves a static check.

PR incoming for both.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

yousefh409:
Fixed by #1374 — init now appends <VendoOverlay /> in the paste step, and the Supabase preset ships a server-env advisory plus doctor code E-AUTH-009.

PR #1374, opened 2026-08-16
host2 fix pack lanes 1-2: Next 14 wire compile, a visible ONE-STEP paste, supabase() env named early (#1369, #1370)

### LINKED ISSUE #1369: Next 14 hosts cannot compile the wire: webpack follows the dynamic undici import in keep-alive-fetch
﻿Installing published @vendoai/vendo into a stock Next 14.1 App Router host: the wire route fails to COMPILE — webpack statically follows the dynamic `import("undici")` in `keep-alive-fetch.js` (via server.js -> connections/sandbox/compose-store/cloud-tools), and Next 14's parser cannot read undici 7's modern syntax (`#target in this`). Every request 500s with a raw module-parse error naming neither Vendo nor a fix.

The import is a perf-only keep-alive Agent pool that already degrades to plain fetch at runtime (`options.fetch ?? keepAliveFetch` seams + `.catch` fallback) — the wall is pure bundler visibility. Verified byte-identical on current main; `undici ^7.29.0` is a hard dependency; plain-tsc dist ships the dynamic import verbatim.

Escape hatch today (undocumented): Next 14's `experimental.serverComponentsExternalPackages: ["undici"]` — note the docs only ever show `serverExternalPackages`, the Next 15+ spelling, and never include undici.

Proposed fix: `/* webpackIgnore: true */` on the dynamic import (tsc passes it through; webpack leaves the import to the runtime; Node resolves undici normally; the existing edge/absent-undici fallback is untouched), plus the Next 14 spelling documented wherever serverExternalPackages appears. PR incoming.

amrib247:
## What

Lanes 1–2 of the host-#2 fix pack (plan: `docs/superpowers/plans/2026-08-16-host2-fix-pack.md`; field source: a stock install into expense.fyi, a Next 14.1 + Supabase host; every finding verified still-present on current main before fixing). Fixes #1369 and #1370; Linear ENG-420/421/422.

Three commits:

1. **`fix(vendo): bundlers never follow the keep-alive pool undici import`** — Next 14 hosts could not compile the wire: webpack statically resolves the pool's dynamic `import("undici")` and Next 14's parser cannot read undici 7 syntax, so every wire-route build died with a raw module-parse error. The pool is perf-only with an existing plain-fetch fallback; a `webpackIgnore` comment leaves the import to the runtime. Test pins the comment; docs gain the Next 14 spelling (`experimental.serverComponentsExternalPackages`) beside every `serverExternalPackages` mention. **Field-verified**: the same host compiles and boots (`[vendo] ready`, 403 identity wall) with `undici` removed from its externals list.
2. **`feat(vendo): the ONE-STEP paste yields a visible agent`** — the frame printed only `<VendoProvider>`, which renders nothing; a verbatim install completed invisible while doctor E-WIRE-006 hard-failed exactly that state. The paste now includes `<VendoOverlay />` (annotated: delete if you render your own surface). This reverses a deliberate design pin ("No overlay line") — the reversal argument and field evidence are recorded in the mountStep comment and the updated test.
3. **`feat(vendo): supabase() names its server env before the first turn fails`** — init detects the Supabase family from dependencies (often only the `NEXT_PUBLIC_*` pair) while the preset verifies sessions with server-side `SUPABASE_JWT_SECRET`/`SUPABASE_URL`; the first signed-in turn failed after every static check passed. Detection attaches an advisory (same mechanism as the next-auth-v4 one) when neither name is in the process env or host env files; doctor grades the same fact statically as **E-AUTH-009** (warning) via one shared helper so init and doctor cannot disagree; troubleshooting page carries the code (registry parity test extended).

## Tests

- New/updated: keep-alive bundler pin, 5 supabase-advisory tests, 3 E-AUTH-009 doctor tests, codes-registry snapshot + docs parity, init paste pins updated to the new contract (the old "no overlay line" pin reversed deliberately).
- Affected scopes green locally: keep-alive 2/2, init-auth 13/13, doctor-codes + docs parity 5/5, init 106/107 and doctor 107/112 — every red reproduces identically on pristine main (the known Windows env class: POSIX paths, spawn-ENOENT, live-server probes); CI is the gate of record.
- Full package-graph build green (0 TS errors).

## Not in this PR

Lanes 3–5 (doctor-vs-identity-wall /status design, signed-out chrome state, approval timeout honesty + the undiagnosed ask-render investigation) are scoped in the plan and filed as #1371–#1373 / ENG-423–426 — each warrants its own review conversation.


amrib247:
**Greptile triage:** the P1 (E-AUTH-009 skipped outside Next) is real and is fixed in a1b37f133. The check gated discovery on the Next catch-all route, so Express/custom hosts wiring `supabase()` never got graded — while init's dependency-based advisory *does* fire there, which broke this PR's own never-disagree contract by omission. Discovery is now framework-neutral: with a Next route the composition file is read as before (a bare `supabase()` call is trusted there); anywhere else the same bounded source scan `detectVendoWiring` uses looks for the preset **import** only — a bare call in arbitrary host source is the host's own Supabase client, not the preset. Covered by four new doctor tests (Express warn/pass, custom-runtime warn, commented-out import stays silent).

amrib247:
**Greptile triage, round two:** the alias P1 is real and fixed in d5ccc45b9 — the framework-neutral scan only knew the scoped specifier, so a host wiring the preset through the unscoped `vendoai` alias re-export (a fully supported spelling; the alias package exports `./auth/supabase`) missed E-AUTH-009. One shared regex (`SUPABASE_PRESET_IMPORT`) now names both supported spellings, and both discovery branches (Next composition read + generic scan) use it. New test covers the alias-wired Express host. The regex spelling — rather than string literals — is deliberate, so the dependency guard never reads an import-shaped alias specifier out of `packages/vendo`.

Issue #1340, opened 2026-08-16
Deterministic theme extraction only speaks shadcn's vocabulary - a Tailwind v4 @theme palette extracts 1 slot of 12


YSharara1:
## Summary

Deterministic theme extraction reads shadcn's token vocabulary and only shadcn's. A Tailwind v4 `@theme` block named the Tailwind way extracts almost nothing, silently, and the palette falls through to the consent-gated model pass — so a host with no model key gets the neutral blue default from a design-token sheet that declares every color it needs.

A Next 16 app whose entire palette lives in one `@theme` block produced this:

```json
// .vendo/theme.extracted.json
{ "format": "vendo/theme-extracted@1", "slots": { "border": "#ece9e3" } }
```

**One slot out of twelve** — and `border` only matched by accident.

## The obvious diagnosis is wrong

My first conclusion was "the extractor doesn't parse `@theme`". It does, and stating this saves the next person the same wrong turn:

- `parseCssVars` (`packages/vendo/src/cli/theme/css-vars.ts`) is a brace-depth scanner. It is at-rule agnostic and reads declarations inside `@theme` exactly as it reads `:root`.
- `lastLightDecl` (`extract-theme.ts:282`) already tries every allowlisted name in its Tailwind-v4 namespaced spelling: `--border` → also `--color-border`.

That second point is precisely *why* `border` came through. The `@theme` block is read correctly. **It is the vocabulary that misses.**

## The actual gap

`EXACT_COLOR_TOKENS` (`extract-theme.ts:259`):

| Vendo slot | Accepted names |
| --- | --- |
| background | `--background` / `--color-background` |
| text | `--foreground` / `--color-foreground` |
| surface | `--card` / `--color-card` |
| accent | `--primary` / `--color-primary` |
| mutedText | `--muted-foreground` / `--color-muted-foreground` |
| border | `--border` / `--color-border` |
| danger | `--destructive` / `--color-destructive` |

The app names its tokens the way Tailwind's own docs name them: `--color-bg`, `--color-ink`, `--color-surface`, `--color-accent`, `--color-muted`, `--color-neg`. Nothing matches.

The sharpest edge: **Vendo's own slot names are not accepted as token names.** A host that writes `--color-surface` and `--color-accent` — the two slots Vendo literally calls `surface` and `accent` — gets neither, because the allowlist maps those slots to shadcn's `--card` and `--primary`. Naming your tokens after the thing you are configuring is the most natural guess available, and it is the one that fails.

[...]

Issue #1339, opened 2026-08-16
openapi.json is the highest-leverage file a host can add, and it is documented nowhere (0/18 -> 18/18 tool schemas)


YSharara1:
## Summary

An `openapi.json` at the app root is the single highest-leverage file a Vendo host can add, and it is documented nowhere.

`vendo init` on a Next 16 app extracted all 18 REST routes correctly, then reported:

```
tool schemas: inputs 0/18 · outputs 0/18 — blind: host_customers_get,
host_customers_list, host_customers_notes, host_demo_reset_create … +12 more
```

Every tool arrived as a method and a path with no parameters and no output shape. `GET /api/orders` supports `status`, `customer_id`, `q` and `limit`; the agent had no way to know any of them existed.

Adding an `openapi.json` at the root and re-running init changed exactly one thing:

```
tool schemas: inputs 18/18 · outputs 18/18
```

Same routes, same command, same everything else. The bindings flip from `kind: "route"` / `inputSchemaSource: "unknown"` to `kind: "openapi"` / `"declared"`, and each tool gains real parameter names, types and descriptions.

**That is the difference between a tool list and an agent that can use it, and nothing tells you the file exists.**

## How it is discovered

`packages/actions/src/sync/extractors.ts` looks for it by convention in six locations:

```ts
"openapi.json",
"openapi.yaml",
"openapi.yml",
path.join("public", "openapi.json"),
path.join("docs", "openapi.json"),
path.join("docs", "openapi.yaml"),
```

Grepping `docs/`, the root README and `packages/vendo/README.md` for "openapi" returns nothing outside a knowledge-eval fixture (`docs/eval/knowledge/golden.json`). The only way to learn this is to notice that `examples/demo-bank`'s tools are richer than yours and go read its directory to find out why.

The init output already knows there is a problem — it prints the `blind: …` list — but treats it as a statistic rather than something with a remedy.

## Related: two dead extension keys

`examples/demo-bank/openapi.json` uses `x-vendo-dangerous` and `x-vendo-formats`. Neither string appears anywhere under `packages/`:

```
$ grep -rn "x-vendo-dangerous\|x-vendo-formats" --include=*.ts --include=*.tsx packages/
(no matches)
```

Nothing reads them. Since demo-bank's spec is the only available template, anyone copying it will reasonably assume risk can be declared inline in the spec, and be quietly wrong. Either wire them up or drop them from the example.

[...]

Issue #1338, opened 2026-08-16
clerk() throws when CLERK_SECRET_KEY is unset, 501-ing the entire wire - the state vendo init --auth clerk leaves you in


YSharara1:
## Summary

`clerk()` **throws** when neither `CLERK_JWT_KEY` nor `CLERK_SECRET_KEY` is set and the request carries any credential — and the throw takes down the whole wire, not just identity. Every vendo route answers `501 Internal Vendo error`.

`packages/vendo/src/auth-presets/clerk.ts`, `sessionClaims`:

```ts
const token = bearerToken(request) ?? cookieValue(request, "__session");
if (token === undefined) return null;
const jwtKey = environment("CLERK_JWT_KEY");
const secretKey = environment("CLERK_SECRET_KEY");
if (jwtKey === undefined && secretKey === undefined) {
  throw new Error(MISSING_KEY_MESSAGE);          // ← line 67
}
const verifyToken = await loadVerifyToken();
try {
  return await verifyToken(token, { /* … */ });
} catch {
  return null; // unverifiable/expired token = no session, mirroring authJs
}
```

Nine lines apart, the same function answers two versions of "I cannot verify this token" in opposite ways. A **forged or expired** token — strictly the more suspicious input — resolves to anonymous and the app keeps working. A **missing local key** throws.

## Why it lands on the first run

`vendo init --framework next --auth clerk` scaffolds a composition with `auth: clerk({...})` and, correctly, writes no secrets. So between finishing init and pasting keys, the host is in exactly the throwing state — and that is the window in which a newcomer runs `vendo doctor`.

Doctor's own probe sends an `Authorization: Bearer …` header, so it trips the throw itself:

```
auth/present   ✗  Internal Vendo error
```

The reported check is `auth/present`, the message names no cause, and the advice that follows points elsewhere. Nothing connects it to `CLERK_SECRET_KEY` — the one thing that would fix it. It reads as "Vendo is broken", not "you have not pasted your keys yet".

The blast radius is the surprising part: with no key set, a keyless `GET` resolves anonymous and works, so the app looks fine until the first authenticated request 501s the entire wire.

## Affected

`packages/vendo/src/auth-presets/clerk.ts` at main (v0.8.0). Any host that runs `vendo init --auth clerk` and does anything before pasting keys — including running the doctor the init output tells them to run.

Not the same as #1256: that is about auth Vendo *cannot read* answering 403. This is a preset throwing on its own missing configuration and surfacing as a 501.

[...]

PR #1397, opened 2026-08-17
feat(examples): crate, an order-ops console host app on the clerk preset

YSharara1:
## What

Adds `examples/crate` — a self-contained order-ops console for a small online store: orders, customers, inventory and shipments, wired to Vendo end to end.

- **16 REST route files → 18 extracted tools**, 9 screens, ~2.4k lines of domain code that owes nothing to Next.
- **A deterministic seed with a story.** ~85 orders across 90 days, 12 customers, 10 products, ~75 shipments, and staged on top: *a customer charged twice for the same espresso machine, three minutes apart. One copy was delivered; the other never shipped.* Refunding the second charge is the thing the app exists to make easy, and the thing an agent should be able to do end to end.
- **63 tests** over the seed's invariants, every domain rule, and the staff roster.
- Runs with **no environment file at all** — a fresh clone shows the shop, not a stack trace about a missing key.

Two root files change: `pnpm-workspace.yaml` gains `'@clerk/shared': false` (its postinstall only prints Clerk's telemetry notice and builds nothing the SDK needs), and `pnpm-lock.yaml` is additive.

## Why

Every existing example is Next.js and the only one with auth uses `authJs`, so `clerk`, `supabase`, `auth0` and `jwt` had **zero coverage**. Crate takes the `clerk` preset on the best-tested framework — an untested auth path without betting the example on an untested framework.

It also exercises three things nothing in `examples/` did before:

**`openapi.json` is the highest-leverage file a host can add.** Without a spec, `vendo init` extracts the same 18 routes but every one arrives *blind* — no inputs, no output shape, nothing but a method and a path. `vendo sync` reports the difference directly:

```
tool schemas: inputs 0/18 · outputs 0/18   ← without openapi.json
tool schemas: inputs 18/18 · outputs 18/18 ← with it
```

**Risk graded by hand, with no model key.** `.vendo/overrides.json` grades all 18: reads run, writes ask, refunds and cancellations are `destructive`, and `resetDemo` is disabled outright — an agent helping a customer has no business discarding the shop's data.

**The identity seam, said out loud.** Clerk answers "who signed in"; Crate's roster answers "and are they staff here?" The join is on email, because a Clerk subject means nothing to Crate and Clerk's default claims carry no email. `clerkEnabled` gates the provider, the proxy and the session read, so the whole app runs keyless as the seeded owner.

## Findings this produced

Building it surfaced four issues, all filed separately:

- **#1338** — `clerk()` throws on any `Authorization: Bearer …` header when `CLERK_SECRET_KEY` is unset, instead of returning null the way it does for an unverifiable token six lines later. That is exactly the state `vendo init --auth clerk` leaves you in, and doctor's own probe sends a Bearer header — so the first `vendo doctor` after init fails with "Internal Vendo error" and the wrong advice.
- **#1339** — `openapi.json` is undocumented. The extractor looks for it in six conventional locations and it appears in no doc; the only way to find it is to read `demo-bank`.
- **#1340** — theme extraction's deterministic pass reads *shadcn's* token vocabulary. A fully-declared Tailwind v4 `@theme` sheet named the Tailwind way matches almost nothing and falls through to the model pass, so without a model key it lands on the neutral default. The `@theme` block parses fine; it is the names that miss.
- **#1253** (existing, commented) — `vendo init` does not add `serverExternalPackages`. Adds the Turbopack-production half: `next build` + `next start` 501s every wire request with "f.instantiateWasm is not a function" while `next dev` is fine.

Crate carries the workarounds for #1338 and #1253 in-tree, and `README.md` documents all four under "Known rough edges" so the next person reading the example is not surprised by them.

## Verification

`vendo doctor` against a running dev server: **24 of 25 ok, 0 failures, exit 0**, including a live `turn/model` over a Vendo Cloud key. The one warning is `auth/act-as`, which needs a configured Clerk app; it is a warning, not a failure, and the example is documented as running keyless.

`examples/crate`: **63 tests pass**, typecheck clean.

- [ ] `pnpm build && pnpm test && pnpm typecheck && pnpm lint` green

That box is deliberately unchecked: the root gate does not run on Windows, which is where this was built. `pnpm test` uses POSIX inline env-var syntax (`VITEST_MIN_FORKS=1 …`) that cmd.exe rejects outright, and `scripts/dependency-guard.mjs` fails to parse `pnpm-workspace.yaml` when `core.autocrlf` gives it CRLF line endings. The committed blob is LF, so neither affects CI — but it does mean I have not personally seen the full stack green, and I would rather say so than tick a box on CI's behalf.


YSharara1:
Superseded by [runvendo/vendo-cloud#79](https://github.com/runvendo/vendo-cloud/pull/79).

This repo is now a one-way copybara projection of `runvendo/vendo-cloud` — its `main` is written only by `vendo-release-bot`, so a PR here can no longer land. `examples/**` is projected, so crate will appear here automatically once #79 merges.

The app is unchanged from what was approved here. The port did need real work: crate was built against 0.8.0 and vendo-cloud is on 0.61.1, with `@vendoai/core` and `@vendoai/ui` folded into the umbrella since — so the imports, `package.json` and `next.config.ts` all had to be rewired. Details and verification are on #79.

Thanks to the three reviewers here; leaving this open rather than closing it in case anyone wants the history, but the live one is #79.

PR #1417, opened 2026-08-17
fix(ui): a column that fits is a column, whatever the pixel rounds to

yousefh409:
## The bug

At some viewport widths `DataTable` folded its last column into the first cell although there was room for it — the column silently stopped being a column. Pre-existing; found during the mapped-rows build and reproduced in a real browser. Both the `columns` API and the mapped-rows children API are affected: `TableRow` does no measuring of its own, it reads the same `visibleCount` off `TableContext`.

## Root cause

The fold compared a sum of ROUNDED widths against a ROUNDED room. `offsetWidth` and `clientWidth` are whole pixels; a laid-out column is fractional.

Three columns filling a 1000px table measure `320.8125 + 387.546875 + 291.640625` — exactly 1000 — but each `offsetWidth` rounds up, so the cumulative edge reads `321 + 388 + 292 = 1001`: one pixel past a room it exactly filled. The error grows with the column count (up to half a pixel each).

The room rounds too. A scroller laid out on a fraction reports a `clientWidth` short of what it has, and the column filling that fraction folds as well — so fixing only the column side would have made the fractional-container case *worse* (41 bad widths vs 17 in the sweep below).

Both readings now carry the fraction their laid-out box really has, so the comparison is fractional on both sides. No tolerance, no epsilon. The room is the scroller's CONTENT box, with the border widths read off the element rather than assumed integral — a fractional themed border handed back as room would keep a column that overflows, the same failure mirrored (found in review, fixed in 1d90311fd, with a test). Nothing is laid out under SSR and jsdom, where the whole-pixel reading stands exactly as it was. The fold design itself is untouched.

## Proof

**Browser sweep** — Chromium, 1000→1200px in 5px steps across four container shapes (flush `100%` plus three fractional widths), asserting the last column is present whenever the true fractional widths fit the scroller's true fractional content box:

| | result |
|---|---|
| before | **FAIL** — 31 of 164 widths folded a fitting column (incl. 1100px) |
| after | **PASS** — 0 |

Screenshots at 1100px: `/tmp/fold-fix/before-1100.png` (CLIENT + AMOUNT only, "Due: Mar 14" folded into the first cell, half the table empty) and `/tmp/fold-fix/after-1100.png` (CLIENT, AMOUNT, DUE as real columns).

**Red→green** — two new tests carry Chrome's own measurements (one for each rounding) and fail on the old code with exactly the reported symptom, 2 columns of 3; they pass on the new. No existing test was modified — the 22 that were already there still pass untouched.

**Gates** — ui suite 1320/1320, `pnpm typecheck`, `pnpm lint`, `pnpm test:affected` (34/35 tasks; the one red was 3 unrelated 30s timeouts in `packages/vendo` under a load average of 56 from a concurrent worktree run — `tests/harness-wire.test.ts` 19/19 and `tests/server.test.ts` 126/126 both green when re-run scoped).

---
## Summary by cubic
Fixes `DataTable` column folding by comparing fractional widths and measuring the room as the content box. Previously we summed rounded column widths against a rounded room from the scroller’s border box, which could fold a column that fits or keep one that overflows when borders were fractional; now we reconstruct fractions from each element’s rect and subtract borders from the room.

- Uses `unrounded(...)` for header edge sums and `unrounded(node.clientWidth, contentWidth(node))` for the scroller room, where `contentWidth` reads the rect and subtracts computed border widths.
- Keeps the existing fold design and “first column always stays” rule; no API changes to `DataTable`, `TableRow`, or `TableContext`.
- Leaves SSR/jsdom unchanged (rects read as zero, whole-pixel behavior persists).
- Adds three tests: fit under subpixel rounding, fit against fractional content width, and no “spending” of fractional borders as room; existing tests remain unchanged.

<sup>Written for commit 1d90311fde7f391d74c6ef2a64e37821334fcb90. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1417?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/ui/src/kit/data/data-table.tsx
+++ b/packages/ui/src/kit/data/data-table.tsx
@@ -133,6 +133,16 @@ export const foldStyle: CSSProperties = {
   fontVariantNumeric: "normal",
 };
 
+/** A width to the FRACTION. `offsetWidth` and `clientWidth` are whole pixels,
+ *  and the element's own rect carries the fraction they rounded away — summing
+ *  rounded column widths against a rounded room overshoots by up to half a pixel
+ *  per column, which folds a column that fits. A rect is all zeroes where nothing
+ *  is laid out (SSR, jsdom), leaving the whole-pixel reading exactly as it was. */
+const unrounded = (whole: number, el: Element): number => {
+  const { width } = el.getBoundingClientRect();
+  return whole + width - Math.round(width);
+};
+
 export function DataTable(props: DataTableProps) {
   const {
     rows: rawRows,
@@ -311,13 +321,14 @@ export function DataTable(props: DataTableProps) {
       if (headers !== undefined && headers.length === expandedHeaderCount) {
         let edge = 0;
         naturalEdges.current = [...headers].slice(0, columns.length)
-          .map((th) => (edge += (th as HTMLElement).offsetWidth));
+          .map((th) => (edge += unrounded((th as HTMLElement).offsetWidth, th)));
       }
       const edges = naturalEdges.current;
       if (edges.length === 0) return;
       // Keep every column that has room, fold only the ones that do not. The
       // first column always stays, however narrow the surface is.
-      setVisibleCount(Math.max(1, edges.filter((right) => right <= node.clientWidth).length));
+      const room = unrounded(node.clientWidth, node);
+      setVisibleCount(Math.max(1, edges.filter((right) => right <= room).length));
     };
     measure();
     const observer = new ResizeObserver(measure);
--- a/packages/ui/src/kit/data/data-table.tsx
+++ b/packages/ui/src/kit/data/data-table.tsx
@@ -133,14 +133,20 @@ export const foldStyle: CSSProperties = {
   fontVariantNumeric: "normal",
 };
 
-/** A width to the FRACTION. `offsetWidth` and `clientWidth` are whole pixels,
- *  and the element's own rect carries the fraction they rounded away — summing
- *  rounded column widths against a rounded room overshoots by up to half a pixel
- *  per column, which folds a column that fits. A rect is all zeroes where nothing
- *  is laid out (SSR, jsdom), leaving the whole-pixel reading exactly as it was. */
-const unrounded = (whole: number, el: Element): number => {
-  const { width } = el.getBoundingClientRect();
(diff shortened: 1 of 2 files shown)


yousefh409:
@greptile-apps — finding accepted and fixed in 1d90311fd; the suggested remedy was not taken, because it reintroduces the bug this PR exists to fix.

**The finding is real.** `unrounded(node.clientWidth, node)` took the room's fraction off the scroller's rect, which is the *border* box. The border cancels only while the two sides sum to a whole pixel, and `borderWidth` is a host theme's own free-form string (`packages/apps/src/contract/catalog.ts:152`), so nothing in the contract guarantees that. A fractional border was spent as room and a column overflowing by a hair stayed put — the fold's own failure, mirrored. There is now a test for exactly your numbers (`does not spend a fractional border as room`): red before the change at 3 columns, green after at 2.

The room is measured as the content box now, with the border widths read off the element instead of assumed integral.

**Why not `node.clientWidth` directly.** That is the whole-pixel reading whose mismatch with fractional column widths *is* this bug. Measured in Chromium over 41 viewport widths per container shape:

| room measured as | folds a fitting column at |
|---|---|
| sum of rounded `offsetWidth` vs `clientWidth` (before) | 31 widths |
| fractional widths vs `clientWidth` (your suggestion) | **41 of 41** widths, fractional-width container |
| fractional widths vs fractional content box (this PR) | 0 |

A scroller laid out on a fraction reports a `clientWidth` rounded down from what it has, and the column filling that fraction folds — which is how the reported 1100px case behaved.

**One correction for the record.** Your repro is not reachable through a themed border in Chromium: the used border width is snapped to a whole pixel there. `0.09375px`, `0.5px`, `0.75px` and `1.5px` all compute to exactly `1px`, at device pixel ratios 1, 1.5 and 2 — so the border total is always integral and the old identity held exactly. The fix stands anyway: no engine promises that snapping, and the assumption is not worth carrying.

PR #1424, opened 2026-08-17
fix(ui): the pin ghost lands flush, and the ring waits for the pin, not a timer

yousefh409:
**Slice 1 of 8 — the pin ceremony.** Bottom of the "how generated apps appear" stack.

When you pin a generated app, the ghost used to dissolve *while it was still flying*, so it never visibly landed anywhere — and the confirmation ring fired off the animation's own timer, which meant the ring claimed "it's there" even when the write that puts it there had failed or never happened.

Now: the 300ms flight and its spring curve are unchanged, the end-of-flight fade is gone (the ghost lands flush on the slot and is removed the frame the slot's loading state ignites), and the ring is gated on a real confirmation in **every** host configuration:

- host wired `pinSlot` — the ring waits on Vendo's `apps.place` write;
- host wired only `onPin` — the ring waits on the host's own handler, which is why `onPin` is now typed `void | Promise<void>` (both mirrored declarations, `context.tsx:35` and `:126`);
- `playPinCeremony()` called directly with no `confirmed` promise — **no ring at all**. An unresolved promise draws nothing, so a pin somebody silently dropped gets the flight and no claim that it landed.

Failure keeps the existing toast and draws no ring.

### Reviewer notes
- `packages/ui/test/chrome/pin-ceremony.test.tsx` is an authorized existing-test edit: 8 call sites gain `confirmed: Promise.resolve()` and 6 bodies gain one `await Promise.resolve()` microtask flush, because a gated ring now lands one microtask after the landing. **No assertion was deleted, loosened, or rebudgeted** — the flight+ring `toBeLessThan(500)` budget is untouched.

### Base drift — resolved
`main` moved three times during this work. This branch is rebased onto current `origin/main`, and the one real conflict (`packages/ui/src/tree/renderer.tsx`) was resolved by slice 4, which owns that file — see PR 4.

### Live proof — real browser, scripted, with controls
**PASS.** Pin flight and the write-gated ring both proven on demo-bank (Maple) in a headed run with every non-GET Vendo call delayed 1500ms, so the write genuinely lags the animation. Screenshots `pC-*` and the `video-pC` recording show the ghost lifting, flying, and landing flush with no end-of-flight dissolve; `pC-ceremony.json` has the frame-level event trace. `pG-01-place-refused.png` / `pH-01-refused.png` are the negative control: a refused `apps.place` draws **no ring** and shows the honest toast. `pH-02-reduced-pin.png` covers reduced motion.

### Gate
Last full local gate (tip `3b70a7016`): `pnpm build` **PASS** (21/21 tasks); `pnpm test:affected` — 9,500+ tests green across 23 packages, with `@vendoai/vendo` 2,370 passed and `demo-bank` 123 passed. Two known issues at that run, both since addressed or explained:

- `examples/demo-bank/tests/vendo/away-drill.test.ts` — a **known load flake**, not a product failure: under full-suite load its `beforeAll` dev-server boot hook timed out at 240s and the suite's 4 tests were skipped. Run alone it is **4/4 PASS in 24.5s**. Unrelated to this stack (grants/automations/session).
- 3 failures in `packages/vendo/tests/server.test.ts`, bisected to lane 6 and **fixed in lane 6** (the arrival mark was inside the response's success path; it now cannot fail or alter a render).

**This gate predates the rebase onto newer `main`, so it is not the authority — CI is the gate of record.** The stack has since been rebased onto current `origin/main`, which moved three times during this work.

Stack: this PR → `main`. Slices 2-8 stack on top; **PR 8 is the merge unit**, the inner PRs are review windows.

---
## Summary by cubic
Pins land flush and the confirmation ring is outcome‑gated. Previously the ghost faded before touchdown and the ring fired on a 300ms timer; now the ghost is removed on landing, and the ring fires only on a real confirmation: with a `pinSlot` it fires when `client.apps.place` succeeds and does not wait on host mirrors; with only `onPin` it waits on the host’s handler; calling `playPinCeremony` without `confirmed` draws no ring. Slow confirmations re‑resolve the destination; failed pins show the toast and no ring; host mirror rejections are caught and logged.

Apps also arrive where people can see them: `GET /apps` returns `AppListRow[]` with `unseen?`, which drives the launcher dot via the shared apps feed; thread renders mark arrival only for the visible window. Pending opens include forming geometry so embeds paint layout while building. Slots remember a pinned app’s silhouette and no longer claim “empty” until placements resolve. Pinned apps gain a ✦ handle with Edit in chat/Refresh/Unpin; “Edit in chat” scopes the overlay to the picked app. Forming sections paint dimmed (“wet”) and kit empty states hold during builds.

**Migration and review**
- Hosts: make `onPin` return `void | Promise<void>` if you mirror pins; with a `pinSlot` the ring no longer waits on your mirror.
- Callers/tests: pass `confirmed` to `playPinCeremony` when you expect a ring, and await one microtask after landing before asserting it.
- Client types: `VendoClient.apps.list()` and `useApps()` now return `AppListRow` with `unseen?`; adjust usages.
- Overlay: use `openVendoConversation({ appId })` to scope “Edit in chat”; the stage records the pick and stays empty until there’s an embed.
- Tests/slots: the empty‑slot CTA is asynchronous; await it. The pre‑resolution skeleton drops aria‑live.
- Core: `ENGINE_ALLOWLIST_VERSION` is 5; tests expect 38 static collections.

<sup>Written for commit aeb80e88224c2a045622f2382e9a5448f751b5d5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1424?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
### Live-proof artifacts for this slice

Captured by a worker that built none of this, on demo-bank (Maple) in a headed browser. Binaries live on the build machine at `/home/ubuntu/proof-artifacts/appear-stack-8/` — GitHub has no API for attaching image files to a PR, and committing 55 PNGs and 3 recordings into the repo would be bloat, so they are cited by name here rather than embedded.

- `pC-10-after-pin.png`
- `pC-ceremony.json`
- `pC.log`
- `pG-01-place-refused.png`
- `pH-01-refused.png`
- `pH-02-reduced-pin.png`
- `video-pC/`

yousefh409:
**Third variant fixed — and it was ours, introduced by the gating itself.** `a9e01a582`

**The bug:** the confirmation ring was *withheld* when a host's `onPin` was slow, on a pin that genuinely succeeded. Verified against source and reproduced with real components before any change.

**Root cause:** `settle` closed over the destination `Element` resolved at measure time, and only dereferenced it when the confirmation landed. Landing the pin announces it, the real `VendoSlot` re-reads and re-renders, and that **replaces** its element — so a slow confirmation measured a detached node, which measures `0`, so `pulse` returned early and no ring was ever drawn. A fast confirmation won the race, which is why this hid behind the earlier `onPin` test.

**The fix:** re-resolve the destination when the confirmation lands, which is exactly what the two `rAF` hops above it already do for the same reason. No timer, no new option, no timeout — a ring after a slow save is correct; a ring that never comes is the bug.

**Harness note worth flagging:** the rect stub in `pin-ring-write.test.tsx` matched on selector alone, so it sized a node that was no longer on the page and made this bug invisible. It now respects `isConnected`, which is what a real browser does. Without that, the new case cannot fail.

**Coverage:** the ring is now gated *and* delivered in every configuration — Vendo's write, an `onPin`-only host's own awaited mirror, a slow mirror (rings late, never never), and a direct `playPinCeremony` caller that confirms nothing (no ring). Mutation-checked: reverting the fix reds the slow case alone and leaves all 18 existing ceremony tests green.

Gate: `@vendoai/ui` build + typecheck + lint clean, 143 files / 1326 tests passing.

yousefh409:
### For the record: one open, unowned finding this train surfaced

Everything else this train found is either fixed or explicitly declined. This one is neither, and it should not evaporate when these PRs close.

**`packages/ui/test/chrome/slot-report-refusal.test.tsx` fails intermittently, because of a real ordering bug in source.**

`report` in `packages/ui/src/hooks/use-placements.ts` stamps its dedupe key *before* the request goes out:

- `:233` `const at = reported.get(key);`
- `:235` `reported.set(key, Date.now());` — stamped here
- `:258` `void client.slots.report(sending).catch(() => forget(sending));` — the POST, with the key cleared only inside that async rejection handler (`forget` at `:100`)

So a second mount that reads `reported` before the rejection resolves is suppressed for the full `SLOT_REPORT_REFRESH_MS` — 24 hours (`packages/core/src/slot-limits.ts:33`). Because `reported` is keyed to a client that outlives the React tree, a failed report then stays indistinguishable from a successful one for the rest of the session. The test's own header comment already documented this in prose before today.

**Observed twice, on two different commits**, both `test-shards (ui-1, …)`, both `expected [] to deeply equal [ 'hero', 'sidebar' ]` at `:71`, both cleared by rerunning the identical commit:

| # | PR | Commit |
|---|----|--------|
| 1 | 1432 | `d120e5a9c` |
| 2 | 1430 | `4fc2a10a5` |

**Not introduced here.** Both the test and `use-placements.ts` are byte-identical to `origin/main` — `git diff origin/main aeb80e882` on either path is empty — and no lane in this train touched them.

**Deliberately not fixed here**, because it is outside every lane's ownership and predates the work. Suggested direction for whoever picks it up: stamp the dedupe key on *success* rather than before the send, so a refused report is never recorded as done and the 24-hour suppression cannot apply to a batch that never landed. That removes the race rather than synchronising around it — and the test should not be relaxed to hide it.

Fuller diagnosis and the per-occurrence evidence are in the two comments on #1432.

Issue #1372, opened 2026-08-16
No signed-out state: anonymous visitors get the full overlay and endless 403 retries


amrib247:
﻿A signed-out visitor to a preset-authed deployment sees the complete overlay chrome; every wire call correctly 403s under the identity doctrine; and the chrome handles it as an error storm: "Previous conversations could not load", the composer error line renders the server's developer-facing resolver paragraph verbatim, and every poller (resource loop, approvals feed, launcher badge) retries forever — the console fills with 403s. There is no signed-out state anywhere in the chrome and no forbidden-aware stop or backoff.

The 403 is right (the host owns sign-in). The chrome's handling makes every signed-out visitor of every preset-authed deployment see a broken-looking agent.

Proposed: type the no-identity refusal in the wire client; the overlay maps it to a quiet, host-brandable signed-out state; pollers stop on forbidden until a page signal; developer-facing copy never reaches end users.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

amrib247:
Fixed across #1442 (every poller stops on the first forbidden refusal; tab switches hold the latch; the documented vendo:identity-changed page signal and any successful read wake everything) and #1445 (the launcher stays, the panel opens to one quiet host-brandable line - signedOutNotice - and the developer-facing resolver paragraph never reaches the surface). Published in 0.29.0. Real-browser evidence on Linear ENG-424.

PR #1442, opened 2026-08-17
fix(ui): a forbidden wire is a full stop for every poller (#1372)

amrib247:
## What

The first half of lane 4 of the host-#2 fix pack (H2-E, #1372, Linear ENG-424): **a signed-out visitor's chrome stops asking.** Field case (expense.fyi, supabase preset): an anonymous visitor gets the full overlay, every wire call correctly 403s under the identity doctrine - and the chrome handles it as an error storm: every poller retries forever, the console fills with 403s, and the app-open retry ladder burns all its attempts on an answer that cannot change.

## How

A forbidden refusal is now a full stop, not a transient:

- **One internal per-client latch** (`hooks/identity-state.ts`, not exported - the HistoryPicker precedent): any wire failure with the typed `VendoError` code `forbidden` closes it (the CODE, never the HTTP status - `blocked` rides 403 too), any successful read opens it, and the page signal `vendo:identity-changed` (the `vendo:approvals-decided` naming precedent) opens it for SPA sign-ins. Full-page redirects remount everything anyway.
- **Gated pollers**: the approvals feed (arm + visibility return), the slot placements poller (read AND its renew report writes - an anonymous visitor generated write traffic too), the parked-press backstop (which swallows its own errors, so it reads the shared latch), and `useResource` - which has no client, so it latches itself with the same predicate and event; this covers all eight collection hooks at once. Tab switches hold every latch: a visibilitychange is not a sign-in.
- **`useApp`**: forbidden short-circuits the 3-attempt retry ladder to terminal on the first answer.
- **Docs**: the auth page's new "Signed-out visitors" section names the event and when to dispatch it.

## Tests

TDD with a stash-proof red baseline: the three new gate tests (resource stop + no-resurrect-on-tab-switch, resource wake-and-heal, feed stop-and-wake) all fail on pristine src and pass with the gates; the control ("any other failure keeps the cadence") passes both ways. The feed's pre-existing contracts ("stops entirely when the last surface unmounts", "pauses while hidden") still hold. Collateral suites green: hooks, app-load-retry, parked-approval-resolution, embeds - 66/66. Typecheck clean.

**Not visual**: no rendered output changes in this half - no screenshots owed. The visible half (the launcher opening to a quiet "Sign in to use the agent" panel instead of a broken-looking thread) follows as its own PR with real-browser screenshots, riding this latch.

### DIFF
--- a/.changeset/a-signed-out-visitor-stops-asking.md
+++ b/.changeset/a-signed-out-visitor-stops-asking.md
@@ -1,0 +1,14 @@
+---
+"@vendoai/ui": patch
+---
+
+A signed-out visitor's chrome stops asking. On a preset-authed deployment
+every wire call for a visitor with no session correctly answers forbidden —
+and every poller retried it forever, filling the console with 403s. A
+forbidden refusal is now a full stop for every poller (the shared resource
+loop, the approvals feed, the slot/placements poller and its report writes,
+the parked-press backstop), a tab switch does not resurrect them, and the
+app-open retry ladder no longer burns its attempts on a refusal that cannot
+change. Everything wakes together when the host dispatches
+`vendo:identity-changed` after an SPA sign-in (a full-page redirect remounts
+everything anyway), or the moment any wire read succeeds again.
--- a/docs-site/production/auth.mdx
+++ b/docs-site/production/auth.mdx
@@ -258,6 +258,25 @@ Audit readers that filter on webhook subjects should match
 
 ---
 
+## Signed-out visitors
+
+A visitor your resolver answers `null` for is refused with `forbidden` — the
+host owns sign-in, so that is correct. The chrome treats the first such
+refusal as a full stop: every poller (threads, approvals, slots) goes quiet
+instead of retrying a 403 forever, and switching tabs does not wake them.
+
+Everything resumes on its own after a full-page sign-in redirect. If your app
+signs users in without a page load, announce it:
+
+```ts
+window.dispatchEvent(new Event("vendo:identity-changed"));
+```
+
+Dispatch it after sign-in, sign-out, or a workspace switch — the chrome
+re-checks immediately.
+
+---
+
 ## Verify the wiring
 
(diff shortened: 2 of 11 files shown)


amrib247:
**Greptile triage:** the finding is real and fixed in e2111d00b. The latch gated the tick (renew + read) but not `report()` itself, which fires from `useReportSlot` on mount/prop change and flushed its own `POST /slots` via microtask regardless of the latch — exactly what the rendered test proved. The flush is now a named path that holds the queue while the latch is closed (including a microtask committed just before it closed) and the latch-open subscription flushes everything held — deferred, never lost. New test reproduces the scenario verbatim: latch, mount after, a settle window with zero requests, then the identity signal sends the held report exactly once.

amrib247:
**Greptile triage, round two:** real, and a sharper find than round one - fixed in the new commit. The stranding was my stamp-at-queue-time choice: a held report stamped `reported` but could die with the poller (`stop()` tears the identity subscription), and the stale stamp then silenced the same slot's remount for the whole refresh window. The stamp now lives in `flushReports` at **send time** - the send-failure `forget()` already encoded that rule - held duplicates dedupe against the queue itself, and the dedupe still re-arms the flush, so a remount after sign-in is precisely what sends the queue the teardown left behind. Red-first test reproduces the full scenario: mount latched, unmount, sign-in with nothing mounted, remount inside the refresh window, report lands.

amrib247:
**Greptile triage, round three:** real - the same class one layer deeper, and fixed in the new commit. A `useReportSlot`-only mount never calls `start()`, so no identity subscription existed to flush its held report. The queue now owns a lazy, self-disposing wake: created only when something is actually held, fires once on the first open transition, and deliberately survives `stop()` - it belongs to the queue, not the placement listeners, which is precisely the scoping round two got wrong. Net effect strictly strengthens round two: a held report now lands AT sign-in even with nothing mounted (no remount needed), and the send-time stamp keeps the remount duplicate-free. Both pinned red-first; 54 tests green across the gate and slot suites.

PR #1444, opened 2026-08-17
refactor(core,apps,vendo,ui,mcp)!: an app is its app.tsx, and no document keeps a tree

yousefh409:
The model never writes layout directly and no production door mints a tree-only
app document, so the stored-tree support was dead weight. This deletes it.

87 files, +714 / −2482.

## What went

- **Legacy tree-serve branch** (`persistence/open.ts`) — the `VENDO_TREE_FORMAT`
  branch and the unregistered-format fallback. A document with no usable source
  now refuses to open as `{kind:"failed"}` with an explicit legacy message
  instead of quietly serving a stored layout. A remix keeps its
  not-found→pending. Forced fallout, all provably dead once the branch went:
  `formingTreeOf`, and `createProgressiveQueryResolver` (~135 lines, zero
  callers), which dropped the `caller` arg from `serveOpenApp`/`createAppOpener`.
- **Dead paint path** (`doors/build-surface.ts`) — `paintSettledTree`/`emitView`,
  gated on `app.tree` and unreachable for every current write.
- **The stored `tree` field** — off `AppDocument`/`appDocumentSchema` and off
  `APP_DOCUMENT_FIELDS`. The schema is `passthrough`, so stripping is explicit:
  `withoutStaleTree` runs on read and on write, so a row carrying BOTH a source
  and a stale tree opens fine with the tree dropped — documents predating this
  change do not brick. Importing a `.vendoapp` whose document has a tree and no
  source FAILS (`assertOpenableSource`) — no silent conversion.
- Cascade from the field: `app-validation.ts` lost the tree/generated-presence/
  fn-machine rules (`fn:` is still gated at call time), `facts.ts` lost
  `treeOf`/`treeCheck` and 8 tree fact checks, `evidence.ts` deleted, `layer.ts`
  lost an unread `deps` option, `sync-impact.ts`/`box-lane.ts`/`automation/lane.ts`
  lost their tree branches.

## What explicitly stayed

`ui: "tree"` as the SURFACE KIND, and the tree as render language — `UIPayload`/
`TreeNode`, the renderer, streamed view parts, and the render-seam. Removing the
persisted field does not touch how apps paint.

`tree-dos.test.ts` KEEPS its DoS caps: `validateTree` is on the live render path
(`checking/component-screen.ts` over a freshly painted tree,
`ui/src/tree/convert-payload.ts` client-side on every open payload), so
`TREE_MAX_NODES`/`TREE_MAX_QUERIES`/dup-id/bad-root bound the RENDER payload and
never depended on the stored field. Only its `validateAppDocument`
fn-requires-a-machine block was tied to the removed field and is gone.

## Tests

60 files moved to code-first fixtures behind 2 shared helpers
(`apps/src/server/testing/screen-document.ts`, `vendo/tests/screen-fixture.ts` —
two because a cross-package helper would mean publishing a fixture on
`@vendoai/apps/testing`).

**Deleted tests are part of this approved change**, listed so nothing is silent:
2 files — `apps/tests/data-unavailable.test.ts` (9 tests; sole subject was the
deleted open-path query resolver) and `vendo/tests/wire/apps.partial-tree.test.ts`
(sole subject was the forming-tree geometry) — plus 18 individual tests whose
sole subject was a removed rule: inclient ×3, checking ×4, floor ×2,
contract/app-document ×6, contract-coverage ×1, tree-dos ×1, interchange ×1,
fn-runtime ×1.

New coverage is exactly three, per plan: open-refusal of a source-less document,
import-rejection of a tree-without-source archive, and stale-tree stripping.

## Called out for review

- **The one thing this PR adds: `CheckInput.rendered`.** The paint gate used to
  hand host checks the rendered tree on a synthetic `document.tree`. That is a
  real capability with two live consumers (`checking/app-floor.test.ts`,
  `fixtures/integration`'s `no-unmasked-accounts`), so it moved to an explicit
  input rather than being lost. Flagging it because a deletion PR that grows API
  surface deserves a second opinion.
- **Boundary crossed, wants the owner's ratification:**
  `apps/src/server/generation/` carries a do-not-edit-on-your-own-initiative
  note. Dropping the field made `asTree`/`asPayload` in `generation/engine.ts`
  uncompilable — 3 lines, zero callers repo-wide — so they and their now-unused
  `Tree` import were deleted. Nothing else in that folder was touched.
- **`PendingSurface.tree` keeps its wire field and its `embeds.tsx` consumer**
  with no server producer for now: absent reads as the beat bar, which is
  correct. The stacked follow-up PR gives it a producer again.
- **Document `components` islands now have no producer at open** — an island
  only ever reached a payload through a stored tree node. The
  `integration-browser` MCP fixture was the last thing minting one by hand and
  is now a screen calling `tools.*`.

## Known interim, accepted

The console in `vendo-web` mirrors this field: `previewable-doc.ts` requires
`doc.tree` before anything previews, and `preview-payload.ts`,
`app-doc-preview.tsx`, `appdoc.ts` and the public share page read it. Until its
tandem PR lands, a document that only ever had a stored tree previews as
"nothing to preview" — quietly, rather than saying it is a legacy document.
Those apps are already dead by design here, so this ships first and the console
fix follows as its own small PR.

Pre-existing, NOT fixed here: `APP_DOCUMENT_FIELDS` never copied
`componentTools`. With the tree gone that is now the only place a document names
a tool, so an imported app reports zero tool references to `sync`. Worth its own
fix.

---
## Summary by cubic
Removes the stored tree from app documents and serves only rendered trees. An app is its `app.tsx`; a tree exists only at render time, which stops documents from lying about what they render and simplifies open/import behavior.

- App behavior: `AppDocument.tree` is removed from `@vendoai/core` and `.vendoapp` archives. Opening a document with no usable source now resolves to `{kind: "failed"}` with a reason; previously it could serve a stored tree or spin. Rows written before this still open; the stale `tree` is stripped on read/write. Import refuses an archive that has a tree without source, with the same message as open. The render language (`UIPayload`/`TreeNode`) and `ui: "tree"` are unchanged.
- Checks and APIs: Add `CheckInput.rendered` to read the on-screen tree at paint time; replace any reads of `document.tree`. `createCheckingLayer` no longer takes `deps` (construct with `{ checks }`).
- Migration: Stop accessing `document.tree` anywhere. Ensure exports/imports include `source[SCREEN_FILE]` (`"app.tsx"`); a tree without source is refused. Remove any paths that serve a stored tree.

<sup>Written for commit c7bc86cc02112fd057a261a1aa93f30c39851c8e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1444?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Triage note: Greptile's P1 here ("legacy import accepts unusable source") is **fixed**, in `1947551c1` at the tip of the stack. The guard now measures `source["app.tsx"].text` non-blank, matching what `open()` actually requires, so a `{tree, source: {}}` archive is refused instead of persisting an app that can never open.

Two independent reviews found the same defect, so it converged three ways.

Fixes for this slice land on the tip branch (#1446) rather than here, because the stack merges into `main` in one motion — the base branch is deliberately left untouched to avoid a rebase. Review this PR for the deletion; read #1446 for the current state of these files.

yousefh409:
Shipped inside #1446's squash (the tip carried this branch's commits).

PR #1445, opened 2026-08-17
feat(ui): the overlay answers a signed-out visitor with one quiet line (#1372)

amrib247:
## What

The second half of lane 4 of the host-#2 fix pack (H2-E, #1372, Linear ENG-424), completing what #1442's poller latch started: **a signed-out visitor sees a quiet panel, not a broken agent.** The launcher still renders - nothing about wire health hides it - but the panel now swaps the thread for one host-brandable line ("Sign in to use the agent.", overridable via the new `signedOutNotice` overlay prop) whenever the per-client forbidden latch is closed. The conversation returns on `vendo:identity-changed` or the first successful wire read.

## How

- `useSignedOut(client)` (internal, `useSyncExternalStore` over the #1442 latch; SSR answers false so hydration always agrees).
- `SignedOutNotice` inside vendo-overlay.tsx (the HistoryPicker not-exported precedent) replaces the thread in the split rail while latched. The server's developer-facing resolver paragraph never renders (consumer-voice law; the connect-dock precedent).
- The thread hook's warm call now feeds the latch, so the panel is up on the overlay's FIRST wire contact instead of waiting for the next badge poll.
- Wire fixture gains `x-vendo-force-forbidden` (the `x-vendo-force-posture` precedent): a client that declares itself identity-less gets the preset hosts' real 403 on every read - used by the tests and the new harness scenario.

## Verification (the UI-affecting rule)

Real-browser evidence: a new harness scenario (`/signed-out`) whose client the wire genuinely refuses - the latch trips through the real path (warm call or badge poll, whichever lands first) - captured via Playwright chromium: **[screenshot on ENG-424](https://uploads.linear.app/d2f8bfb4-6137-45b3-b628-5235c594c6c8/35f6ca6c-fbc1-4c01-a95d-509c74dd7c89/79e1c77f-f685-48f5-9fe3-f81b2485a92e)** (panel open on the Maple theme: one centered line, launcher present, no error furniture; `docs/verification/` gitignores images, so it is Linear-hosted). The scenario + spec entry are committed, so CI regenerates the same capture.

Two Windows fixes fell out of the capture: `screenshotPath` now uses `fileURLToPath` (URL#pathname produced `C:\C:\...`, so every capture on Windows failed after the page proved ready), and the run used the dev server because `vite build` crashes on this machine (known env class - CI builds fine).

## Tests

jsdom suite `signed-out-overlay.test.tsx`, 4 red-first (stash-proven): the panel + launcher + no-resolver-paragraph pin, custom copy, the wire-driven path (no manual seeding), and the identity-signal return of the conversation. Gates: export-surface green (nothing new exported), approvals-feed 7/7, typecheck clean; the eject pack-and-inspect red is the pre-existing Windows tar class (reproduced on pristine state).

### DIFF
--- a/.changeset/the-panel-says-sign-in.md
+++ b/.changeset/the-panel-says-sign-in.md
@@ -1,0 +1,12 @@
+---
+"@vendoai/ui": patch
+---
+
+A signed-out visitor sees a quiet panel, not a broken agent. When the wire
+refuses a visitor for missing identity, the overlay's launcher still renders —
+nothing about wire health hides it — but opening it now shows one
+host-brandable line ("Sign in to use the agent.", or the new `signedOutNotice`
+overlay prop) instead of a conversation that can only error. The server's
+developer-facing resolver message never reaches the surface, and the
+conversation returns on `vendo:identity-changed` or the first successful wire
+read. Completes the signed-out state the poller latch started.
--- a/packages/ui/e2e/harness/main.tsx
+++ b/packages/ui/e2e/harness/main.tsx
@@ -1971,8 +1971,29 @@ function scenario(pathname: string): { title: string; theme?: Partial<VendoTheme
     case "/affordances": return { title: "Affordances (Maple) — copy, attach, connect dock", content: <AffordancesScenario theme={mapleTheme} />, ownProvider: true };
     case "/affordances-dark": return { title: "Affordances — dark", content: <AffordancesScenario theme={darkTheme} />, ownProvider: true };
     case "/toasts": return { title: "Toasts", content: <ToastsScenario />, ownProvider: true };
+    case "/signed-out": return { title: "Overlay — signed out (H2-E)", content: <SignedOutScenario />, ownProvider: true };
     default: return { title: "Unknown scenario", content: <p role="alert">Unknown browser scenario: {pathname}</p> };
   }
+}
+
+/** H2-E — the overlay for a visitor the wire refuses: its OWN client (the
+ *  latch is per client; the other scenarios' clients stay signed in) whose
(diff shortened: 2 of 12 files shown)


amrib247:
**Round-one triage:** both real, both fixed in the new commit. (1) CI complexity ceiling: the signed-out ternary tipped `VendoOverlay` over the lint bound the shell already sits at - the branch is extracted to a `RailBody` component, so the shell carries no new decisions; blocking-config eslint passes locally. (2) Greptile P1: `warmedClients` marked the client before `warm()` settled, so a refused warm - which primed nothing - permanently denied the post-sign-in conversation its cache warm-up. A forbidden refusal now forgets the mark (ordinary failures keep it - once per page life stays the rule for everything but a refusal), and the wire-driven test signs in after the refusals and asserts the remounted conversation re-warms, red-first.

amrib247:
**Round-two triage:** the P1 is real, and it also explains the ui-shard flake in the same round - both were the same race. Fixed by giving the latch an **epoch**: every open signal (the identity event, any successful read) starts a new one; noters stamp the epoch their request began in; a stale refusal is ignored. The warm mark is now epoch-keyed, so the post-sign-in re-warm is a property of the epoch changing rather than of delete-vs-remount ordering - which makes the previously flaky wire-driven test deterministic. The generation-guarded pollers (resource loop, approvals feed, placements) already drop stale results before noting, so they need no stamp - only the two unguarded noters (warm, parked backstop) do. Three unit pins on the epoch rules; 32 tests green across the latch suites; blocking-config eslint and both typechecks clean locally.

PR #1447, opened 2026-08-18
fix(repo): unbreak the root gate on Windows

YSharara1:
## What

Four unrelated fixes, all of which have to land together before a Windows
checkout can run a single root script:

| # | Fix | Symptom it removes |
| --- | --- | --- |
| 1 | floor `turbo` at `^2.10.2` (lockfile → 2.10.10) | every turbo script exits `3221225781` |
| 2 | normalise line endings in `build-icons.mjs --check` | `icons.gen.ts is stale` on a byte-identical file |
| 3 | `split(/\r?\n/)` in `dependency-guard.mjs` | `unreadable workspace entry: - "packages/*"` |
| 4 | `cross-env` on the three `test*` scripts | `'VITEST_MIN_FORKS' is not recognized…` |

None of these are visible in CI, which is Linux-only, and none of them are
the contributor's machine. Before this branch, `pnpm build`, `pnpm test`,
`pnpm typecheck` and `pnpm lint` all fail at the root on Windows — so the
only gate available to a Windows contributor is opening a PR and reading
GitHub Actions.

## Why

**1 — turbo 2.10.1 is a broken npm release.** Every turbo-backed script
died with exit `3221225781` = `0xC0000135`, `STATUS_DLL_NOT_FOUND`. VC++
redist was present, long paths were not the cause, and the platform
package was installed and intact (44,786,176 bytes).

Parsing the PE import table of `@turbo/windows-64@2.10.1`'s `turbo.exe`
names the missing library:

```
bcryptprimitives.dll
ghostty-vt.dll          ← not shipped, not a system DLL
kernel32.dll
…
```

The published tarball contains no DLL at all — `LICENSE`, `README.md`,
`package.json`, `bin/turbo`, `bin/turbo.exe`, and nothing else. Its
`dist.unpackedSize` (44,788,572) is the exe plus those four files exactly,
so this is not a partial download.

The import is unique to that one release. Neither `2.9.5` nor `2.10.2`
imports `ghostty-vt.dll`, and both print their version and exit 0 when run
directly on the same machine.

Our range was already `^2.1.0`, so only the lockfile was pinning the bad
version. Flooring at `^2.10.2` means it cannot be resolved again.

**2 — `build-icons.mjs --check` fails on a file it would generate
verbatim.** The check is an exact `!==` between the LF string it builds
and the file read off disk, and `core.autocrlf=true` puts CRLF in the
working tree. So the build fails with

```
packages/ui/src/kit/icons.gen.ts is stale — run: pnpm --filter @vendoai/ui build:icons
```

running `build:icons` produces an **empty** `git diff`, and the check
fails again. There is no sequence of commands that gets a Windows
contributor past it.

**3 — `dependency-guard.mjs` cannot read `pnpm-workspace.yaml`.** It
splits on `"\n"`, so each line keeps a trailing `\r`, which defeats the
`$` in `/^\s*-\s*["']?([^"'\s]+)["']?$/`. The first entry throws and the
guard never runs at all. (The committed blob is LF, so CI never saw it.)

**4 — `pnpm test` used POSIX inline env syntax.** `VITEST_MIN_FORKS=1 …
turbo run test` is not something cmd.exe can parse. `cross-env` is the
smallest fix that keeps the vars turbo already declares as task inputs.

## Verification

On Windows 11, pnpm 11.10.0, Node 24.18.0:

```
pnpm build                     Tasks: 21 successful, 21 total    2m8.01s
node scripts/dependency-guard.mjs    dependency-guard: OK (30 packages checked)
node scripts/portability-gate.mjs    portability-gate: all legs green
pnpm --filter @vendoai/ui check:icons    icons.gen.ts is current (227 icons)
```

Both `.gen.ts` files were deleted and re-checked out first, so they were
CRLF on disk when the check passed — the fix is exercised, not bypassed.

`portability-gate.mjs` was never broken; it is listed only because it runs
in the same `lint` script and is now reachable.


YSharara1:
One thing this branch deliberately does **not** fix, so it isn't mistaken for a clean bill of health.

With all four fixes in, `pnpm test` runs to completion on Windows but still has failures — and they are a fifth, separate bug of the same family. Every one is a spawn:

```
Error: spawnSync pnpm ENOENT
Error: spawn …\fixtures\host-app\node_modules\.bin\next ENOENT
```

On Windows `pnpm`, `npm`, `npx` and `node_modules/.bin/*` are `.cmd` shims, which Node cannot execute by bare name. Of the 38 spawn sites across 24 files, most launch `process.execPath` and are fine; only the ones shelling out to a package manager break. The fix is a small shared helper that appends `.cmd` on win32 for both the tool name and the `.bin` path — not `shell: true`, since several of these interpolate temp paths into the command.

That is a bigger diff touching test files across six packages, so it belongs in its own PR rather than riding along here.

(Unrelated to either: `@vendoai/genbench`'s failures are just a missing local Playwright browser — `pnpm exec playwright install chromium`.)

YSharara1:
Superseded by [runvendo/vendo-cloud#82](https://github.com/runvendo/vendo-cloud/pull/82).

This repo is now a one-way copybara projection of `runvendo/vendo-cloud` — its `main` is written only by `vendo-release-bot`, so a PR here cannot land. The change will appear here automatically once #82 merges.

Not a straight cherry-pick: the fold (#71, S11d) moved several of the files this touched, and every bug was re-verified as still live on current main before re-cutting. Details on #82.

PR #1466, opened 2026-08-18
perf(guard,harnesses,vendo): a tool call is decided once, and host calls stop reopening the socket

yousefh409:
## What changed

**One guard pass per tool call.** Every guarded call ran the whole policy pipeline twice for one logical call: the harness previews with `previewCheck` before it dispatches, and the guard-bound registry then evaluated the same call again from scratch moments later — the approvals read, the grants read, the rules, the org layer, and the judge (15s budget, paid twice). The preview's verdict now carries to the dispatch that follows it.

**Host-API calls ride the keep-alive pool** the store already uses. Node's stock dispatcher drops an idle socket after ~4s — shorter than the gap between two of an agent's tool calls — so nearly every host round trip paid a fresh TCP+TLS handshake. An explicitly passed `fetch` still wins (adapter rule).

## Why this cannot let a blocked call through

The verdict is **single-use and pinned** to the exact call: id, arguments, descriptor hash, subject, venue, presence, app, run key. A caller cannot preview one call and dispatch another, and cannot preview with a lenient descriptor and dispatch against the registry's real one — the hashes differ, the key misses, and the pipeline decides fresh.

The preview was always the whole evaluation; it simply never *spent* anything. The dispatch is what commits, and it commits the same things in the same order:

- **Breakers are read live**, never remembered — a previewed "run" that a concurrent call has since put over the write budget or the call-rate window still parks.
- **Authority is never remembered.** A standing grant is re-read, so a permission taken back between the two passes still stops the call (`grant-filter.test.ts` covers exactly this and stays green). The human's single-use yes is CLAIMED by the pass that dispatches, after the breakers so a call that ends up parked never burns the tap; a lost claim sends the call back through the full pipeline.
- **The kill switch** is still re-read uncached immediately before dispatch — untouched.
- **THE LAW's unattended gate** reads the same verdict and the same *effective* descriptor, both carried through — untouched.
- **An "ask" is never carried forward at all.** The caller waits for a person, and the tap that answers them IS the fresh verdict the dispatch reads.

## Tests

- `packages/guard/test/security/preview-reuse.test.ts` — one pipeline pass per call (counted through the judge), the budget spent exactly once, single-use verdict, no bleed across arguments or subjects, freeze mid-call still blocks, a budget exhausted mid-call still parks, a block audited once, and park → approve → run → ask-again.
- `packages/harnesses/tests/guard-single-pass.test.ts` — the SEAM, with the real guard on one side and real `turn-tools` on the other: no double anywhere in the pair.
- `packages/vendo/tests/host-fetch-keep-alive.test.ts` — a composed deployment's host tool call carries the shared pool's dispatcher, and an explicitly passed fetch wins.

Each new assertion was watched red before it was made green.

## Not in this PR

Two items of the approved slice are parked for the conductor, both because landing them needs files outside this lane:

1. **Bookkeeping fire-and-forget** (audit row + effect receipt). Implemented and measured: 20 existing tests across 7 files in `packages/guard/test/` go red, because they read `vendo_audit` with raw SQL against the PGlite fixture and so bypass any flush hook the guard can offer. Separately, the effect receipt IS the mutation-dedupe ledger — making it async opens a window between the in-flight slot's release and the receipt landing in which an identical call re-executes a real mutation. Both writes left awaited.
2. **Inference keep-alive.** `@vendoai/harnesses` may not import `@vendoai/vendo` (dependency-guard layering, and it would be a cycle), and `keepAliveFetch` deliberately stays out of core because core is bundled for edge targets. The seam that works is a `fetch` option on `VendoModelOptions` fed from `packages/vendo/src/models-config.ts` — off this lane's file list.
3. **The "duplicate" connect check** is not redundant: the `tool-bridge` preflight stops an approval being minted for an unconnected tool, and `connectGate.bind` is the only gate for the MCP door, automations, compound steps and BYO resume. Deleting either loses a documented guarantee. The real waste is `subjectHasToolkit` refetching on every negative lookup (`compose-discovery.ts`) — also off this lane's file list.

---
## Summary by cubic
Decides each guarded tool call once and keeps host and inference HTTP connections alive to cut duplicate policy passes and TCP/TLS handshakes. Previously the guard evaluated at preview and again at dispatch; now dispatch commits the exact previewed call while rechecking safety‑critical gates.

- Guard (`@vendoai/guard`): preview verdict is single‑use, keyed by id/args/descriptor hash/subject/venue/presence/app/run, and expires after 5s. Dispatch re‑reads the kill switch first, then breakers and the org‑admin layer, re‑resolves the effective risk grade, and re‑checks authority (standing grants re‑read; single‑use approvals are claimed at commit). “Ask” is never reused. Any landed call for the same subject (any grade, any run/session) voids that subject’s outstanding previews so the judge sees the latest trail.
- Host keep‑alive (`@vendoai/vendo`): composed host‑API calls default to a shared keep‑alive `fetch`; an explicitly passed `fetch` still wins.
- Inference keep‑alive (`@vendoai/harnesses`, `@vendoai/vendo`): adds `VendoModelOptions.fetch`; `@vendoai/vendo` passes the keep‑alive `fetch` to composed model seats; a host‑provided model or `fetch` is respected.
- Connect lookup (`@vendoai/vendo`): caches a negative “not connected” result for 1s so the bridge preflight and gate check don’t do two broker round trips per call; positives keep the 60s TTL.

<sup>Written for commit c392a78e77686176573d70d6a982901708f2b8b2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1466?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Shipped. This slice landed on `main` in the squash commit ebf101a568c56bca0e0a50766d14b7ee4e8e7be1 (#1463), which carried the whole four-PR latency stack.

`main` is protected by a merge-queue ruleset that mandates SQUASH, so the stack was merged once at its tip and this PR's head SHA (c392a78e7) never landed individually — GitHub therefore could not auto-close it. Verified: the tree of ebf101a56 is byte-for-byte identical to the stack tip 69900f7ec.

Closing as shipped, not abandoned. Branch `yousefh409/latency-tool-diet` is intentionally left in place.

PR #1467, opened 2026-08-18
fix(harnesses,vendo): a turn sets up in parallel, and the estimate stops billing tools it never sends

yousefh409:
Lane C of the latency build. Stacks on #1461 (`yousefh409/latency-telemetry`).

A turn did its setup one thing at a time. The system prompt was assembled only
after the opening store reads had finished, and inside the assembler the guard's
directions and the knowledge index — two waits on two different backends, neither
of which reads the other — were awaited one after the other. So a turn paid the
sum of three waits that had no reason to be in a line.

The prompt now starts with the turn and is awaited after the store phase, and the
assembler's two reads run together. **The assembled bytes are unchanged and
pinned**: section order comes from the `sections.push` sequence, never from which
read settles first, and `prompt-assembly-parallel.test.ts` assembles the whole
prompt with every section present and asserts exact equality against the
sequential form. The prompt-cache breakpoint layout in `loop.ts` is untouched
(`cache-breakpoints.test.ts` still green).

`promptMs` stays honest across the overlap. The phase split has to keep summing
to the wall clock, since `modelMs` is whatever the other four leave over — so an
overlapped span is billed once, to whichever phase was still waiting. `storeMs`
is still the store phase's own wall time; `promptMs` is now what the prompt still
cost the turn once the reads were done, which for a prompt that finished first is
honestly ~0. #1461's `TurnTimings` collector, its ttft mark and its
`durationMs`-from-the-top are all unchanged.

Three more places the turn was paying for nothing:

- The runtime re-validated the composed turn's transcript **against itself**.
  Composition hands it the very array it just read and validated, and one array
  cannot differ from itself, so `validateUpsert`'s double stringify per message
  and the history classification behind it were an O(n) pass per turn proving a
  tautology. The skip is keyed on that identity and nothing else — a
  client-posted transcript, the away runner, or a host's own runtime still takes
  both checks in full, which `transcript-restated.test.ts` pins.
- `vendo()` projected the host's catalog twice to set a turn up and again after
  every single tool call. It projects once now, and re-projects after a call only
  when that call can change what is reachable — the connector door.

And one real accounting bug: the compaction estimate billed every equipped tool
while the call only ever carries the active loadout, so a curated surface was
charged for a catalog it never sent. The trigger fired on tokens that were never
there and the shed floor was handed a figure the prompt had never reached. The
estimate bills the active set now. The characters-per-token ratio is unchanged.

## Not done in this PR (parked, with evidence)

- **Caching the knowledge manifest read / a failed `status()` with a TTL.** Both
  halves collide with shipped acceptance criteria.
  `packages/knowledge/tests/prompt-note.test.ts:153-179` asserts that the call
  immediately after a failed `status()` MUST retry, or a recovered engine is
  never picked up — the exact opposite of "the second turn makes no network
  call". And a TTL on the manifest read breaks `:106-120` and `:133-146`, which
  flip the manifest and require the very next resolve to see it. The read itself
  measures **21µs/turn** on a realistic 15KB manifest. Left untouched rather than
  reinterpreting either criterion.
- **Hoisting `residentTools` to per-process.** Every wrapper's `execute` closes
  over *this* turn's `turn`, and the object is mutated per turn by
  `refreshEquipped` and `find_tools` — sharing it across turns in one process
  would call a dead turn's guard and let one user's discovery equip another
  user's model. Measured value: 0.31 ms/turn at 50 tools. The two safe halves of
  that bullet (duplicated projection, per-call refresh) shipped.

## Gate

Scoped packages + typecheck, run locally on an idle box, output read back from files:

| command | result |
|---|---|
| `npx vitest run` (harnesses: runtime, transcript-restated, harness-state, vendo/) | 21 files, **253 passed**, 4 skipped |
| `npx vitest run` (vendo: 7 touched-surface files) | 7 files, **41 passed** |
| `pnpm typecheck` | 42/42 tasks successful |
| `pnpm build` | 21/21 tasks successful |
| `pnpm lint` | 0 errors |

---
## Summary by cubic
Turns set up in parallel and compaction estimates bill only tools the turn actually sends. Old: the prompt assembled after store reads and awaited guard directions then knowledge in sequence, and the estimate charged all equipped tools. New: the prompt assembles beside the store phase with directions+knowledge awaited together and deterministic section order; the estimate counts only the active loadout; slow knowledge failures cool off instead of taxing every turn.

- `@vendoai/vendo`: start system prompt assembly before store reads finish; await guard directions and knowledge together with `Promise.all`; section order is pinned by push order; `promptMs` now measures only post-store wait (phase totals still ≤ wall clock).
- `@vendoai/harnesses`: skip `validateUpsert` and history classification when `transcript.list()` returns the same array as `messages`; compute wrappers and the initial loadout from one listing; re-project the catalog only after connector‑discovery calls; compaction estimate uses `activeTools` when present, otherwise all equipped tools.
- `@vendoai/knowledge`: treat a slow `status()` failure as a 1‑minute cool‑off and serve the last good/empty index during it; fast failures are retried next turn.

<sup>Written for commit bb972a636f349ed9d716dda63adf8b13164e206b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1467?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/knowledge-sick-engine-cool-off.md
+++ b/.changeset/knowledge-sick-engine-cool-off.md
@@ -1,0 +1,12 @@
+---
+"@vendoai/knowledge": patch
+---
+
+A hanging knowledge engine no longer taxes every turn. The prompt index asks the
+adapter for `status()` when the sync state moves, and the wire client aborts that
+call at its own 30s timeout — so an engine that was UP but not answering charged
+30 seconds to every single turn before the prompt could be built, forever. A
+status check that fails SLOWLY now leaves the engine alone for a minute; the turns
+inside that window serve exactly what the failed check would have served, minus the
+wait. A check that fails FAST — a refused connection, microseconds — is unchanged
+and still retried on the very next turn, so a recovered engine is picked up at once.
--- a/.changeset/turn-startup-parallel.md
+++ b/.changeset/turn-startup-parallel.md
@@ -1,0 +1,24 @@
+---
+"@vendoai/harnesses": patch
+"@vendoai/vendo": patch
+---
+
+A turn stops doing its setup one thing at a time. The system prompt is now
+assembled BESIDE the turn's opening store reads instead of after them, and the
+two independent waits inside it — the guard's directions and the knowledge index
+— run together rather than in sequence; the assembled bytes are unchanged,
+because section order comes from the assembler and not from which read settles
+first. The runtime no longer re-validates the composed turn's transcript against
+itself: composition hands it the very array it just read and validated, and one
+array cannot differ from itself, so an O(n) double stringify per turn was
+spent proving a tautology. `vendo()` projects the host's catalog once to set a
+turn up instead of twice, and re-projects it after a tool call only when that
+call could actually change what is reachable — the connector door — rather than
+after every call.
+
+Fixes a compaction accounting bug in the same pass: the prompt estimate billed
+every equipped tool while the call only ever carries the active loadout, so a
+curated surface was charged for a catalog it never sent — the trigger fired on
+tokens that were never there, and the shed floor was handed a figure the prompt
+had never reached. The estimate bills the active set now. The characters-per-token
+ratio is unchanged.
--- a/packages/harnesses/src/runtime.ts
+++ b/packages/harnesses/src/runtime.ts
@@ -302,21 +302,33 @@ export function createHarnessRuntime(deps: HarnessRuntimeDeps): HarnessRuntime {
       let carried: string | undefined;
       try {
         before = await deps.transcript.list(input.ctx.principal, input.threadId);
-        // A client-sourced history is not trusted history. The shipped rule
-        // (`validateUpsert`) is the one that decides what a caller may change:
-        // fresh USER messages, and answering a pending approval. Anything else —
-        // an assistant message the client authored, a rewritten past turn — is a
-        // history-forging attempt and must not reach the model or the store.
-        const persisted = [...before];
-        for (const message of input.messages) {
-          validateUpsert(persisted, message);
-          const at = persisted.findIndex((candidate) => candidate.id === message.id);
-          if (at === -1) persisted.push(message);
-          else persisted[at] = message;
+        // The composed path RE-STATES its own transcript: `harness-turn.ts`
+        // answers `list` with the very array it passes as `messages`, having
+        // already applied the upsert rule to the one message the client
+        // contributed. One array cannot differ from itself, so both passes below
+        // are decided before they start — every upsert matches and the history is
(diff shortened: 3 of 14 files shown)


yousefh409:
Re: **Step-zero preparation replaces the turn snapshot** (`packages/harnesses/src/vendo/loop.ts:645-651`) — declining. The window described cannot open on any shipped path, and pinning step zero would contradict the documented behaviour.

First, scope: `prepareStep`'s per-step `activeTools()` re-read at `loop.ts:645` is **untouched by this PR**. The diff only moves the turn-start read to *before* the await, so compaction's token estimate and `streamText`'s top-level loadout agree about one moment.

**1. The loadout cannot mutate during `await turnModelMessages`.** Both callers of `startTurn` are in `vendo.ts`. The resident's `activeToolNames` (`packages/harnesses/src/vendo/vendo.ts:592-596`) changes only through `find_tools.execute` → `refresh()` (`vendo.ts:571-588`), which the model can invoke only *during a step* — i.e. after `startTurn` has resolved and `streamText` has begun. The subagent's is `() => [...equipped]` over a snapshot array (`vendo.ts:388`), immutable by construction. The reproduction mutates a synthetic closure between the two reads; no shipped `activeTools` can do that.

**2. Pinning step zero would be a regression against the stated contract** at `loop.ts:633-637`: a tool equipped mid-turn is meant to be choosable on the very next step. Freezing step zero to the snapshot makes the first step the one step that ignores the current loadout.

**3. The estimate/actual divergence named here is inherent to the per-step re-read** and exists identically at steps 1..n — compaction budgets once at turn start and cannot track a loadout that changes later. Pinning step zero removes none of that class; it only makes step zero inconsistent with every step after it.

No code change.

yousefh409:
Shipped. This slice landed on `main` in the squash commit ebf101a568c56bca0e0a50766d14b7ee4e8e7be1 (#1463), which carried the whole four-PR latency stack.

`main` is protected by a merge-queue ruleset that mandates SQUASH, so the stack was merged once at its tip and this PR's head SHA (bb972a636) never landed individually — GitHub therefore could not auto-close it. Verified: the tree of ebf101a56 is byte-for-byte identical to the stack tip 69900f7ec.

Closing as shipped, not abandoned. Branch `yousefh409/latency-turn-startup` is intentionally left in place.

PR #1472, opened 2026-08-18
feat(vendo,ui,core): the agent can read the files a user shared

yousefh409:
Part of a 4-PR stack (`main` → S1 #1464 → S2 #1465 → **S3 (this)** → S4). The merge unit is the tip — this PR is not meant to land alone.

## What it adds

A user can hand the agent a file, and the agent can read it back later. Two halves of one door:

- **Upload door** — a file dropped in the composer is *saved*, not just carried for one turn. It lands in the user's own drawer, scoped to that principal.
- **`vendo.putUserFile`** — the programmatic half, on the root handle, through the SAME server-side write, so the agent cannot tell a chat drop from a host-code push. Reaches the user the next time they chat; nothing is delivered and no turn is started. Same name replaces the file that was there (`/user` is last-write-wins).

The agent sees two **read-only** drawer tools: `vendo_user_files_list` and `vendo_user_files_read`.

## Public surface

- `Vendo.putUserFile({ principal, name, content, contentType? }) → Promise<UploadedFile>`
- `UploadedFile` exported from `@vendoai/vendo`
- `createVendo({ files })` — the BYO adapter slot; unset, the store's own blobs back it
- ui client: `files.upload`
- Boot summary gains a `files` row, but only when the host actually filled the seam
- Docs: `customize/user-data`

## Scope, stated plainly

- **PDF → app-shelf is deferred.** Not in this slice.
- **Binary reads return type + size only** — no content extraction.

## How it was tested

Containment is the whole claim, so it is proven by a full negative-path suite, not by a happy path: a principal cannot list or read another principal's drawer, and the read tools cannot write. Seam tests write through the real upload path and read back through the real tool path with no stub on either side (`user-files-seam`, `user-files-read.seam`, `user-files-client.seam`).

282 vendo + 1372 ui + 787 core tests green.

---
## Summary by cubic
Lets the in-product agent list and read files a user shared across conversations, while keeping transcripts light. Old: attachments lived on one message. New: non-images save under /user/files and messages send only a reference; images stay inline. Outside agents via the MCP door cannot access the user file tools.

**Review notes**
- Server: adds POST /files (raw bytes; ?name=…; requires header `x-vendo-upload`; 5 MiB cap checked on Content-Length and after read). Exempt from JSON CSRF. `UPLOAD_HEADER` exported from `@vendoai/core`.
- Programmatic write: `vendo.putUserFile({ principal, name, content, contentType? }) → Promise<UploadedFile>` in `@vendoai/vendo`; same write path as /files; not door-capped; same-name replaces.
- Agent/tools: adds read-only `vendo_user_files_list`, `vendo_user_files_read` (200-line windows; non-text returns type+size). Hidden from outside agents via MCP `withholdTools`; available to the in-product agent.
- Runtime/prompt: non-image files reach models as textual references, not base64; images remain inline. System prompt teaches persistence, snapshot semantics, and refresh-on-new-file behavior.
- UI: `client.files.upload(file)` added in `@vendoai/ui`; composer uploads non-images and sends only the reference; saved files render as a non-downloadable pill; images remain inline.
- Observability/config: optional `createVendo({ files })` to back uploads with your bucket; boot summary shows a files row when configured; `vendo doctor` reports the per-upload cap and whether files land in the Cloud store or this deployment’s store.

**Rollout**
- Update to the latest `@vendoai/ui` and `@vendoai/vendo`; built-in chat needs no code changes.
- If you post directly to /files, include `x-vendo-upload`, set the file’s Content-Type, and pass the file name as ?name=.
- If you previously read non-image bytes from transcripts, read from /user/files/* instead or expose your own route.
- Ensure guard posture allows the new read-only tools for the in-product agent; outside agents will not see them via MCP.

<sup>Written for commit 1b4267c3a5a8129e2c72c303ff4c2102bff0c008. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1472?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
Landed on `main` as part of the ingest stack via #1473 (squash `6bc5cc851`). Closing — the squash merge cannot mark this PR merged automatically.

PR #1479, opened 2026-08-18
feat(actions,apps,ui,vendo,core): vendo sync splits a <Remixable> into a port the model edits and wiring the host hooks up

yousefh409:
**Slice 4 — the splitter.** Base `main`. **67 files, +4493/−198.** (It previously read 310 files; that was an artefact of a stale base branch that had already merged, not the size of this change.)

`vendo sync` splits a `<Remixable>` host component into three things:

- a **PORTED half** — real React in the screen dialect, with generated shims preserving the original call-sites
- a **HOME half** — `.vendo/generated/remix-wiring.ts`, binding the component's data fetches to tools and its npm/sub-components to holes
- a **GRADE** — provably splittable, or a refusal with a loud line in the sync report while every other slot in the run still captures

`seed.from` then seeds the port as the new app's own `app.tsx`, and `edit_app` edits **that** rather than writing a replacement from nothing.

## What is proven, in a real browser, on a fresh production build

Both Maple cards port for real — via holes and `<button>`→Kit-Button rewrites — and an edited remix **paints**.

- **NetWorthView — `wishInSource: true`, `wishPainted: true`.** The edit lands and renders: "Tracked monthly" appears on the card with TOTAL BALANCE, the value, the delta pill, the range tabs and the real `AreaChart` hole all intact. Reproduced twice. A revert was also exercised, producing the chat saying verbatim *"I wasn't able to add the 'Tracked monthly' label — it failed both times I tried."*
- **QuickActionsView — `wishInSource: true`, checks floor passed, ZERO refusals.** Its stored row keeps `DEFAULT_ACTIONS`, `displayName` and `export default`, with the wish verbatim at line 41, and `background` is gone — replaced with allowlisted `fontSize` / `fontWeight` / `color` / `marginBottom`. **The card still shows the original because it is `review: true`**, and Vendo is correctly holding a review-kind remix pending Maple's reviewer (`remix/review.ts:177-181`). That is the design working as specified: port built, edit floor-accepted, pending approval.

The demo host hands the generated wiring to `createVendo` in `examples/demo-bank/src/vendo/server.ts`; both baselines (`NetWorthView.json`, `QuickActionsView.json`) carry their ported halves. `.vendo/generated/` stays uncommitted and is regenerated on every sync.

## Two limits, stated here rather than discovered later

**1. Painting *after* reviewer approval is unproven.** Exercising it needs a reviewer session, and nobody had a credential for one. Review-gated code is out of scope for this build, so this is a known blank rather than a claim.

**2. A second edit of an already-remixed app is still blind.** The workspace checkout only fills an *empty* workspace, so the first edit sees the port and later ones do not. Closing it would mean a store read on every ordinary edit, which `compose-apps.ts:232` explicitly warns against — so it is a **named follow-up, not a regression**: today's behaviour is unchanged from before this slice.

## Integration note

This branch carries `main` merged in (twice — `cca5cef4d`, then `50596b729`). The second merge conflicted in exactly two hunks, both in `packages/apps/src/server/checking/screen-typings.ts`, and both were resolved keeping **both** sides:

- `jsxFrame` — `key` on every display tag unconditionally (main's), with `className` only when `ported` (this branch's).
- `componentPropsText` — main's `typed` including the `TEXT_TYPE`→`ROW_TEXT_TYPE` formatter substitution, with this branch's `required` exception for a ported `Button`'s `label`.

Rebase was deliberately **not** used: `cca5cef4d` carries real conflict resolutions and four later commits were authored on top of that merged state, so replaying them would have discarded those resolutions and re-derived them.

Local gate after wiping `core`/`apps`/`ui` `dist` and rebuilding clean: **build 22/22, typecheck 44/44, lint 0 errors** (4 pre-existing demo-bank warnings). CI on this head is the gate of record.

## The generated-wiring import, and the follow-up it deserves

CI caught a real defect on this head's first validation against `main`: `examples/demo-bank/src/vendo/server.ts` imports `.vendo/generated/remix-wiring`, which is **gitignored build output**. `predev`/`prebuild` write it, but there was no `pretest` — and CI's `integration (node)` leg builds with `--filter=demo-bank^...`, *dependencies only*, so demo-bank never builds, `prebuild` never runs, and the import failed with `Cannot find module`. It broke any fresh clone too, not just CI.

Fixed here by adding `pretest`, mirroring the two hooks already beside it. Verified by deleting `.vendo/generated/` and running the suite: the hook regenerates the wiring and 28 files / 136 tests pass.

**Why that is safe rather than circular:** running `vendo sync . --no-ai` now leaves every tracked `.vendo` fixture byte-identical, so the `ported` halves both cards carry survive it, and `seed-survives-edit` (3 tests) plus the four vendo seed-fixture suites (6 tests) stay green. The older habit of `git checkout -- examples/demo-bank/.vendo/` after a sync was a symptom of the port **failing**; now that both cards port, sync rewrites the same bytes.

**That is a behavioural improvement worth naming, not a side effect of adding a hook:** the ritual of restoring `.vendo/` after every build no longer has a reason on this branch, and someone will otherwise keep performing it. Verified directly — a full `vendo sync . --no-ai` leaves the tracked `.vendo` tree with **zero** modifications, both `ported` halves intact and all four `.vendo/components/*.json` present.

**The real fix is a follow-up, and it is worth stating plainly.** A hard import of generated code contradicts this feature's own **fail-closed** posture — no wiring means no ✦, by design, which is exactly why an unsynced host shows no broken affordance. A hard import turns "you haven't synced yet" from a graceful no-features state into a **build failure**, so a host who adds `<Remixable>` and hasn't synced cannot compile their app at all. That is the worst possible first-run experience for the feature being shipped, and `server.ts`'s own comment — resting on "the `prebuild`/`predev` step, so it exists before anything imports it" — is the tell that the assumption was load-bearing and unguarded. Guarding that import so a missing wiring degrades instead of exploding is the correct answer; `pretest` unblocks it.

## Gate

All four required checks green on `b611c3fbb`. One `test-rest` failure on the way there was runner contention, not a defect: `packages/actions/tests/sync/sync.fixture.test.ts` timed out at 30s in CI but passes locally in **10.8s on an idle box**. Worth knowing that its headroom is now thin — roughly 3x on an idle machine — because the splitter added ~1,900 lines to the sync path this fixture exercises, so it will flake again under load.

---
## Summary by cubic
Splits each <Remixable> into a ported React component and a generated wiring module, so ✦ seeds start from real code and runtime paints with host classes and live props. Old: remix seeded a blank screen and ports could lose styling or fail without props; New: sync emits a port plus wiring, the floor grades/paints in the ported dialect using sample props, and holes resolve to host components.

- Sync & wiring
  - Carves unportable DOM/SVG/hooks into generated holes, rewrites plain <button> to Kit `Button`, narrows inline styles to the allowlist; `<article>` joins the display allowlist.
  - Captures declared sample props, grades with them, and stores a `ported` half on the baseline (source, tools, holes). Adds `pins.ported` to the report and prints both wiring hookup call sites.
  - Emits `.vendo/generated/remix-wiring.ts` and per-slot hole modules; a wiring-registered hole outranks a Kit name for ported nodes.

- Runtime & rollout
  - One dialect end to end: the floor derives `ported` from the row, stamps node `source`, and the renderer paints host `className` only on ported nodes; the VM mounts with the same props the grade used; press observations record without a blanket verdict on ports.
  - Seed starts from the port: `seed.from` writes the port as `app.tsx`, replays the first edit using `sampleProps`, retries once on a race; the screen agent checks out the stored screen only for remixes.
  - Required actions: git-ignore `**/.vendo/generated/`; run `vendo sync` in `prebuild`/`predev` (and `pretest` if tests import the wiring); import the generated file and pass it to both call sites — `createVendo({ remixWiring })` on the server and, if supported, `<VendoProvider remixWiring={remixWiring}>`.
  - Tests: adds end-to-end and seam coverage for ports/holes/props; raises the sync fixture timeout to avoid CI flake.

<sup>Written for commit dc35087fb6669ec78ac2831f32b9cbff3c9213c5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1479?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
@greptile-apps review

This PR has just changed shape substantially and **no machine reviewer has ever seen this diff against `main`**, so this is a deliberate review round before merge.

What changed since you last looked:
- Base retargeted from `yousefh409/remix-s3-wish-history` (long since merged) to **`main`**, and the draft lifted. The old 310-file diff was an artefact of that stale base; the **real slice is 67 files, +4493/−198**.
- `main` merged in at `50596b729`, conflicting in exactly two hunks — both in `packages/apps/src/server/checking/screen-typings.ts`, both resolved keeping **both** sides (`key` unconditional + `className` only when `ported`; main's `TEXT_TYPE`→`ROW_TEXT_TYPE` substitution + this branch's ported-`Button` `label` exception). Those two resolutions are the highest-value thing to check.
- The body has been rewritten from shipped reality; the previous "not a merge candidate" framing no longer applies.

Local gate, after wiping `core`/`apps`/`ui` `dist` and rebuilding clean: build 22/22, typecheck 44/44, lint 0 errors (4 pre-existing demo-bank warnings). CI on this head is the gate of record and is running now.

Two limits are stated in the body rather than hidden: painting after reviewer approval is unproven (needs a reviewer credential; review-gated code is out of scope), and a second edit of an already-remixed app is still blind because the workspace checkout only fills an empty workspace — a named follow-up, not a regression.

PR #1494, opened 2026-08-19
feat(core,agents): a turn's answer has one shape, and a tool must say what it does

yousefh409:
PR1 of the `@vendoai/agents` v1 DX stack. It builds no feature — it freezes the
contracts PR2–PR7 code against.

## What changed

**One shape for what a turn answered.** `packages/core/src/turn-result.ts` (new)
declares `TurnResult`: a turn finished (`ok`), needs a person (`interrupted`),
was called time on (`stopped`), or broke (`error`). A caller reads `status` once
and every field it then touches is present. `Interruption`, `Question`,
`Decision` and `Decisions` come with it, each with a zod schema, because those
four cross a wire (`@vendoai/ui` renders them, `POST /turns/:id/resume` accepts
them). `TurnResult` itself has no schema: the interrupted arm carries a live
`resume()` closure, and a function is not a wire value.

The interrupted arm's `resume()` returns the agents-level `Turn`, which core may
not import, so the union carries a second type parameter agents binds
(`TurnResult<T, Turn<T>>`). There is ONE definition of this union; nothing
re-declares it. `turnId`/`threadId` reuse the existing ids, `toolCalls` reuses
the `{ call, outcome }` pair `AgentRunReport` already carries, and `TurnUsage`
is field-for-field the `UsageTotals` a harness hands the runtime — restated in
core only because core may not depend on `@vendoai/harnesses`; harnesses' copy
is untouched.

The `input` interruption arm is wire-defined now and **emitted post-v1**: v1's
`ask_user` still ends the turn and the answer arrives as the next message.
It is frozen anyway so a shape added later is not a shape two shipped readers
reject.

**A tool must say what it does.** `tool()` now requires a non-empty
`description`, validated beside the existing name check. The reason is in the
error message, not a comment: the description is the only thing the model reads
when it decides whether to call the tool, so an undescribed tool is effectively
invisible.

**`tool()` infers `execute`'s input from zod.** `inputSchema` now takes a zod
schema as well as a raw JSON Schema; a zod one types `execute`'s `input`, and
the schema is converted to JSON Schema once, at `tool()` time, because a
descriptor crosses the wire to a model, a box and a browser, none of which can
run a validator. A raw JSON Schema is passed through byte-identical, and `input`
stays `Json` there — JSON Schema is data, with nothing in it for TypeScript to
read.

> **Type inference only — NOT runtime validation.** Nothing new checks arguments
> before `execute` runs. What reaches an existing host's tool is exactly what
> reached it before.

**One install.** `@vendoai/agents` re-exports `vendo()` from its root and the
turn-contract types from `@vendoai/core`, so a host names a turn's answer without
adding a dependency.

## The one forced amendment to the spec

The spec's hello-world reads
`import { agent, tool, claudeCode } from "@vendoai/agents"`. `claudeCode` could
not go on the root barrel: `packages/vendo/src/server.ts` imports that barrel and
must bundle for a Worker, and the Claude Agent SDK reaches `readline` /
`async_hooks`. On the barrel it broke two `portability-gate.mjs` legs (server
entry, workerd fixture boot) — which is exactly why `claudeCode` is a subpath in
`@vendoai/harnesses` and `vendo()` is not.

So it is a subpath here too — `@vendoai/agents/claude-code` — and the import
splits over two lines. The spec's actual promise, "install one package", is
intact. Escalated separately.

## Retired-behaviour test edits (called out explicitly)

`tool()` accepting a missing or empty `description` is retired, so every call
site that relied on it had to gain one. Source and docs: `packages/agents/README.md`.
Tests, all mechanical — a `description` added, no assertion changed:
`tests/agent.test.ts:95`, `tests/door.test.ts:139`, `tests/respond.test.ts:213`,
`tests/session.test.ts:223`, `tests/session.test.ts:325`, and ten call sites in
`tests/tools.test.ts`. Every pre-existing assertion in those files (name pattern,
risk default, `outputSchema` passthrough, collision detection) still passes
unmodified.

New tests live beside them: `tests/tools.test.ts` covers the missing and the
whitespace description (and that the message names the fix), zod inference plus
the JSON Schema it produces, and raw JSON Schema passthrough;
`packages/core/tests/turn-result.test.ts` covers each schema's valid and invalid
wire values — an unknown `Interruption.type`, a `Decision` that is neither
literal nor an answers object — and pins each `TurnResult` arm's narrowing.

## Gates

`build` · `typecheck` · `lint` (dependency-guard, portability-gate, eslint) ·
`packages/core` suite (800 tests) · `packages/agents` suite (154 tests) — all
green. The full affected cone is CI's job.

---
## Summary by cubic
Unifies a turn’s return contract as `TurnResult` and makes `tool()` descriptions mandatory so models can discover tools. Also adds zod-powered `inputSchema` typing and exposes both engines from `@vendoai/agents` with `claudeCode` via a subpath for Worker portability.

- `@vendoai/core`: New `TurnResult` discriminated union (`ok` | `interrupted` | `stopped` | `error`). `Interruption`, `Question`, and `Decision` have zod schemas for wire safety; the `interrupted` arm carries a live `resume()` closure and stays schema-less. The `input` interruption shape is defined now but not emitted in v1 (freezes the wire for post-v1).
- `@vendoai/agents`: `tool()` now requires a non-empty `description` (old: optional/empty allowed). `inputSchema` accepts zod or raw JSON Schema; zod types `execute(input)` and is converted once to JSON Schema in the descriptor. No new runtime validation is introduced. Root re-exports `vendo()` and core turn-contract types; `claudeCode` moves to `@vendoai/agents/claude-code` to keep the barrel Worker-safe. Tests pin the new contracts and error messages.

**Migration**
- Add a meaningful `description` to every `tool({...})`. Calls without one now throw at boot.
- Import the Claude builder from `@vendoai/agents/claude-code` instead of the root barrel.
- Optional: switch `inputSchema` to a zod schema to get typed `execute(input)`; raw JSON Schema continues to work unchanged.

<sup>Written for commit 2bd396f1acfff3c7cca8504954309873cf46070f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1494?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/agents/README.md
+++ b/packages/agents/README.md
@@ -12,8 +12,9 @@ import { claudeCode } from "@vendoai/harnesses/claude-code";
 const support = agent({
   name: "support",
   harness: claudeCode({ model: "claude-sonnet-5" }),
-  // `outputSchema` is optional: the result shape the model is told to expect.
-  tools: [api(), tool({ name: "refund", risk: "write", inputSchema, outputSchema, execute })],
+  // `description` is required — it is all the model reads to decide to call the
+  // tool. `outputSchema` is optional: the result shape the model is told to expect.
+  tools: [api(), tool({ name: "refund", description, risk: "write", inputSchema, outputSchema, execute })],
   mcp: [{ url: "https://mcp.example.com", headers: { authorization: "…" } }],
   skills: ["./skills/product-docs"],
   egress: ["api.stripe.com"],
--- a/packages/agents/package.json
+++ b/packages/agents/package.json
@@ -32,6 +32,10 @@
     ".": {
       "types": "./dist/index.d.ts",
       "default": "./dist/index.js"
+    },
+    "./claude-code": {
+      "types": "./dist/claude-code.d.ts",
+      "default": "./dist/claude-code.js"
     }
   },
   "files": [
--- a/packages/agents/src/claude-code.ts
+++ b/packages/agents/src/claude-code.ts
@@ -1,0 +1,11 @@
+/**
+ * `@vendoai/agents/claude-code` — the builder engine, from the one package a
+ * host installed.
+ *
+ * A subpath rather than a root export for the reason it is one in
+ * `@vendoai/harnesses`: the Claude Agent SDK reaches Node built-ins, and the
+ * root barrel is bundled for Worker targets through `packages/vendo/src/server.ts`
+ * (portability-gate.mjs). Everything here is a re-export — one definition, in
+ * harnesses.
+ */
+export { claudeCode, type ClaudeCodeOptions } from "@vendoai/harnesses/claude-code";
--- a/packages/agents/src/index.ts
+++ b/packages/agents/src/index.ts
@@ -1,8 +1,8 @@
 /**
  * `@vendoai/agents` — spawn a governed, harness-grade agent in any Node
  * backend in a few lines. One runtime, always host-run; a Vendo Cloud key
- * fills the sandbox slot when it is left unset. Harness factories live in
- * `@vendoai/harnesses` (`claudeCode`, `vendo`).
+ * fills the sandbox slot when it is left unset. The engines are DEFINED in
+ * `@vendoai/harnesses` and re-exported here, so a host installs one package.
  */
 export {
   agent,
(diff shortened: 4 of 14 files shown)


yousefh409:

PR #1495, opened 2026-08-19
feat(agents): serve() is the lifecycle, so a standalone agent's triggers fire

yousefh409:
## What changes

`agent.on("0 9 * * 1", "summarize the week")` used to be a declaration with nowhere to go: it validated the cron and stashed it, and only `createVendo`'s boot ever reconciled it. A standalone `agent()` had triggers that could never fire.

`serve({ agents })` is the missing lifecycle:

```ts
support.on("0 9 * * 1", "summarize the week and email ops");
const runtime = await serve({ agents: [support] });
await runtime.close();
```

A trigger stays INERT until a lifecycle reconciles it — `serve()` or `createVendo`'s boot. That is now said plainly in `on()`'s hover doc, which previously named only the umbrella.

## How

An assembler, not new machinery — every mechanism already existed. ~95 lines: build the engine from the first agent's composition, register one runner per agent by name, reconcile the declarations, start the engine's own ticker, hand back its own stop function.

Two decisions worth calling out:

- **Owner is `vendo:code`**, byte-identical to the umbrella's. `reconcileAutomations` filters stored records on the owner subject, so a different one would make `serve()` and `createVendo` blind to each other's records and each would disarm what the other armed.
- **A failed boot reconcile REJECTS**, where the umbrella's deliberately swallows. The umbrella's rides a memoized `ready()` latch, so a rejection there becomes every route's answer for the life of the process. `serve()` has no latch to poison and is an explicit awaited call — a host holding a resolved runtime has to be able to trust its triggers are armed.

## Layering

One new allowed edge, `agents -> automations`, in `scripts/dependency-guard.mjs`, plus the honest manifest entry. No cycle: `@vendoai/automations` depends on `@vendoai/core` alone. A standalone install now carries `croner` and `jsonata`; accepted, and the same shape of tradeoff already documented for `mcp` and `apps`.

## A correction to the spec

The spec said "cron runs as the agent principal." It does not, and this PR does not make it. A firing's principal is the record's owner (`vendo:code`); the agent's name rides the record and selects the brain. Making the literal reading true would mean owning records per-agent, which reintroduces the duplicate-arm problem above. Flagged rather than silently changed.

## Tests

8 new in `packages/agents/tests/serve.test.ts`: inert before `serve()`; reconciled as armed and owned by `vendo:code`; a due schedule fires through the right agent's runner; `close()` stops the ticker; two agents each get their own runner; a non-`agent()` value is a clear error; a reconcile failure rejects rather than handing back a dead runtime. Clock driven by faking only `setInterval`/`clearInterval`, so the store's real async I/O is untouched.

## Gate

`build` · `typecheck` · `lint` (dependency-guard OK, 30 packages) · `packages/agents` 17 files / 162 tests — all green, re-verified after restacking onto PR1.

Stacked on #1494.

---
## Summary by cubic
Makes `serve({ agents })` the runtime lifecycle so `agent.on(...)` triggers arm and fire in standalone installs, and adds a single HTTP mount for a standalone agent. Also introduces durable resume of interrupted turns across processes, per-user memory, and a user-bound facade.

- Lifecycle: assembles the engine from the first agent’s store/guard/tools, registers each agent by name, reconciles as owner `vendo:code`, starts the scheduler; `close()` stops the ticker. Triggers are inert until a lifecycle runs. Duplicate agent names fail; boot reconcile rejects.
- Multi-agent composition: secondary agents now run against the deployment’s store/guard/files, so approvals, runs, and threads land where hosts read them.
- HTTP mount: `agent.handler({ basePath, resolveUser })` serves the door, the approvals/grants wire under this basePath, and chat/threads; unresolved users get 401. `@vendoai/agents/http` provides the shared route runtime used by `@vendoai/vendo`.
- Interrupted turns: `turns.list({ status: "interrupted" })` and `turns.resume(turnId, decisions)` re-dispatch approved calls exactly once. Guard rows now carry `turnId` and `agent`; resumes are scoped to the agent that parked them; expiry is 7 days; partial decision maps are rejected.
- Memory: per-user memory with a capped `[Memory]` prompt block and a `remember` tool. Reads auto-inject; writes are explicit and guard-checked; rows are strictly scoped to the subject.
- Facade: `agent.forUser(subject, { profile, context })` binds a person once and exposes their chat, turns, threads, and memories. `run({ as })` and `session()` are deprecated but unchanged. Resumes carry bound context (never headers).
- UI: `@vendoai/ui` ships `useVendoChat` for the standalone mount; approvals post to the permission wire so parked calls actually run.
- Layering: adds `@vendoai/automations` to `@vendoai/agents` (no cycles); standalone includes `croner` and `jsonata`. Spec clarified: cron runs as code owner, not the agent principal.

**Rollout**
- Standalone hosts using `@vendoai/agents` must call and await `serve({ agents: [agent(...), ...] })` at startup.
- Mount `support.handler({ basePath: "/api/agent", resolveUser })` (or `agentHandler(support, …)`) to expose chat and the permission wire.
- Prefer `agent.forUser(subject)` over `run({ as })`/`session()`; use `user.turns.resume(turnId, decisions)` for durable resumes (7‑day TTL).
- Enable memory with `agent({ memory: true })` or supply a `MemoryAdapter` if desired.

<sup>Written for commit 7bfeb8ef04597600de8b4c6f731cef13b8d2c1d0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1495?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/fixtures/integration-browser/e2e/agent-chat.spec.ts
+++ b/fixtures/integration-browser/e2e/agent-chat.spec.ts
@@ -1,0 +1,84 @@
+/**
+ * THE SEAM: the shipped `useVendoChat` in a real browser against a real
+ * `agentHandler()` over real HTTP, with no stub on either side.
+ *
+ * One script, one run, the whole flow — a turn that parks on a real guard
+ * approval, a person answering it in the page, and the parked call actually
+ * running. What makes it a seam test rather than a demo is the last assertion:
+ * the refund is read back from the SERVER's own record, so a page that merely
+ * rendered a convincing sentence cannot pass.
+ */
+import { expect, test } from "@playwright/test";
+
+test("a parked approval is answered in the page, and the parked call then runs", async ({ page, request }) => {
+  await page.goto("/agent.html");
+  await expect(page.getByTestId("composer")).toBeVisible();
+
+  // Nothing has run yet: the mount's own record is the baseline.
+  expect((await (await request.get("/__agent/refunds")).json() as { refunded: string[] }).refunded).toEqual([]);
+
+  await page.getByTestId("composer").fill("refund invoice 7");
+  await page.getByTestId("send").click();
+
+  // 1. The turn reached the agent and came back with the conversation's id on
+  //    the response header — the round trip, not a rendered guess.
+  await expect(page.getByTestId("thread-id")).toHaveText(/^thr_/);
+
+  // 2. It PARKED. The guard's ask crossed the stream as a real
+  //    `approval-requested` part and the hook surfaced it as an interruption,
+  //    naming the tool the agent is waiting to run.
+  await expect(page.getByTestId("interruption")).toHaveCount(1);
+  await expect(page.getByTestId("interruption-tool")).toHaveText("host_refund");
+  await page.screenshot({ path: "e2e/artifacts/agent-chat-parked.png", fullPage: true });
+  // Still nothing has run — the park is real, not cosmetic.
+  expect((await (await request.get("/__agent/refunds")).json() as { refunded: string[] }).refunded).toEqual([]);
+
+  // 3. The person answers, in the page, through the hook's own `resume`.
+  await page.getByTestId("approve").click();
+
+  // 4. The parked call RAN — server-side evidence, not screen text.
+  await expect(async () => {
+    const { refunded } = await (await request.get("/__agent/refunds")).json() as { refunded: string[] };
+    expect(refunded).toEqual(["inv_7"]);
+  }).toPass({ timeout: 20_000 });
+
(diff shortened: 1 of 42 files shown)


yousefh409:

PR #1498, opened 2026-08-19
feat(agents,harnesses): an agent answers in one line, and a turn that needs a yes stops asking you to run it again

yousefh409:
## What changes

`chat()` and `run()` now return the same thing — one `Turn`, resolving to one `TurnResult`:

```ts
const turn = user.chat("refund invoice #7", { threadId });
turn.threadId; turn.turnId;          // readable before it finishes
const result = await turn;           // ok | interrupted | stopped | error
```

Two things a caller could not do before:

- **A turn that hits an approval now ENDS and says so** (`status: "interrupted"`, carrying the parked calls) instead of holding the connection open for 90 seconds hoping someone taps. The away lane's old advice — "show these, then call `run()` again" — is retired along with `refs.approvals`.
- **The turn is the drainer of record.** It executes and persists whether or not anyone reads `events` and whether or not anyone awaits it. `events` is an independent single-reader tap, not the thing that makes the turn happen.

`respond()` is untouched and undeprecated — it stays the one-liner for an existing route.

## How

The chat lane and the away lane turned out to be one turn assembled twice, so this is mostly a deletion: **`away.ts` goes from 696 lines to 198**, keeping only `RunOptions`, the away ctx, the summary narrowing, and `awayRunner`'s `AgentRunReport` mapping (which the automations engine consumes and which is NOT retired). Everything else — the turn, the event queue, the typed-output tool, the failure watch, the thread/id/gate opening — is now one implementation in `turn.ts`.

`session.stream()` stays as it is: it is the blocking streaming-`Response` lane, and unifying it would have broken `respond()`.

## Notable

- **`turnId` is minted once.** One line in `packages/harnesses/src/runtime.ts:294` now honours a ctx that already carries a turn id instead of overwriting it. Without it, "readable before completion" would have meant two ids per turn and a `TurnResult.turnId` joined to nothing. Nobody sets it before `runtime.run` today, so it is behaviour-identical for every existing caller.
- **A park's refusal now depends on who is there.** Previously every early-ended turn told the model "nobody is here to give it" — true for a 3am automation, false in a live chat where the person is right there and answering asynchronously. 12 lines in `turn-tools.ts` pick on `ctx.presence`; the away wording is byte-identical to today's. In production `!interactive` and `presence === "away"` were equivalent until this PR, so no shipped path changes what it says.
- **Explicit-`undefined` `threadId` throws synchronously** (anti-fork), while a foreign thread still rejects as `not-found`.
- `turnId` is stable across park→resume, so PR3's `turns.resume(turnId)` addresses one turn.

## Retired-behaviour test edits

Listed in full because the law requires it: `tests/run.test.ts` (6 assertions moved off `AgentReport`/`refs.approvals` onto `TurnResult`), `tests/unattended.test.ts` (3, same retirement), and one line in `packages/harnesses/tests/stale-approvals.test.ts` where a test double's ctx said `present` while asserting the away wording. `away.test.ts`, `session.test.ts` and `respond.test.ts` pass unchanged.

## Gate

`build` · `typecheck` (42/42) · `lint` (dependency-guard OK; portability-gate all 9 legs green) · `packages/agents` 166 · `packages/harnesses` 568 · `packages/core` 800 — all EXIT=0.

Stacked on #1494.

---
## Summary by cubic
Unifies `chat()` and `run()` under one turn contract so both return a `Turn` that resolves to a single `TurnResult`. Turns that need approval no longer block or ask you to run again; they end as `interrupted` and carry the parked calls.

- The turn is the drainer of record. It executes and persists even if no one reads `events` or awaits it. `threadId` and `turnId` are readable immediately and match what gets written.
- Old: away runs held the connection ~90s and required a second `run()` after approvals via `refs.approvals`. New: the first turn ends `interrupted` and can be resumed; `refs.approvals` is gone.
- `respond()` and `session.stream()` stay the blocking streaming lane and are unchanged.

**Review notes**
- Most logic is new in `packages/agents/src/turn.ts`. `packages/agents/src/away.ts` deletes duplicate turn code and delegates to the shared turn.
- `packages/harnesses/src/runtime.ts` now honors a pre-set `turnId` in ctx (keeps one id per turn; enables “readable before completion”).
- `packages/harnesses/src/turn-tools.ts` adjusts refusal text based on `ctx.presence` (away vs present).
- Public API: `packages/agents/src/index.ts` stops exporting `AgentRun`/`AgentReport`; adds `Turn`, `RunEvent`, `ChatOptions`. `awayRunner` remains and maps to `AgentRunReport` for automations.

**Migration**
- Update callers of `agent.chat()` and `agent.run()` to handle a `Turn`:
  - Read `turn.threadId`/`turn.turnId` immediately; `await turn` for `{ status: "ok" | "interrupted" | "stopped" | "error" }`.
  - Replace `report.refs.approvals` with `result.status === "interrupted"` and read parked calls from the result. Do not call `run()` again to continue; plan to resume the same `turnId`.
- Do not pass `threadId: undefined`. Omit `threadId` or pass a valid id; explicit `undefined` now throws synchronously.
- Treat `events` as an optional, single-reader tap; do not rely on it to drive execution.

<sup>Written for commit f94b34f6a15779a56ba45c7b2b7f8de6bce5085d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1498?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/agents/src/agent.ts
+++ b/packages/agents/src/agent.ts
@@ -30,7 +30,8 @@ import { createStore, hostedStore, storeFiles, type VendoStore } from "@vendoai/
 import type { LanguageModel, UIMessage } from "ai";
 import { randomUUID } from "node:crypto";
 import { declareAutomation, type OnOptions } from "./automations.js";
-import { startRun, type AgentRun, type RunOptions } from "./away.js";
+import { startRun, type RunOptions } from "./away.js";
+import { startChat, type ChatOptions, type Turn } from "./turn.js";
 import { resolveDoor, type DoorConfig } from "./door.js";
 import { withEgress, type EgressConfig } from "./egress.js";
 import { PERMISSIONS_PATH, schemaReadyPrincipal, type AgentPrincipal } from "./permissions.js";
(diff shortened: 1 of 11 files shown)


yousefh409:

PR #1499, opened 2026-08-19
fix(core): a tenant's connectors survive a Cloud-hosted store (#1497)

### LINKED ISSUE #1497: vendo_tenant_connectors is missing from the engine allowlist — tenant connectors are dead against a hosted store
`vendo_tenant_connectors` is written as an ordinary records collection at
`packages/vendo/src/tenant-connectors.ts:78` (`const COLLECTION = "vendo_tenant_connectors"`),
but it is missing from the engine allowlist registry in
`packages/core/src/engine-collections.ts:37-104`. Every sibling collection is
there — `vendo_channel_links`, `vendo_app_seen`, `vendo_slots`, and the rest,
each with a `// <CONST>, <file>:<line>` comment naming its writer.

Against a Cloud-hosted store (the posture a host with `VENDO_API_KEY` and no
store slot gets) the hosted door refuses it:

```
collection "vendo_tenant_connectors" is not an engine collection (engine allowlist v5)
```

Observed in `examples/demo-bank` running `next dev`. A local store has no
allowlist in front of it, which is why the suite is green — the same class of
miss `packages/vendo/tests/channel-hosted-store.test.ts` was written to catch
after `vendo_channel_links` shipped missing from this list.

Filed separately from #1496 (the production QuickJS engine boot). Different
package, different mechanism, no shared cause: #1496 is a module-resolution
failure in `packages/apps`; this is a missing registry entry in
`packages/core`. Adding this entry needs `ENGINE_ALLOWLIST_VERSION` bumped and
Vendo Cloud's mirror of the list updated in the same session.

yousefh409:
Fixes #1497.

## The bug

`vendo_tenant_connectors` — the collection tenant registrations are written to
(`packages/vendo/src/tenant-connectors.ts:78`) — was missing from
`ENGINE_COLLECTION_REGISTRY`. Against a Cloud-hosted store (the posture a host
with `VENDO_API_KEY` and no store slot gets) the engine door refused the first
write:

```
collection "vendo_tenant_connectors" is not an engine collection (engine allowlist v5)
```

`register`, `list`, `test` and every tenant's own tools were dead on arrival on
a live deployment. The suite stayed green because every existing
tenant-connector test composes a local store, and a local store has no
allowlist in front of it — the same class of miss the text channel's three
collections shipped with (#1276).

## The fix

One registry entry, in the generic-table section, with the `file:line` comment
every neighbour carries:

```ts
vendo_tenant_connectors: { kind: "storage" }, // COLLECTION, packages/vendo/src/tenant-connectors.ts:78
```

`ENGINE_ALLOWLIST_VERSION` moves 5 → 6, as that constant's contract requires;
the historical "5, not 3 or 4" note is replaced with one that tells the truth
about the new number. The stale "37 collections with nothing to walk forward
through" count becomes 38 — a direct consequence of the entry.

## The test

`packages/vendo/tests/tenant-connectors-hosted-store.test.ts` drives
`createTenantConnectors` through the real `hostedStore` + `hostedStoreOps` +
`fakeConsole`, whose gate is the live door's gate. No stub on either side.
Three cases: the register/list/test/remove round trip, per-org overlay
isolation, and a check that the row crossed the engine door under the right
collection name with the token on the secrets door only.

**Red-green-red proof** — remove ONLY the registry line, rebuild `@vendoai/core`:

| run | result |
|---|---|
| baseline | 3 passed |
| registry line removed | 3 **failed**, `collection "vendo_tenant_connectors" is not an engine collection (engine allowlist v6). App data belongs to the appData family…` |
| line restored | 3 passed |

## A trap for whoever writes the next test against this API

**Inspecting recorded wire REQUESTS proves nothing here. Assert the answer's
`status` first.**

The first draft of case 3 passed while the bug was still present. It watched the
requests the client recorded — and `register` does not throw when the door
refuses it, it RETURNS a refusal. So a 403 and a success produce the identical
request log, and the assertion could not tell them apart. A test that can only
ever be green is not a test.

It is the mocked-counterparty failure in a new costume: not a stub on either
side this time, but an observation point that sits upstream of the disagreement,
so producer and consumer still cannot contradict each other. This is the fifth
or sixth time this shape has surfaced in this codebase. The fix is the same
every time — observe the ANSWER, not the attempt. Case 3 now asserts
`status: "ok"` before it looks at anything else, and all three cases go red
together when the registry line is removed.

## Other tests checked

- `packages/core/tests/engine-collections.test.ts` hardcoded the version and the
  count — updated to 6 / 39.
- `packages/store/tests/engine-allowlist.test.ts` needs nothing: it mirrors only
  the reserved and dedicated collections, and this one is generic.

## Changeset

`@vendoai/core: patch`, matching the identical prior fix (85fc7329b, the
channel's collections joining this same allowlist). No new API — a broken thing
made to work.

## Gate

`pnpm build && pnpm test:affected && pnpm typecheck && pnpm lint` on the touched
scope. Build, typecheck and lint all exit 0. `test:affected` pulled in 45 tasks
(core changed, so nearly everything): 44 successful — `@vendoai/core` 53 files,
`@vendoai/vendo` 253 files / 2618 tests, `@vendoai/store` 63 files, and the rest
green. `demo-bank`'s `away-drill.test.ts` went red under load (one test took 92s
against a ~5s budget while three agents shared the box); re-run alone it is
4 passed in 25s. Unrelated to this change either way — it exercises unattended
automation runs, not the allowlist.

**Note for the same session:** Vendo Cloud's mirror of this list needs the entry
and the version bump too, per the issue.

---
## Summary by cubic
Unblocks tenant connector registration on Cloud-hosted stores by adding `vendo_tenant_connectors` to the engine allowlist. Previously, hosted writes failed with “collection 'vendo_tenant_connectors' is not an engine collection,” so `register`, `list`, `test`, and tenant tools did not work; now the collection is admitted and credentials remain in secrets.

- Bumps `ENGINE_ALLOWLIST_VERSION` 5 → 6 and updates counts to 39 in `packages/core/tests/engine-collections.test.ts`.
- Adds `packages/vendo/tests/tenant-connectors-hosted-store.test.ts` to exercise `hostedStore`/`hostedStoreOps`, assert the row lands in `vendo_tenant_connectors`, and the token is stored via secrets.
- Migration: update the Cloud allowlist mirror to include `vendo_tenant_connectors` at version 6; bump pinned allowlist versions in BYO stores. No API changes; `@vendoai/core` patch release.

<sup>Written for commit e6c9e4300922d189ef0110c4c00a125b2d258170. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1499?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/tenant-connectors-engine-allowlist.md
+++ b/.changeset/tenant-connectors-engine-allowlist.md
@@ -1,0 +1,20 @@
+---
+"@vendoai/core": patch
+---
+
+`vendo_tenant_connectors` joins the engine allowlist. The collection tenant
+connectors write their registrations to (`packages/vendo/src/tenant-connectors.ts`)
+was never added to `ENGINE_COLLECTION_REGISTRY`, so a deployment on a
+Cloud-hosted store — the posture a Cloud host gets by leaving the store slot
+unset — had its first registration refused with `collection
+"vendo_tenant_connectors" is not an engine collection`, which left `register`,
+`list`, `test` and every tenant's own tools dead on a live deployment while the
+suite stayed green: every tenant-connector test composes a local store, and a
+local store has no allowlist in front of it. Exactly the miss the text channel's
+three collections shipped with. The name is added with the `file:line` comment
+each entry in that list carries, `ENGINE_ALLOWLIST_VERSION` moves 5 → 6 as that
+constant's contract requires, and a new seam test drives
+`createTenantConnectors` through `hostedStore`, `hostedStoreOps` and the fake
+console — which serves the same gate as the live door precisely so a fake cannot
+bless a collection production refuses. If your BYO store pins the allowlist
+version, bump it.
--- a/packages/core/src/engine-collections.ts
+++ b/packages/core/src/engine-collections.ts
@@ -3,10 +3,9 @@ import { VendoError } from "./errors.js";
 /** Named in every refusal so a caller can tell "this name was never an engine
     collection" from "this build's list is older than yours". Bump it whenever
     ENGINE_COLLECTIONS or ENGINE_COLLECTION_PATTERNS changes. */
-// 5, not 3 (ours) or 4 (main's): both sides added a collection, so the merged
-// list is a list neither version ever named. A refusal quoting v3 or v4 would be
-// describing a build that never existed.
-export const ENGINE_ALLOWLIST_VERSION = 5;
+// 6: v5's list plus `vendo_tenant_connectors`, which shipped missing from it. A
+// refusal still quoting v5 would be describing a build without that entry.
+export const ENGINE_ALLOWLIST_VERSION = 6;
 
 /** What a collection HOLDS. `knowledge` is the retrieval corpus — documents and
     the chunks an engine mints from them; everything else is `storage`.
@@ -19,7 +18,7 @@ export type CollectionKind = "storage" | "knowledge";
 export interface EngineCollectionSpec {
   kind: CollectionKind;
   /** The fields an `engine.list` watermark may bound, because THIS collection
-      keeps them indexed. Absent for the 37 collections with nothing to walk
+      keeps them indexed. Absent for the 38 collections with nothing to walk
       forward through: an unindexed bound is a full table scan wearing a filter's
       clothes, so it is refused rather than served slowly. */
   indexed?: readonly string[];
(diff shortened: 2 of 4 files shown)


yousefh409:
One correction to the auto-generated summary above: there is no "pinned allowlist version" in a BYO store to bump. `ENGINE_ALLOWLIST_VERSION` is a diagnostic string, not a negotiated protocol field — it appears in exactly one runtime site, interpolated into the refusal text at `packages/core/src/engine-collections.ts:192`, and it never crosses a wire. A BYO store enforces the list itself by calling `assertEngineCollection` (`packages/store/src/ops.ts:412`), so it is fixed by upgrading `@vendoai/core` and nothing else.

The Cloud mirror is a genuine follow-up, but it is NOT a blocker on this PR — see the version-compatibility note in the thread.

PR #1500, opened 2026-08-19
feat(agents,core): an agent remembers one user's own facts, and only theirs

yousefh409:
## What changes

```ts
const support = agent({ name: "support", memory: true, … });
```

The agent gets a per-user memory. Reads are automatic — a capped `[Memory]` block in the per-turn prompt. Writes are deliberate — a visible, guard-bound `remember` tool the model has to call. `memory: true` is the store-backed default; a `MemoryAdapter` is used verbatim for BYO, on the same ladder as every other slot.

Per-user only. Never cross-user.

## Isolation is the whole feature

Memories are generic store rows carrying `refs.subject`, and that subject is the filter on every read. `delete` resolves the id through the subject filter first, so a foreign id finds nothing; `clear` only deletes what the subject-scoped walk returned; the `remember` tool takes its subject from `ctx.principal` and its schema has no field to name anyone else.

An adversarial reviewer who built none of this spent a round trying to break it — 11 hostile subject pairs (`u_4`/`u_42`, `a:b`/`a`, `a\nb`/`a`, case and trailing-space variants, an empty subject, and a JSON-injection payload as a subject) × all five verbs, against a real database. It held. The reason is structural: the store filters with jsonb containment, which is exact value equality, so there is no id-concatenation scheme to collide.

Free bonus, verified: `erase.bySubject` already sweeps rows by `refs`, so a forgotten user's memories are deleted by the existing erase path with no new code.

## Four defects that round found, all fixed here

- **A principal with no `subject` was a master key.** `JSON.stringify` drops an undefined value, so the filter degraded to "match every row" — a `clear` would have wiped every user. Now a principal that cannot be scoped resolves to nothing on reads and refuses loudly on the write door.
- **The cap was not where the promise is.** The spec says a *capped* `[Memory]` block, but the limit lived in the default adapter's write — so a BYO adapter, an explicitly supported rung, could flood the system prompt. The cap moved to the render, where the promise is made. Both constants are now exported so a BYO author can read the numbers they are asked to honour. The comment that claimed "already capped by the caller" no longer lies.
- **A NUL in model-authored text** crashed the write with a raw Postgres error instead of a Vendo validation error.
- **`String.trim()` does not strip U+0085**, so a blank memory rendered an empty bullet. The blank filter now runs the `LINE_TERMINATOR` set this file already defines.

## Forgery-safety

`[Memory]` is a third sibling to `[User]` and `[Situation]` in `packages/core/src/prompt-blocks.ts` — one implementation, because a prompt-injection defence with two copies is a defence that gets fixed once. Memories are labelled as the user's own recorded words, not instructions, and ride the same continuation-indent. Tested against all eight line-terminator forms: none can close the block or forge a `Directions` section.

## The seam

`chat()` was reaching memory only by an undeclared structural pass-through after the rebase onto #1498 — green suite, live feature, nothing written down. `memory` is now declared on `TurnDeps`, and there is a real seam test: one agent, two `chat()` turns on one thread, real store and real guard, asserting on the brief the harness was *actually handed*. Cutting the wire turns it red, so it is load-bearing rather than decorative.

## Gate

`build` · `typecheck` (42/42) · `lint` (dependency-guard OK; portability all legs green) · `packages/agents` 215 · `packages/core` 802 — all EXIT=0.

Stacked on #1498.

---
## Summary by cubic
Adds optional per-user memory. Agents now render a capped [Memory] block per turn and write via a guard-checked `remember` tool, so the model can recall a person’s own facts without prompt forgery. Previously: no memory or tool.

- agent({ memory }): `true` uses the composition’s store-backed default (`storeMemory`); a `MemoryAdapter` is used verbatim; unset means no memory and no `remember` tool.
- Prompt: `[Memory]` renders beside `[User]` and `[Situation]` via `memoryPromptBlock` in `@vendoai/core`. Caps at render with `MEMORY_RECALL_LIMIT` items and `MEMORY_TEXT_MAX_CHARS` per item, so BYO adapters cannot flood. Blank entries drop; shared indent defense prevents forging `Directions`.
- Tool and guard: ships a `remember` tool graded `write`, bound by the guard; unattended runs and `readonly` policy block it. NUL in subject or text is a validation error. A host tool named `remember` now fails boot.
- Isolation: all verbs scope by `principal.subject`. Undefined or NUL subjects read nothing; writes reject. `delete` resolves through the subject filter; `clear` walks only that subject; `erase.bySubject` sweeps memories.
- Wiring and exports: `TurnDeps`/`SessionDeps` declare `memory`; `resolveSystem` recalls per turn for the current principal. Exports `MEMORY_RECALL_LIMIT`, `MEMORY_TEXT_MAX_CHARS`, `storeMemory`, `Memory`, `MemoryAdapter` from `@vendoai/agents`. No schema changes; data is generic records under `vendo_memories`.

- Required action: rename any existing host tool named `remember`.

<sup>Written for commit 413247c5af50cb9d44fd17c5a59c2b795241a225. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1500?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/agents/src/agent.ts
+++ b/packages/agents/src/agent.ts
@@ -35,6 +35,7 @@ import { startChat, type ChatOptions, type Turn } from "./turn.js";
 import { resolveDoor, type DoorConfig } from "./door.js";
 import { withEgress, type EgressConfig } from "./egress.js";
 import { PARKED_TURN_TTL_MS } from "./interruptions.js";
+import { rememberTool, storeMemory, type MemoryAdapter } from "./memory.js";
 import { PERMISSIONS_PATH, schemaReadyPrincipal, type AgentPrincipal } from "./permissions.js";
 import type { SystemPromptHook } from "./prompt.js";
 import {
@@ -70,6 +71,18 @@ export interface AgentConfig {
   store?: VendoStore;
   /** Unset + key → the ladder (E2B key, Cloud pool). */
   sandbox?: SandboxAdapter;
+  /**
+   * What this agent remembers about each of its users, per user and never
+   * across them. `true` → the store-backed default, on this composition's own
+   * store; a {@link MemoryAdapter} is used verbatim (BYO). Unset is no memory at
+   * all: no `[Memory]` block in any prompt, and no `remember` tool for the model
+   * to call.
+   *
+   * Reads are automatic (a capped `[Memory]` block, read as the user's words,
+   * never as instruction); writes are the visible, audited, guard-checked
+   * `remember` tool, scoped to the turn's own principal.
+   */
+  memory?: MemoryAdapter | true;
   /** Where a thinker that runs outside this process dials back to reach your
    *  tools; unset → `VENDO_BASE_URL`. Required by any harness that declares
    *  `requires.toolDoor` — see {@link resolveDoor}. */
@@ -173,6 +186,9 @@ export interface AgentComposition {
   skills: readonly Skill[];
   /** Present only for a harness that thinks on a machine. */
   sandbox?: SandboxAdapter;
+  /** Present exactly when the host asked for memory — the adapter a surface over
+   *  this agent reads and forgets a person's memories through. */
+  memory?: MemoryAdapter;
   /** The seats a harness that does NOT bring its own brain reads (`vendo()`). */
   models?: SeatModels<LanguageModel>;
   instructions?: string;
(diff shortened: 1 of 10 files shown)


yousefh409:

PR #1503, opened 2026-08-19
feat(agents,core,guard): a turn waiting on a yes can be answered a week later, from anywhere

yousefh409:
## What changes

```ts
await user.turns.list({ status: "interrupted" });               // findable after ANY restart
await user.turns.resume(turnId, decisions, { headers, context }?);
```

A turn that parked on an approval is now a durable thing you can find and answer later — from a different request, a different process, or a week later. Previously the advice was "show the cards and call `run()` again", which re-ran the whole turn.

## No new persistence

The guard was already writing the exact `ToolCall` and its context when it parked a call. Resume reads that back and re-dispatches it byte-for-byte; the guard's existing one-time receipt makes it run **exactly once**. Re-*running* the turn cannot work — a fresh model turn mints a new tool-call id and the approval no longer matches — which is why this re-dispatches instead.

The one thing missing was an address: nothing linked a `turnId` to a parked approval. That is 3 lines — the guard now writes `ctx.turnId` onto the row it was already writing. No new collection, no new write.

## Proven across a real restart

`interruptions.test.ts:109-158`: a real store on disk, agent A parks a destructive refund, **the store is closed**, and a second store and a second `agent()` are built over the same directory — new guard, new harness, new tool closure, new counter. The new composition lists the turn with its original id, resumes it, and the refund runs there and only there. A retried resume gives `conflict` with the counter still at 1.

## What the adversarial round found, and what is fixed here

Exactly-once **held** under 8 concurrent resumes, two independent compositions racing, and a 3-way race on a two-interruption turn. Cross-user ownership **held** against seven hostile subject spellings. Three defects did not:

- **A yes given to one agent dispatched a different agent's implementation.** Two agents over one store shared the approvals collection, and two same-named tools hash identically — so a person's approval for `ops`' refund ran `support`'s refund. Resuming through an agent that lacked the tool was worse: it marked the approval *approved*, failed to dispatch, and burned the one-shot yes forever. Fixed with an agent axis (`RunContext.agent`, carried onto the parked row exactly as `turnId` is), read by both `list` and `resume`, failing closed and refusing **before** anything is decided. It is the agent's *name*, not an instance id — a per-instance id would have broken cross-process resume, which is the whole point of the slice.
- **The seven-day expiry was enforced on one of the two resume faces.** `turns.resume` refused an expired ask while the in-process `result.resume()` approved and ran it.
- **An answer-shaped decision given for an approval was silently a denial** — permanently rejecting an ask on an answer the person never gave.

The last two collapsed into one pre-flight in the single door both faces pass through, which let the duplicate expiry logic be **deleted** rather than copied.

## The seven days

Founder decision: parked interruptions get a 7-day TTL instead of the guard's 60-minute default, and a resume past it fails naming what happened and how it ends. The default goes on the guard `agent()` composes for itself; a passed-in guard instance still wins verbatim, and `createVendo` is untouched at 60 minutes. Enforced at read time — nothing in this package sweeps, so read time is where the promise can actually be kept.

## One correction to the brief, pinned by a test

A user typing another message does **not** destroy a pending interruption. Abandonment only flips transcript parts left in `approval-requested`, and this lane never leaves one. That is the streaming lane's behaviour. So the ask survives, stays listed, and stays resumable — better than the spec promised, and now tested.

## Gate

`build` · `typecheck` (42/42) · `lint` (dependency-guard OK; portability all legs green) · `packages/agents` 193 · `packages/guard` 362 · `packages/core` 800 · `packages/harnesses` 568 — all EXIT=0. Adversarial suite 18/18.

Stacked on #1498.

---
## Summary by cubic
Resume interrupted agent turns by id from any process within seven days, without re-running the turn. Previously approvals required calling run() again (minted a new tool-call id that no longer matched); now we re-dispatch the parked call exactly once.

- Adds `createTurns` with `turns.list({ status: "interrupted" })` and `turns.resume(turnId, decisions, { headers, context })` in `@vendoai/agents`. Exactly-once is enforced by the guard’s one-time receipt.
- Stores `turnId` and `agent` on the existing approval row; `list`/`resume` filter by agent. A cross-agent `turnId` now returns not-found, and resuming through an agent lacking the tool refuses without spending the approval.
- Default TTL is seven days via `agent()`; lists omit expired asks. `resume` applies the same deadline: it proceeds with the live asks only and requires decisions for those; it returns a conflict only when every ask is expired.
- Concurrent turns on the same thread no longer collect each other’s approvals; each turn only resumes the asks it parked.
- Validation and ownership: `resume` requires a verdict (“approve”/“deny”) for each parked id; non-verdict shapes are rejected. Foreign/unknown turns return not-found; already-answered turns return conflict. Resumed calls run with the resuming call’s headers/context, never the parked request’s.

**Rollout/Migration**
- Agents created via `agent()` now default to `approvals.parkedCallTtlMs = 7 days`. To keep 60 minutes, set `guard: { approvals: { parkedCallTtlMs: 60*60_000 } }` or pass a guard instance.
- When calling `turns.resume`, pass a verdict for every listed interruption id. Do not include expired ids; partial or malformed maps are refused without side effects.
- Multi-agent hosts over one store should keep distinct agent names; turns are scoped by name.

<sup>Written for commit 82e52871921641032a23a1e7f0eefe66f08e9759. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1503?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/agents/src/agent.ts
+++ b/packages/agents/src/agent.ts
@@ -34,6 +34,7 @@ import { startRun, type RunOptions } from "./away.js";
 import { startChat, type ChatOptions, type Turn } from "./turn.js";
 import { resolveDoor, type DoorConfig } from "./door.js";
 import { withEgress, type EgressConfig } from "./egress.js";
+import { PARKED_TURN_TTL_MS } from "./interruptions.js";
 import { PERMISSIONS_PATH, schemaReadyPrincipal, type AgentPrincipal } from "./permissions.js";
 import type { SystemPromptHook } from "./prompt.js";
 import {
@@ -319,7 +320,15 @@ export function agent(config: AgentConfig): VendoAgent {
   // completed with this composition's store.
   const guard = isGuardInstance(config.guard)
     ? config.guard
-    : createGuard({ store, ...config.guard });
+    : createGuard({
+      store,
+      ...config.guard,
+      // An agent's parked turn waits for a PERSON, not for a BYO loop that
+      // will retry in the hour the guard defaults to (interruptions.ts). The
+      // host's own number still wins — this fills the slot, it does not take
+      // it — and a guard INSTANCE passed in is taken verbatim, TTL included.
+      approvals: { parkedCallTtlMs: PARKED_TURN_TTL_MS, ...config.guard?.approvals },
+    });
   const tools = mergeSources(config.tools ?? [], config.mcp ?? []);
   const bound = guard.bind(tools);
   const skills: Skill[] = loadSkillFolders(config.skills);
--- a/packages/agents/src/agent.ts
+++ b/packages/agents/src/agent.ts
@@ -160,6 +160,10 @@ export interface VendoAgent {
  * object stays exactly `{ name, session }`.
  */
 export interface AgentComposition {
+  /** WHICH agent this composed — `agent({ name })`. Declared here because a
+   *  consumer that scopes rows to one agent (`createTurns`) reads it off the
+   *  composition, never off a name a caller passes in beside it. */
+  agent: string;
   harness: Harness<unknown>;
   store: VendoStore;
   files: FilesAdapter;
@@ -379,6 +383,7 @@ export function agent(config: AgentConfig): VendoAgent {
 
   const deps = {
     name: config.name,
+    agent: config.name,
     harness,
     store,
     files,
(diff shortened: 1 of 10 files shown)


yousefh409:

PR #1504, opened 2026-08-19
feat(agents,ui,vendo): a standalone agent is one mount, and the page it answers is stock useChat

yousefh409:
## What changes

```ts
const handle = support.handler({ basePath: "/api/agent", resolveUser });
// Next:  export { handle as GET, handle as POST, handle as DELETE }
// Hono:  app.all("/api/agent/*", (c) => handle(c.req.raw))
```
```tsx
const chat = useVendoChat({ api: "/api/agent", threadId, onThreadId });
chat.interruptions;                 // pending, durable across reload
await chat.resume({ [id]: "approve" });
```

One mount and one hook. Before this, a host wiring `agent()` hand-wrote three routes and the chat wire from the docs.

## One router, not two

`@vendoai/vendo` had a real HTTP route runtime; `@vendoai/agents` had none — so the moment agents grew a chat route it would have copied ~80 lines of `server.ts`. The generic half moves down into `@vendoai/agents/http` (matching, dispatch, error→status, body parsing, mount stripping) and `wire/shared.ts` becomes a re-export shim, so **all ~40 existing importers keep their specifier**.

That is the proof this was behaviour-preserving: **2,582 `packages/vendo` tests pass with zero test files edited**, and the whole extraction is **+51 net source lines**.

## The seam test found a dead feature

`CLAUDE.md` asks for a test where neither side mocks the other, because this repo has shipped a dead feature four times that way. It happened again here, and the test caught it:

**The page rendered a correct consent card, the Approve click looked like it landed, and the refund never ran.** The turn sat on its waiter for 90 seconds and gave up with "the approval request expired unanswered". Both unit suites were green throughout — each side was right about its own half.

Root cause: a parked call waits on a **guard** decision, and only `guard.approvals.decide` resolves that waiter. The hook was answering through the ai-SDK's *local* approval channel, which repaints the page and tells the server nothing. Fixed: `resume()` posts to the permission wire, and the mount serves that wire under its own `basePath` behind its own `resolveUser` — a client that only knows `api` cannot reach a plane pinned elsewhere, and a standalone host configures identity once.

Two tests written earlier in this PR were then corrected, stated out loud, because the browser proved them wrong. A real browser outranks a unit test that agreed with the bug.

## What the browser test actually exercises

Real Chromium → the shipped `useVendoChat` → real HTTP → the real `handler()` → a real `agent()` with a real guard policy and a real store. Only the *thinker* is scripted. It asserts the refund ran **from the server's own record**, not from text on screen. A second test clears `localStorage`/`sessionStorage` and reloads: transcript and pending approval both come back from the server, which is the no-hidden-storage claim.

It lives in `fixtures/integration-browser` because `packages/ui` may only depend on core — so a test inside `packages/ui` physically cannot import the real counterparty, which is exactly how this class of bug survives.

## Two spec options deliberately absent

`publicOrigin` would be dead on arrival — the door's origin is fixed at `agent()` time and already throws the exact boot error the spec describes. `mcp` governs a public MCP plane that does not exist in this package, so absent and default-off are the same behaviour. Both are named in a comment on `HandlerOptions` so they read as considered, not forgotten; both are additive later without breaking anyone.

`POST /turns/:id/resume` answers 501 naming the seam — #1503 owns those rules, and inventing a second set is how one rule becomes two.

## Gate

`build` · `typecheck` · `lint` (dependency-guard OK; portability all legs green) · `packages/agents` 203 · `packages/ui` 1384 · **browser seam 2/2** — all EXIT=0.

Stacked on #1498.

---
## Summary by cubic
Makes a standalone agent one HTTP mount and ships a thin chat hook that speaks to it, so hosts stop hand-wiring routes and approvals actually unblock parked turns.

- Adds `agent.handler({ basePath, resolveUser, headers? })` and `agentHandler(agent, options)`. Routes: POST `/threads` streams a turn and returns `x-vendo-thread-id`; GET `/threads` lists; GET/DELETE `/threads/:id` read/delete; `POST /turns/:turnId/resume` is wired but returns 501 to name the durable-resume seam.
- Serves approvals/grants under this mount and behind its `resolveUser`; unauthenticated requests return 401. The MCP door stays on its absolute path and never challenges identity.
- Extracts the shared HTTP route runtime to `@vendoai/agents/http` and re-exports it in the umbrella so both mounts share paths and envelopes; existing `@vendoai/vendo` importers keep their `./wire/shared` specifier.
- Introduces `useVendoChat` in `@vendoai/ui`: stock `useChat` transport with no client storage. It round-trips the thread id, reloads the transcript (including pending approvals) from the server, and `resume()` posts decisions to this mount’s permission wire. Refused decisions now bubble (e.g. 409), thread switches update the live id, and the hook is re-exported from `@vendoai/vendo`.

- Minimal adoption
  - Mount once: create `const handle = support.handler({ basePath: "/api/agent", resolveUser })` and export it as GET/POST/DELETE (or `app.all("/api/agent/*", c => handle(c.req.raw))`). Ensure your server also routes the door’s absolute path if the engine thinks outside the process.
  - In the client, replace hand wiring with `useVendoChat({ api: "/api/agent", threadId?, onThreadId? })` and render `interruptions` with a call to `resume({ [id]: "approve" | "deny" })`.
  - If a standalone client previously pointed approvals at `/api/vendo`, use this mount’s `/approvals/*` instead. Existing umbrella consumers do not change.

<sup>Written for commit cdb7e12bcb3daccb04554f1c56dd907f0dd509f8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1504?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/fixtures/integration-browser/e2e/agent-chat.spec.ts
+++ b/fixtures/integration-browser/e2e/agent-chat.spec.ts
@@ -1,0 +1,84 @@
+/**
+ * THE SEAM: the shipped `useVendoChat` in a real browser against a real
+ * `agentHandler()` over real HTTP, with no stub on either side.
+ *
+ * One script, one run, the whole flow — a turn that parks on a real guard
+ * approval, a person answering it in the page, and the parked call actually
+ * running. What makes it a seam test rather than a demo is the last assertion:
+ * the refund is read back from the SERVER's own record, so a page that merely
+ * rendered a convincing sentence cannot pass.
+ */
+import { expect, test } from "@playwright/test";
+
+test("a parked approval is answered in the page, and the parked call then runs", async ({ page, request }) => {
+  await page.goto("/agent.html");
+  await expect(page.getByTestId("composer")).toBeVisible();
+
+  // Nothing has run yet: the mount's own record is the baseline.
+  expect((await (await request.get("/__agent/refunds")).json() as { refunded: string[] }).refunded).toEqual([]);
+
+  await page.getByTestId("composer").fill("refund invoice 7");
+  await page.getByTestId("send").click();
+
+  // 1. The turn reached the agent and came back with the conversation's id on
+  //    the response header — the round trip, not a rendered guess.
+  await expect(page.getByTestId("thread-id")).toHaveText(/^thr_/);
+
+  // 2. It PARKED. The guard's ask crossed the stream as a real
+  //    `approval-requested` part and the hook surfaced it as an interruption,
+  //    naming the tool the agent is waiting to run.
+  await expect(page.getByTestId("interruption")).toHaveCount(1);
+  await expect(page.getByTestId("interruption-tool")).toHaveText("host_refund");
+  await page.screenshot({ path: "e2e/artifacts/agent-chat-parked.png", fullPage: true });
+  // Still nothing has run — the park is real, not cosmetic.
+  expect((await (await request.get("/__agent/refunds")).json() as { refunded: string[] }).refunded).toEqual([]);
+
+  // 3. The person answers, in the page, through the hook's own `resume`.
+  await page.getByTestId("approve").click();
+
+  // 4. The parked call RAN — server-side evidence, not screen text.
+  await expect(async () => {
+    const { refunded } = await (await request.get("/__agent/refunds")).json() as { refunded: string[] };
+    expect(refunded).toEqual(["inv_7"]);
+  }).toPass({ timeout: 20_000 });
+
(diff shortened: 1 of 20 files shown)


yousefh409:

PR #1508, opened 2026-08-19
feat(agents): one call binds the person, and every surface below it is theirs

yousefh409:
## What changes

```ts
const user = support.forUser("u_42", {
  profile: { name: "Dana", plan: "pro" },  // model-visible facts — durable
  context: { tenantId },                   // guard/tool context — durable
});

user.chat(msg, { headers: req.headers });  // request authority rides PER CALL, never bound

await user.turns.list({ status: "interrupted" });
await user.threads.list();  (await user.threads.get(id)).messages();
await user.memories.list();
```

One call binds the person; `turns`, `threads` and `memories` are all scoped to them from then on. A facade — it binds a subject and delegates; the behaviour lives in the slices below.

## The distinction that matters, and how it is proven

`profile` and `context` are **durable** — bound once, carried by every call. `headers` are **request-lifetime** — they ride per call and are never bound, because one request's authority must not leak into another's.

The test binds one facade and makes three real turns with different auth headers, then reads everything back off the context a **real tool** was handed and the brief the **real harness** received — never off the options object. Subject and context identical across all three; headers `Bearer one`, `Bearer two`, then absent. Cross-user isolation is a second facade over the same real store: a foreign `threads.get` is `not-found`, a foreign `turns.resume` is `not-found` with the refund counter still at 0, and a foreign `memories.delete`/`clear` removes nothing.

## Two decisions worth a reviewer's eye

**`memories.*` throws when the agent has no memory configured**, rather than answering with an empty list. "Nothing remembered" and "nothing is ever remembered" are different facts, and a settings screen that renders the first while the second is true is lying to the person; a `delete` reporting success having removed nothing is worse. It is host-code error with a one-line fix, so the error says so.

**The facade's durable `context` DOES apply to `turns.resume`.** PR3's rule is that authority comes from the resuming call and is never replayed from the dead request — and the facade's standing context is the live caller's, not the dead request's. Leaving it out would mean a guard rule keyed on `tenantId` evaluates differently for a resumed turn than for the turn that parked it, with nothing in the API explaining why. Bound context underneath, per-call on top, exactly as `chat()` merges them; `headers` still never bind. `interruptions.ts` is untouched.

`turns`/`threads`/`memories` each latch `store.ensureSchema()` in front, for the reason `permissions.ts:18-25` already documents: an approvals inbox or a memories screen can legitimately be the first thing a deployment ever calls.

## Deprecations — path, not removal

`RunOptions.as` → `forUser`. `session()` and `AgentSession` → `forUser`/threads (the thread is the durable noun). JSDoc `@deprecated`, the convention already in this repo. Both still work. **`respond()` is NOT deprecated** — and `SessionOptions` is deliberately untouched, since `RespondOptions extends` it and deprecating the base would have tainted a live verb.

## Gate

`build` · `typecheck` (42/42) · `lint` · `packages/agents` **250 tests** — all EXIT=0.

Stacked on #1500.

---
## Summary by cubic
Binds a person once with `agent.forUser(subject, { profile, context })` and scopes chat, turns, threads, and memories to them. Before, callers re-sent `as/user/context` per call and resumes dropped context; now durable `profile/context` ride every call (including `turns.resume`), while `headers` remain per-call and are never stored. `memories.*` throws when memory is disabled to avoid misreporting “nothing remembered.”

- Adds `packages/agents/src/facade.ts` and exposes `AgentUser` and related types from `index.ts`. The facade delegates to existing implementations (`startChat`, `createTurns`, store thread/message helpers, memory adapter).
- `turns.resume` merges the facade’s bound `context` with any per-call `context`; headers never bind. Threads are user-owned (`get` on foreign is `not-found`; `delete` is idempotent).
- Facade ensures and latches schema on first read so first-call reads (threads/turns) work on a fresh store.
- `RunOptions.as` and `agent.session()` are deprecated but still work; `respond()` is unchanged.

**Migration**
- Use `agent.forUser(subject, { profile, context })` instead of `RunOptions.as` and `agent.session()`; stop passing `as/user/context` per call. Continue passing `headers` per call.
- If exposing memories, enable memory via `agent({ memory: true })` or provide a `MemoryAdapter`; handle the validation error otherwise.

<sup>Written for commit f86c459141da6122a200b912d81b17f2a416e61e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1508?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/agents/src/agent.ts
+++ b/packages/agents/src/agent.ts
@@ -34,6 +34,7 @@ import { startRun, type RunOptions } from "./away.js";
 import { startChat, type ChatOptions, type Turn } from "./turn.js";
 import { resolveDoor, type DoorConfig } from "./door.js";
 import { withEgress, type EgressConfig } from "./egress.js";
+import { createUser, type AgentUser, type UserOptions } from "./facade.js";
 import { PARKED_TURN_TTL_MS } from "./interruptions.js";
 import { rememberTool, storeMemory, type MemoryAdapter } from "./memory.js";
 import { PERMISSIONS_PATH, schemaReadyPrincipal, type AgentPrincipal } from "./permissions.js";
@@ -117,6 +118,21 @@ export interface VendoAgent {
    */
   chat(message: string, options?: ChatOptions): Turn;
   /**
+   * Everything this agent does FOR ONE PERSON, with who they are bound once —
+   * their turns, their conversations, and what it remembers about them.
+   *
+   *     const user = support.forUser("u_42", {
+   *       profile: { name: "Dana", plan: "pro" },
+   *       context: { tenantId },
+   *     });
+   *     await user.chat("where is my refund?", { headers: request.headers });
+   *
+   * `profile` and `context` are FACTS about the person and are bound here.
+   * `headers` are the authority of ONE request, so they ride per call and are
+   * never kept — see facade.ts.
+   */
+  forUser(subject: string, options?: UserOptions): AgentUser;
+  /**
    * `respond` answers a person over HTTP: one turn, an AI-SDK UI-message-stream
    * `Response` to return from your route, with the conversation's id on
    * `x-vendo-thread-id`.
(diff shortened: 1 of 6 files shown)


yousefh409:

PR #1511, opened 2026-08-19
feat(vendo,docs): one canonical agent prompt, one playbook — docs own every step

yousefh409:
> **Sequencing — merges right after #1530.** The prompt drops the `vendo doctor --json` done-gate and the playbook is written against #1530's init (facts + one Continue link, no pastes, no doctor). Landing this before #1530 publishes a flow the CLI does not yet have.

Rebased onto main via merge (additive, no rebase) after #1507 merged. Realigned to #1509's direction — the standalone backend agent is the front door and is deliberately CLI-free.

## What changed

| # | Item | Status |
| --- | --- | --- |
| 1 | Canonical prompt module `packages/vendo/src/agent-prompt.ts` — `buildAgentPrompt({ src, signedIn })`, re-exported from the package root | done |
| 2 | Live copies swapped: the card's twin copies at `src=docs`, the README block at `src=readme`; four dead prompt variants deleted from `agent-prompt.jsx` | done |
| 3 | Three docs-rot tripwires in `packages/vendo/tests/agent-prompt-docs.docs.test.ts` | done |
| 4 | `docs-site/agents/index.mdx` rewritten: one shared start, four path sections, init's Continue link as the router | done, realigned to #1530 + #1509 |
| 5 | Front door: "about five minutes" dropped; `give-it-to-your-agent.mdx`'s duplicated prose replaced by the card + one playbook link | done — **fourth card dropped, see below** |
| 6 | Quickstart arrival cues for readers arriving from init | done — **backend cue dropped, see below** |

### 1 — the canonical prompt

```
Install Vendo in this repo. Read https://vendo.run/agents.md and follow it exactly. Relay Vendo's setup questions to me and use my answers. Ask me before creating any account or key — Vendo Cloud is the recommended option. You're done when my app runs and the agent answers from my own API. Then ask me whether I'd like to star it via https://vendo.run/star?src=<surface> to support the project.
```

`signedIn: true` swaps the ask-first sentence for the console's existing wording, matching `vendo-web`'s `lib/prompts.ts`. Only `buildAgentPrompt` is exported; `signedIn: true` has no in-repo caller yet (it exists for the console).

### 3 — the tripwires

- **the paste** — the card's two copies and the README's block each match `buildAgentPrompt` under their own `src`. The clipboard string is byte-exact; the two *rendered* copies compare whitespace-collapsed, because JSX collapses a wrapped text child anyway and rewrapping a README fence is not drift.
- **the Next externals** — the array in `index.mdx`, `agents/index.mdx`, `existing-agent/ai-sdk.mdx` and `production/troubleshooting/e-cfg-004.mdx` compared to `NEXT_SERVER_EXTERNALS`. The troubleshooting page is in the list because it's where a reader lands *when the list is wrong*.
- **the `lib/vendo.ts` contract** — kills the "phantom `resolvePrincipal` import" class. Builds init's real scaffold with `compositionModuleSource(...)` in both shapes, collects what the quickstart shows the reader adding by hand, and asserts every name the existing-agent pages import is in that union — plus that the union is genuine, so a slice matching nothing can't pass forever.

All three proven red by mutation and green on restore, twice (once per cut).

## Realignment to main — what changed vs my original items

| Area | Before | Now, and why |
| --- | --- | --- |
| `docs-site/backend/quickstart.mdx` | my "Already ran `vendo init`?" arrival cue | **Conflict — took main whole.** That page now opens *"One package, no CLI, no `init`"*; a cue about init contradicted its own first paragraph. File is byte-identical to main. |
| Fourth "In your backend" direction card + `direction-backend.svg` | added | **Dropped.** main's `index.mdx` already features the standalone path in "Just want an agent?" *above* the init flow, and the three cards answer init's three use cases. A backend card was redundant and wrong. Grid back to `cols={3}`; SVG deleted. `index.mdx` is now main's minus the time promise. |
| Playbook backend section | `vendo init --use-case agent-loop`, `respond()`, `refs.approvals`, "call `run()` again" | **CLI-free rewrite.** No init, no Continue link, one package; `chat()` / `handler()` / `run()`; the `TurnResult` story — `ok` carries `output`, `interrupted` carries `interruptions` + `resume()`, and `resume()` is not a rerun. `respond()` and `refs.approvals` died in #1498/#1509. |
| Continue-link table | four rows | **Three.** `/backend/quickstart` removed — backend takes no init, and #1530's own item 2 flags its `backend` use case as pending a product call for exactly this reason. Two new pointers send a backend-only reader straight past the install. |
| Step 4 | "prints ONE Continue link" | Now shows #1530's real tail verbatim — `Wired:` / `Detected:` / `Guard:` / `Continue:` — plus the receipt fields (`wrote`, `detected`, `guardPosture`, `continueUrl`), and states that nothing in it is a paste. |
| `VENDO_BASE_URL` | "one of the two paths" | One path. main's backend pages no longer mention it at all. |
| Consent gates | absent | Added the two #1530 keeps: AI-spend consent ahead of the grading pass, and loosening proposals queued as **pending** rather than applied ("no loosening without a human"). |
| README "Which one are you?" | 3 paths | 4, with the backend line using `agent()`/`chat()` and saying "no CLI". |

## Second pass — nine claims the source contradicted

An independent review of the first cut caught facts inherited verbatim from the old playbook. Each re-verified against source before the fix (commit `c4c36ca`):

| Claim as published | What source says |
| --- | --- |
| `oauth` is ignored when `auth` is set | `createVendo` **throws** naming the mixed keys — `compose-config.ts:120-130`. The page said both. |
| pass `--ai` and a sync that cannot ask reports `judgment: skipped`; `init --agent` rejects the pair | Backwards and stale. The note fires when `ai !== true` (`sync-flow.ts:364-366`); neither flag is rejected and agent mode grades by default (`cli.ts:185-188`, since #1484). |
| "init errors rather than guessing" on undetected frameworks | `guardUnattendedRun` returns early for `--agent` (`init.ts:1801`); `resolveFramework` falls through to `"custom"` (`:344-349`) — silently, in the exact mode this page prescribes. |
| the cautious preset tightens the posture | It **replaces** `.vendo/policy.json`'s rules: inline wins with no merge (`policy.ts:130-133`). |
| a parked door call returns `pending-approval` with where to resolve it | An in-band prose error; `pending-approval` is internal and the door serves no approval UI (`door.ts:1679-1682`). |
| Express **and custom** hosts use `mountVendo()` | Generated into the Express scaffold only, and terminal (`init-scaffolds.ts:559`); custom uses `handleVendoRequest`. |
| `not-implemented` over the door is an `actAs` problem | Server-action-bound tools answer `not-implemented` over MCP regardless (`registry.ts:380-386`). |
| pass components to the provider so screens are built from them | Generation reads the **server** catalog; the provider's `components` prop is the render half (`compose-surfaces.ts:216-221`). |
| `respond()`/`run()` are peers; `refs.approvals` means nothing happened | Both methods on what `agent()` returns; parking is per call. *(Superseded again by #1509 — now `chat()` and `TurnResult`.)* |

Plus: version floor `0.28.0` → `0.29.0`, and the existing-agent cue no longer implies init installed `@ai-sdk/react` (it installs `ai` and the provider only — `provider-deps.ts:22`).

## Spec-changed tests

None. No existing test asserted behavior this change removes; the six docs-rot suites pass unchanged, including their whole-tree link check against the rewritten playbook.

## Noticed, not fixed — outside this slice's fence

1. **`packages/vendo/skills/vendo-setup/SKILL.md` still gates on doctor** (`:3`, `:148`) and claims `judgment` is always `delegated` (`:43-44`). Init writes it into the user's repo and the playbook advertises it, so it is the one surface an installing agent actually loads. Belongs to #1530's slice.
2. **The docs-rot gates can replay a cached green.** They live in `@vendoai/vendo` but read `docs-site/**`, outside the package, and turbo's `test` task declares no `inputs` — a docs-only edit doesn't invalidate the cache, and `test:affected` may scope the package out entirely. Pre-existing, but it blunts what this PR buys. Cheapest fix: add `docs-site/**` and `README.md` to the `test` task's `inputs`.
3. `docs-site/reference/handler-options.mdx:83` describes `cautious` without **ungraded**; `:85` says a missing policy file "auto-runs with no notice" — `guard.ts:1680-1684` still asks for destructive and ungraded.
4. `docs-site/production/troubleshooting/e-wire-001.mdx` / `e-wire-002.mdx` fail to compile in `mint dev`. Pre-existing.

## Gate

On the merged commit: `pnpm build` 22/22 · `pnpm typecheck` 44/44 · `pnpm lint` 4/4 (`demo-bank` carries 4 pre-existing warnings, 0 errors) · `pnpm test:affected` · six docs-rot suites run directly, 2-worker capped, **69 passed**.


yousefh409:
### Greptile P1 — resolved, and it was a real bug

**Finding:** `docs-site/backend/quickstart.mdx` told standalone `@vendoai/agents` readers to answer *"Through my own agent loop (AI SDK / Mastra)"*, which makes init emit `vendoTools`/`streamText` wiring for a pre-existing chat loop — unrelated to the `agent()` runtime shown directly below it. Correct call, and the T-Rex fixture run proves it.

**Where it stood:** the finding is against `c4c36ca57`, before this branch merged main. On `main`, #1509 rewrote that page to be deliberately CLI-free — *"One package, no CLI, no `init`, no files moved around"* — which deletes the `agent-loop` instruction and the whole install/login preamble the finding is about.

**What I did:** the merge hit a conflict on exactly that file and I resolved it to main's version whole, which also deleted my own `Already ran vendo init?` arrival cue there — that cue contradicted the page's new first paragraph, so it had to go rather than be reworded.

Verified on `dfca7f127`:

```
$ grep -nE "agent-loop|my own agent loop|vendo login|streamText|vendoTools" docs-site/backend/quickstart.mdx
(no matches)
```

The only surviving `vendo init` mention on that page is an optional aside for readers who already have an HTTP API (`api()` tool extraction), which is a different claim and is accurate.

**Same root cause, fixed one level up:** the finding is that the docs conflated two integration contracts. The playbook (`docs-site/agents/index.mdx`) had the identical conflation — its backend section claimed init wrote that path with `--use-case agent-loop`. That section is now CLI-free too, the Continue-link table dropped its `/backend/quickstart` row, and two pointers send a backend-only reader straight past the install. #1530's own item 2 flags its `backend` use case as pending a product call for this same reason, so I did not encode it.

PR #80, opened 2026-07-06
follow-up: sandbox route channel (usePathname/useSearchParams/useParams return the real host route)

yousefh409:
## What

A read-only `__vendoRouteData` channel (parallel to the existing `__vendoAnchorData`) so generated sandbox components' `next/navigation` shims return the **host's real route** instead of empty values.

### Plumbing (mirrors the proven anchorData path)
- `vendo-stage/stage-host.ts` — `StageRoute` type (`{ pathname; search; params? }`, params `string | string[]` for catch-alls); optional `route` on init + update payloads; `update()` now forwards it.
- `vendo-stage/runtime.ts` — injects `window.__vendoRouteData` on init, re-patches on `ui/update`.
- `vendo-sandbox-shims/next-navigation.ts` — `usePathname`/`useSearchParams`/`useParams` read the channel (safe empty fallback when absent).
- `vendo-react/stage-adapter.tsx` — `route` prop threaded into initialize + a route-change update effect.
- **Packaged `@vendoai/next` client** — `SandboxStage` → new `RoutedStage` sources live `next/navigation` under a `<Suspense>` boundary and feeds `route` to `VendoStage`. Real installs get populated route data, not just demo-bank.

### Also fixed (Codex review)
- Catch-all params preserved (no more `join("/")` shape loss).
- Real `useSearchParams` subscription (was an unsubscribed `location.search` read that went stale on query-only nav).
- **Pre-existing bug:** `stage-host.ts` `update()` silently dropped `anchorData` from the `ui/update` payload — the live anchor re-patch (saved-view refresh on same-structure data) never reached the controller. Now forwarded + pinned by a test.

## Architecture flag for you (your call at merge)
Auto-supplying route from the packaged client required adding **`next` as a peerDependency of `@vendoai/next`**, reversing the package's previously deliberate next-free stance (the old `onNavigate` prop pattern). For the Next adapter this is defensible and gives "it just works," but it's a real coupling decision. Alternative: thread `route` as a host-supplied prop through `VendoRoot` (keeps the library next-free, but the host must wire route themselves). I went with auto-supply per the "make it work for packaged installs" direction — happy to switch to the prop approach if you prefer. **Note:** if PR #68 (`@vendoai/next` → `vendo` umbrella) lands, this needs reconciling there.

[...]

PR #1551, opened 2026-08-20
docs: cut what CLAUDE.md can already read, move the browser loop to a skill

YSharara1:
## What

Two cuts and one move in the root `CLAUDE.md`, which every session in this repo loads in full.

**Cut — the codebase already says it:**
- The whole `## Commands` section. `pnpm install` / `build` / `test` / `test:affected` / `typecheck` / `lint` and `pnpm --filter demo-bank dev` are `package.json`'s `scripts` read back. The one non-obvious thing that line carried — `test:affected` is the local default, never run the full suite locally — is already stated under `## Tests`.
- Three of the four `## Layout` bullets (`examples/`, `corpus/`, `docs-site/`). That is `ls` output.

The `packages/` bullet **stays**: "the behavior contract is the exported types/zod schemas and the test suites — there are no prose contract docs" is a gotcha a new session would get wrong, and it names the enforcement (`scripts/dependency-guard.mjs`).

**Moved to lazy loading:** the browser-spec workflow (the warm `VENDO_HARNESS_DEV=1` harness while editing, the one fresh production build that is the proof of record, and "script the flow, don't step through it click-by-click") now lives in `.claude/skills/browser-specs/SKILL.md`. `CLAUDE.md` keeps a one-line pointer. Only the skill's description stays resident; the body loads when someone is actually writing a browser spec.

8,301 → 6,851 chars.

## Why

Every line of the root `CLAUDE.md` is in context on every session, whatever the task. A line a session could reconstruct with `ls` or by reading `package.json` is paid for on every one of those sessions and buys nothing — the model would have gone and read the manifest anyway.

The browser-spec loop is different: it is not derivable and it is worth keeping, but it only matters while writing a browser spec. A skill is exactly that shape — resident description, lazy body.

**Nothing under `## Rules` moved.** A prohibition that lives in a lazily-loaded skill is a prohibition that may not be loaded at the moment it matters, so those stay where they are regardless of size.

## Verification

- `CLAUDE.md` 8,301 → 6,851 chars; no section other than `## Commands` and the three `## Layout` bullets is touched — see the diff.
- No code, config, or build input changes: the diff is 2 files, both documentation.
- The skill's frontmatter carries `name` and `description`, so it appears in the skill listing and is model-invocable.

The one judgment call worth a reviewer's eye: whether the `## Layout` bullets earn their place for orientation even though `ls` covers them. I cut them; easy to argue back.

https://claude.ai/code/session_01Eb3f3SpvVx9pHvxhhA5u5s

### DIFF
--- a/.claude/skills/browser-specs/SKILL.md
+++ b/.claude/skills/browser-specs/SKILL.md
@@ -1,0 +1,34 @@
+---
+name: browser-specs
+description: How to iterate on and prove a Playwright browser spec in this repo — the two-speed warm-harness loop for editing, the fresh production build that is the only result worth reporting, and the rule that an agent scripts a browser flow rather than stepping through it click-by-click. Use when writing, debugging, or verifying any browser/Playwright spec, or when a UI-affecting change needs real-browser proof.
+---
+
+# Browser specs
+
+## Iterating is two-speed
+
+While you are editing, start ONE warm hot-reloading harness (`VENDO_HARNESS_DEV=1`
+plus a pinned `VENDO_HARNESS_PORT`) and rerun specs against it with the same two
+variables — reuse is dev-mode-only, so every rerun skips the build and starts in
+seconds.
+
+Then one run WITHOUT the flag — a fresh production build — is the proof of
+record, and the only result worth reporting. A production run never reuses a
+server: `vite preview` serves whatever was built last, so a reused one greens the
+previous build.
+
+## Script the flow, don't step through it
+
+An agent proves a browser flow by SCRIPTING it, never by stepping through it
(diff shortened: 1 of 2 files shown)


YSharara1:
Superseded by [runvendo/vendo-cloud#80](https://github.com/runvendo/vendo-cloud/pull/80).

This repo is now a one-way copybara projection of `runvendo/vendo-cloud` — its `main` is written only by `vendo-release-bot`, so a PR here cannot land. The change will appear here automatically once #80 merges.

Not a straight cherry-pick: the fold (#71, S11d) moved several of the files this touched, and every bug was re-verified as still live on current main before re-cutting. Details on #80.

PR #1563, opened 2026-08-20
fix(core,apps,vendo): a remix's wish list records what landed, and a follow-up edit changes the port

yousefh409:
A cold walk on `main` found one user request leaving **four** entries on a remix's `seed.wishes` — and the attempt that finally landed had thrown the ported source away and rebuilt the app out of the host's catalog, losing the first wish.

`seed.wishes` replays in order on every **Update**, so both halves are data integrity: the list was recording what the *builder attempted*, and one of those attempts was a different app.

## Half 1 — a refused attempt became a wish

`vendo_make`'s edit arm recorded the ask whether or not the change landed (`make-tool.ts:287`), and the memory door appended it to `seed.wishes` unconditionally (`write-surface.ts:511`).

That is right for `memory.asks` — the next editor genuinely wants to read *"asked for X, then asked for X again, narrower"* — and wrong for the replay list beside it. The two answer different questions and were sharing one write.

`AppsRuntime.remember` now takes `landed`. The ask is recorded either way; only a change that reached the screen becomes a wish.

## Half 2 — a blind edit is answered out of the catalog

There is no explicit "fall back to a rewrite" branch. There is something worse: **the model was never shown the code it was supposed to edit.**

The port reaches it through `startingSource`, which is filled from the CHECKOUT — and the checkout only ever fills an **empty** workspace (`screen-agent.ts:1506`). The loop carries no file hand and `edit_app` can only replace passages the model quotes back character for character, so it cannot read the workspace itself. Once the first edit's save had landed a file, every later edit of that remix arrived with **no code at all** in front of it, and no "never replace it with something built from the catalog" line either.

An ask with nothing to change is answered out of the catalog. That is the whole fallback, and it is emergent — which is why fixing the cause makes the rewrite unreachable rather than detected. When the port genuinely cannot take an edit, `edit_app` refuses, nothing saves, and #1535's channel reports the real reason and re-arms the repair round. No second failure channel was invented.

The stored screen is now read for **every** edit of a remix, not only the first. It is still never written over a file a save left behind — showing it is not writing it.

Measured, before the fix, through the real path:

```
PROBE first-edit sees port:                       true
PROBE workspace holds app.tsx after first edit:   true
PROBE follow-up sees source:                      false
PROBE follow-up contains catalog warning:         false
```

## Proof

Two tests in `packages/vendo/tests/remix-wish-integrity.e2e.test.ts`, through a real PGlite store, a real `createVendo`, the real `vendo_make` tool called the way a calling agent calls it, the real checks floor, the real screen agent and the real workspace façade. Only the MODEL is scripted. Both failures are forced **deterministically** — a run that never saves, and a source-sensitive model that writes a catalog card exactly when the loop hands it nothing, which is what the live model did.

Each proven able to fail, reverting one fix at a time:

**1. drop the `landed` gate**
```
× a follow-up edit that does not land leaves the wish list exactly as it was
AssertionError: expected [ 'say it is mine', 'say it louder' ] to deeply equal [ 'say it is mine' ]
```

**2. restore the `!(await base.exists(checkout))` guard** — the prompt assertion goes red, and with it muted so does the outcome:
```
× the follow-up edits the PORT, so a refused edit can never become a catalog rewrite
AssertionError: expected 'import { Stack, Text } from "@vendo/s…' to contain 'text="Net worth at Maple, mine"'
+ <Text text="Rebuilt from the catalog" variant="heading" />
```
The catalog card really lands and the first wish is really gone. Restored → both green.

## Not weakened

Asserted or covered by suites re-run green: the list stays ordered and never trimmed, replays in order on Update, records an inapplicable wish on `seed.unapplied` and says it out loud, and the legacy `instruction` lift still parses pre-existing remixes. `memory.asks` still records the ask that failed — asserted in the new test, so the two halves can never quietly re-merge.

## Gate

`pnpm build` 22/22 · `pnpm typecheck` 44/44 · `pnpm lint` 0 errors · `@vendoai/apps` 1544 passed · `@vendoai/core` 813 passed · every remix/seed/memory suite in `@vendoai/vendo` green. Full suite is CI's.

---
## Summary by cubic
Fixes remix wish integrity: `seed.wishes` now records only landed edits, and follow-ups always edit the ported screen instead of regenerating from the catalog. Previously, refused edits appended to `seed.wishes`, and later edits saw no source after the first save, causing catalog rewrites that could drop earlier wishes.

- API: `AppsRuntime.remember` adds optional `landed`. `@vendoai/apps` passes `landed: true` when an edit reaches the screen. `memory.asks` still records all asks.
- Assembly: `@vendoai/vendo` reads the stored screen for every remix edit and includes it in the prompt, even when the workspace already holds a file; it never overwrites a saved file. Catalog fallback applies only when no screen exists.
- Tests: Adds `remix-wish-integrity.e2e.test.ts` proving both invariants end-to-end. Anchors the follow-up prompt assertion on the second ask to avoid brittle index-based checks.
- Migration: No data changes. Existing remixes may still contain historic refused asks in `seed.wishes`; cleanup is optional.

<sup>Written for commit d42e566940353597b03536c50f5ffc561bb21cfb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1563?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/remix-wish-integrity.md
+++ b/.changeset/remix-wish-integrity.md
@@ -1,0 +1,29 @@
+---
+"@vendoai/core": patch
+"@vendoai/apps": patch
+"@vendoai/vendo": patch
+---
+
+A remix's wish list records what the person GOT, and a follow-up edit changes the
+port instead of replacing it.
+
+One follow-up ask on Maple was refused three times and left four entries on
+`seed.wishes` — a list every Update replays in order, so one ask became four
+edits the person never made. The front door recorded the ask whether or not the
+change landed, which is right for `memory.asks` (the next editor wants to read
+"asked for X, then asked for X again, narrower") and wrong for the replay list
+beside it. `AppsRuntime.remember` now takes `landed`, and only a change that
+reached the screen becomes a wish. The ask itself is still recorded either way,
+the list is still ordered and never trimmed, and an inapplicable wish still lands
+on `seed.unapplied` and is still said out loud.
+
+The fourth attempt then abandoned the ported source and rewrote the app out of
+the host's catalog, losing the first wish's edit. The port reaches the model
+through `startingSource`, which was filled from the CHECKOUT — and the checkout
+only ever fills an EMPTY workspace. So once the first edit's save had landed a
+file, every later edit of that remix arrived with no code at all in front of it
+(the loop has no file hand and cannot read the workspace itself), and an ask with
+nothing to change is answered out of the catalog. The stored screen is now read
+for every edit of a remix, not only the first; it is still never written over a
+file a save left behind. A port that genuinely cannot take an edit now fails
+through the one channel there is, rather than succeeding as a different app.
--- a/packages/apps/src/server/doors/make-tool.ts
+++ b/packages/apps/src/server/doors/make-tool.ts
@@ -110,8 +110,10 @@ interface MakeCall {
   ask: string;
   /** The client parts this execution may publish, when the caller opened a stream. */
   stream: ((update: VendoViewStreamUpdate) => void) | undefined;
-  /** The ask, onto the app's memory. Best-effort, always. */
-  remember(appId: string): Promise<void>;
+  /** The ask, onto the app's memory. Best-effort, always. `landed` says the
+   *  change reached the screen, which is what makes it a remix WISH as well as
+   *  an ask — the create arms leave it off because a new app has no wish list. */
+  remember(appId: string, landed?: boolean): Promise<void>;
 }
 
 const makeNewApp = async (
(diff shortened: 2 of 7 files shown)


yousefh409:
Thanks — round 1. Two taken, two answered.

**Taken · `remix-wish-integrity.e2e.test.ts:285` (P3, hard-coded index)** — real, and a good catch. A later drive of the *first* ask satisfies all three assertions (it too carries the port with the first wish applied), so a fixed index would keep passing against the wrong prompt the day the drive count changes. Now anchored on the second ask, which only the follow-up's prompts can carry — the memory block is written *after* the edit, so the first run has never heard of it. Re-proved able to fail with the new anchor: reverting the `storedScreen` change reddens it against a real, non-empty prompt.

**Taken (as documentation) · `screen-agent.ts:1519` (P3)** — you offered "documenting/guarding that the two are assumed identical for a landed save" as a resolution, and that is the honest one, so the code now says it: a landed save wrote both copies, and §3.2 makes the ROW the truth where they part.

**Answered · `screen-agent.ts:1520` (P2, "use the held workspace file as `source`")** — same trade-off, and I think it goes the other way. The one way row and workspace part is a save the **floor refused**: `persistSource` commits those bytes to the workspace unconditionally, while `authoredScreen` only writes the row on floor `ok`. So the workspace copy in a divergence is *bytes the floor has already rejected*, and the row is the screen the person is actually looking at. Showing the workspace would hand the model rejected bytes to build on; showing the row costs at most a missed `edit_app`, which is reported through #1535's channel over a screen that still stands. Either way this is not a regression — before this PR that run was shown **nothing at all**, which is what produced the catalog rewrite.

**Answered · `make-tool.ts:292` (P2, no-op save recorded as a wish)** — correct that a save identical to the stored screen reports no `failure` and would record a wish. Out of scope here, deliberately: separating "landed" from "changed the document" means changing what the edit door reports about success, which is a wider change than this defect (a *refused* attempt becoming a wish) and belongs on its own. Worth noting the direction of travel — before this PR that no-op recorded a wish **and so did every outright failure**; this strictly reduces false wishes rather than adding any.

PR #1583, opened 2026-08-20
feat(ui,docs): an embed dropped on a bare page finds the wire itself

yousefh409:
## What

`VendoProvider` becomes an override instead of a prerequisite. `<VendoToolResult>`, `<VendoAppEmbed>` and `<VendoApprovalEmbed>` now render, poll, approve, deny and mount apps with no provider anywhere on the page.

- `useVendoProvider()` answers a shared default context when no provider is above it (`packages/ui/src/context.tsx`). Everything the provider carries already had a universal default: the wire at `/api/vendo`, auth riding the session cookie the browser already sends, and Vendo's own `--vendo-*` tokens.
- That default is ONE lazily-built value per bundle, so N embeds share ONE client. A fresh client per embed would be a fresh wire per embed — every poll keys its effect on client identity, so the loops would restart on every render, and each client would print the mount-mismatch paragraph on its own.
- Precedence is the `??`: a provider is non-null context, so the defaults are never consulted inside one. Provider-present behaviour is unchanged, observably.
- No new props, no config object, no environment variable. The headless hooks ride the same seam, so they self-boot too.

Docs: `existing-agent/embeds.mdx` splits into "No setup" and "The provider, when you need it"; the quickstart's render step no longer wraps the chat.

## Why

The first thing a BYO chat had to do was setup, for settings it could have assumed. It is not a second system — the provider was always "settings for the components inside me", and making it optional only adds "here's what I assume when you don't say".

## Verification

- [x] `pnpm build && pnpm test:affected && pnpm typecheck && pnpm lint` — build, typecheck, lint green; `@vendoai/ui` tests green. The four red packages (`store`, `actions`, `vendo`, `install-seam`) fail on `Cool%20Code` path decoding, the documented worktree gotcha — none of them depends on `@vendoai/ui`, and CI runs on a clean path.
- [x] Browser (local gate, production build): full `packages/ui` Playwright suite — 77 passed, 4 failed. The 4 (`accessibility` axe on thread-citations, `remixable-one-door` x3) fail identically on a clean stash of `origin/main`; pre-existing.
- [x] New browser spec `e2e/byo-embed-bare.spec.ts` over the new `/byo-embed-bare` harness scenario: both embeds bare on one page, the app polls `/api/vendo/apps/app_1/open` to its live surface, the approval polls `/api/vendo/approvals/apr_1` to a real consent card, both chrome boundaries themed off the defaults.
- [x] Prove-it-can-fail: restoring the throw reddens the 4 new bare unit tests and the 3 new bare SSR tests; making the defaults win over the provider reddens both new precedence tests; adding a `baseUrl?` prop to `VendoAppEmbedProps` reddens the new no-new-knob contract test at typecheck.
- [x] UI-affecting change verified in a real browser (production build, real Chromium).

---
## Summary by cubic
Embeds in `@vendoai/ui` no longer require `VendoProvider`. Previously they threw without a provider; now `<VendoToolResult>`, `<VendoAppEmbed>`, and `<VendoApprovalEmbed>` self-boot against `/api/vendo` with session-cookie auth and default theme. A provider still overrides, and bare embeds share one lazily-built client.

- Core change: `useVendoProvider()` now returns a shared default context when no provider is found (`packages/ui/src/context.tsx`). Provider precedence is unchanged; the default is built once per bundle, shared across renders, and is SSR-safe.
- No API changes: embeds and headless hooks self-boot; no new props or config. Migration is not required and provider-backed pages behave the same.
- Docs and tests: `existing-agent/embeds.mdx` adds “No setup” and provider guidance; quickstart stops wrapping the chat. Added e2e for a bare page and unit/SSR tests for provider precedence, client sharing, default theme, and default wire calls.

<sup>Written for commit 32980db1ccc4f5135ae826653769b651c7c7bbc4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1583?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


yousefh409:
CI note: `ai-dual (vendo-1..4)` is red here, and it is not this branch. The identical four jobs fail on the merge_group run for #1575 (run 32429119413) and on #1582's branch — every run since ~23:17Z. The ai@6 `test-shards (vendo-1..4)` pass on this PR, `ai-dual (ui-1, ui-2)` pass, and this branch changes no `@vendoai/vendo` source. The failures are `app build failed: this run never saved a screen` beside an `AI SDK Warning … "specificationVersion" is used in a compatibility mode`, which reads as a fresh `ai@7` / `@ai-sdk/*` publish the mock model in `packages/vendo`'s tests no longer satisfies.

PR #1599, opened 2026-08-23
fix(repo,vendo): the repo's spawns find their tools on Windows

YSharara1:
## What

Node has refused to exec a `.cmd`/`.bat` without a shell since the fix for CVE-2024-27980. On Windows every package manager and every `node_modules/.bin/*` entry **is** one of those shims, so `spawn("npm", …)` and `spawn(<…>/.bin/next, …)` ENOENT before the child ever starts. 9 sites across `packages/`, `fixtures/`, `corpus/` and `examples/`.

Two fixes, applied by kind:

- **a package manager** → branch on the platform and hand Windows ONE command string with `shell: true`. Not command + args: that combination is DEP0190, and Node 24 prints the warning by default. Args are quoted because `cmd.exe` reads a bare `^` (as in `ai@^6`) as an escape — the reasoning `provider-deps.ts` already carried.
- **a `node_modules/.bin/<tool>` entry** → don't run the shim at all. Spawn `process.execPath` against the tool's real JS entry (`next/dist/bin/next`, `typescript/bin/tsc`): no shell, no quoting, one code path on both platforms. The repo already did this at `fixtures/existing-agents-e2e/src/journey.ts:202`; this makes it the rule.

## Why

CI is ubuntu-only — every `runs-on:` in `.github/workflows`. Nothing that runs on a green PR can observe any of this. Measured on a Windows checkout of this branch:

| | before | after |
|---|---|---|
| `spawn(node_modules/.bin/next, …)` | `ERROR ENOENT` | `ok Next.js v16.2.11` (via `process.execPath` + `next/dist/bin/next`) |
| `vendo` install, on a user's screen | `[DEP0190] DeprecationWarning: …security vulnerabilities…` | *(nothing)* |

The second one is the one that reaches users: the CLI printed a Node security-deprecation warning about its own spawn, in the middle of its own output, on every install.

**Rebased onto #1623.** This branch originally carried 13 sites. #1623 deleted `packages/box-template` outright and rewrote `packages/apps/box/build-template.mjs` so it no longer spawns at all — which retires 4 of them, along with the `taskkill` teardown fix that belonged to the deleted `box-template.test.ts`. Those are gone from the diff rather than carried forward; the hazard they covered no longer exists. The gate re-run on the merged tree is green, so #1623's replacement box machinery did not reintroduce it.

**On `pnpm lint`.** It dies on step 1 on Windows too — `dependency-guard.mjs` cannot parse a CRLF `pnpm-workspace.yaml`. I hit it here and fixed it before finding that **#1447 already carries that fix** (ENG-430, bug 3), in a better form: it splits on `/\r?\n/` rather than rewriting the file first. Backed out rather than shipping a second, colliding fix — `pnpm lint` on Windows is #1447's to unblock. The gate added here runs standalone, so nothing in this PR waits on it.

## Verification

- `packages/vendo` `tests/cli/provider-deps.test.ts` — **50/50**, including 4 new, re-run after the rebase. `installSpawnPlan` takes its platform as an argument, so the win32 shape is asserted **on ubuntu** — the only place in the repo it is checked at all.
- `node scripts/windows-spawn-gate.mjs` — ok, 2214 files, 1 documented exception. **Proved non-vacuous twice**: once by reintroducing a plain hazard, and once by probing the exact false negative review named — an unrelated `win32` near the top of a file and an unguarded spawn at the bottom, which the original file-level check passed and the neighbourhood check catches at the right line.
- `tsc --noEmit` clean on all six fixtures, `corpus/harness`, `packages/actions`, and on `provider-deps.ts`. The remaining errors in `packages/vendo` are stale-dist `has no exported member` ones, identical on an unmodified tree (checked with `git stash`).

### The gate

`scripts/windows-spawn-gate.mjs`, wired into `pnpm lint` beside the portability gate. It judges each hazard by its **neighbourhood**: not per line, because a correct fix keeps the POSIX spawn verbatim in its else branch and a line-level scan flags the very shape it is asking for; and not per file, because one `win32` mentioned for a path separator would vouch for an unguarded spawn hundreds of lines away. `win32` has to appear within 1200 characters of the spawn, which is where the branch that fixes it would be. Comments are stripped (blanked, so line numbers survive) for the same reason. `genbench/src/codex.ts` is a named exception carrying its reason (codex is a native binary published through npm; it has no JS entry to run on `process.execPath`), kept as an explicit list so that skipping something stays visible in review.

**From review (`fa26e790`).** Two gaps Greptile named, both real, both closed. `installSpawnPlan` now **refuses** an arg carrying a `"` of its own instead of wrapping it and hoping — cmd.exe has no escaping story worth trusting, and this builds a line for a shell, so a wrong guess is a command injection rather than a typo. And the gate judged per FILE, which let any `win32` anywhere vouch for every spawn in it; each hazard is now judged within 1200 characters of itself. Comments are blanked rather than deleted when stripped, so the reported `file:line` still points at the real line.

### Not in scope

15 `corpus/harness` failures, **pre-existing and unchanged** — verified by running the suite with and without this branch's one change to that package: 15 failed / 174 passed both times. Those tests write extensionless shell scripts as fake `npm`/`pnpm` binaries, which `cmd.exe` cannot run. That is the test doubles, not the product's spawns, and it wants its own change.

- [x] `pnpm lint` — `windows-spawn-gate` green standalone. `dependency-guard` needs #1447 to run on Windows at all; `portability-gate` reads built output and needs `pnpm build` first. Both left to CI.
- [ ] Screenshots attached (if UI-affecting) — n/a

### DIFF
--- a/.changeset/the-cli-install-stops-warning-about-itself.md
+++ b/.changeset/the-cli-install-stops-warning-about-itself.md
@@ -1,0 +1,17 @@
+---
+"@vendoai/vendo": patch
+---
+
+The Windows install no longer prints a deprecation warning about our own spawn.
+Package managers are `.cmd` shims there, so the install has to go through the
+platform shell — but it was passing an args array alongside `shell: true`, and
+that pair is DEP0190. Node 24 prints it by default, onto the CLI's own stderr,
+so `vendo` interrupted its install with a security warning about itself. Windows
+now gets the whole line as one command string; nothing changes on any other
+platform.
+
+The quoting the warning was complaining about is still there and still needed —
+`cmd.exe` reads a bare `^` (as in `ai@^6`) as an escape character outside double
+quotes, and unquoted that spec reaches npm as a different, pinned `ai@6`. It has
+just moved into `installSpawnPlan`, which takes its platform as an argument so
+the Windows shape can be asserted from an ubuntu-only CI.
--- a/corpus/harness/src/ai/matrix.ts
+++ b/corpus/harness/src/ai/matrix.ts
@@ -271,12 +271,19 @@ function resolveAgentSdk(sdkDir: string): string | null {
   }
 }
 
+/** npm is a `.cmd` shim on Windows and Node has refused to exec one without a
+ *  shell since CVE-2024-27980, so a bare spawn ENOENTs and the SDK provision
+ *  fails before npm ever starts. Windows takes the whole line as ONE string —
+ *  an args array alongside `shell: true` is DEP0190 and warns on every run. The
+ *  spec is quoted because cmd.exe reads `^` (as in `@^1.2`) as an escape
+ *  character outside double quotes; the same reasoning as provider-deps.ts. */
 function npmInstallAgentSdk(sdkDir: string): Promise<void> {
   return new Promise((resolve, reject) => {
-    const child = spawn("npm", ["install", "--no-audit", "--no-fund", `${AGENT_SDK_PACKAGE}@${AGENT_SDK_VERSION}`], {
-      cwd: sdkDir,
-      stdio: ["ignore", "ignore", "pipe"],
-    });
+    const spec = `${AGENT_SDK_PACKAGE}@${AGENT_SDK_VERSION}`;
+    const stdio: ["ignore", "ignore", "pipe"] = ["ignore", "ignore", "pipe"];
+    const child = process.platform === "win32"
+      ? spawn(`npm install --no-audit --no-fund "${spec}"`, { cwd: sdkDir, stdio, shell: true })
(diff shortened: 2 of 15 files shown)


YSharara1:
Superseded by [runvendo/vendo-cloud#81](https://github.com/runvendo/vendo-cloud/pull/81).

This repo is now a one-way copybara projection of `runvendo/vendo-cloud` — its `main` is written only by `vendo-release-bot`, so a PR here cannot land. The change will appear here automatically once #81 merges.

Not a straight cherry-pick: the fold (#71, S11d) moved several of the files this touched, and every bug was re-verified as still live on current main before re-cutting. Details on #81.

PR #1647, opened 2026-08-26
fix: reseal the stale MCP shim bundle and delete the dead in-client review chain

yousefh409:
Finishes the in-client removal: fixes a checked-in stale compiled executor (security), deletes the dead `review`-flag chain and orphaned code, and keeps the live guards. One PR, three priorities.

## Priority 1 — the stale MCP shim executor (security)
`packages/mcp/src/shim/shim-html.gen.ts` is a checked-in generated bundle (built from `packages/ui/src/tree/mcp-shim/entry.tsx`) that was compiled **before** the in-client removal. The committed bytes still carried a live `new Function` executor for `source:"generated"` nodes and lacked the current refusal.

- **Regenerated** it. The fresh bundle now contains the refusal (`"…no longer runs in the host page"`) and has **zero** `new Function` executors (verified by grep). The only residual `inClient`/`granted` tokens are the live strip defense (`delete A.inClient,delete A.pinDrift`) and the zod schema field that gets stripped — no executor.
- **Made the generator deterministic**: it now forces `NODE_ENV=production` before importing vite, regardless of the caller's ambient env. Without this, running the generator under `NODE_ENV=test` bundled the ~70 KB-larger React dev build, so the same source produced a different file for different callers.
- **Enforcer chosen: a vitest drift guard** (`packages/ui/test/mcp-shim-bundle.test.ts`) that spawns the generator to a temp file and asserts it byte-matches the committed bundle. Runs in CI's existing `ui` shard (deps are built there before tests), needs **no** workflow change, and is far less invasive than wiring a 3.3 MB regeneration into every `build` + a dirty-tree check. Proven it fails on drift (red) and passes fresh (green, ~2.2 s). The generator takes an optional `VENDO_MCP_SHIM_OUT` output override so the test never touches the committed file.
- **Fixed the e2e spec** (`packages/ui/e2e/mcp-shim.spec.ts`): dropped the `inClient: GRANTED` fixtures; the interactive "Run shim action" node and the theme-proof node are now built-in kit nodes (`Button`) — `source:"generated"` execution is categorically gone, so the action rides a kit node and the theme proof rides an accent Button whose background is `var(--vendo-color-accent)`.

## Priority 2 — delete the dead `review`-flag capture chain
Nothing downstream acts on `seed.review` / `baseline.review` anymore. Removed `WrapperSite.review`, `reviewFlag()`, the per-site capture, the `sameCapturedPayload` equality term, the "wrapped both with and without review" warning, and the persist (`packages/actions/src/sync/seeds.ts`); the `review?: boolean` field + zod entry from `SeedBaseline` (`packages/apps/src/contract/seed.ts`) and `AppSeed` (`packages/core/src/app-document.ts`); `baseline.review` carry in `seed-surface.ts`; and the `review` term in `upToDate()` (`packages/vendo/.../seed-baselines.ts`).

**Critical:** the `if (!review)` guard that suppressed the "reaches into host plumbing" warning is deleted — the warning is now **unconditional** (every fork renders sandboxed, so the suppression was wrong).

## Priority 3 — orphaned harness + stale comments
Deleted the `inClientSource`/`inClientTree`/`InClientScenario`/`reviewStandingTree`/`ReviewStandingScenario` scenarios and their `/tree-inclient` + `/tree-review` routes (`packages/ui/e2e/harness/main.tsx`, no spec routed to them), the dead `"in-client-approve"` audit union member (`audit-reports.ts`, `runtime-context.ts`, zero call sites), and two stale comments.

## Keep (live, not dead)
The `inClient`/`pinDrift` strip defense, the `IN_CLIENT_*`/`isPinnedPackage`/`isIslandResolvableSpecifier` sandboxed-path constants, the `--review` CLI flag, `validateWrittenApps({review})`, `reviewModel`/`seats.review`, and the renderer `onReview` (Guard approvals).

## Gate
`pnpm build && pnpm typecheck && pnpm lint` all green; `pnpm test:affected` green except the two **pre-existing environmental** failures the brief called out — `actions/tests/sync/split.test.ts` (macOS `/private/tmp` symlink path) and the `vendo` `init`/`doctor`/`provider-deps`/`init-install-dx` ai@6/7 + local-workspace-detection family — none of which this PR touches. The drift guard's one failure during the gate was itself: it correctly caught that the committed bundle was stale relative to the Priority-2 schema edits, and regenerating resolved it. Changeset: minor bump for the five affected published packages (dependents auto-bump). No browser run here — the e2e specs are the local pre-PR gate for a maintainer spot-check.


yousefh409:
Both P2s on the shim drift-guard are addressed in 0fdf357e0 (test-only, `packages/ui/test/mcp-shim-bundle.test.ts`).

**P2 #1 (security — both drift AND reintroduce).** Added a direct executor-absence assertion that runs on the committed bytes and on a fresh rebuild, independently of the byte-compare. Each bundle must (a) NOT match an executor primitive and (b) contain the refusal marker `no longer runs in the host page`. The regex is anchored to the executor call form and gated on a non-identifier, non-`.` prefix:

    /(^|[^.\w$])(new\s+Function|Function|eval)\s*\(/

This is precise against the shipped bundle: the minified QuickJS interpreter contains 33 innocuous `...Function(` identifiers (callFunction(, parseFunction(, QTS_NewFunction(, …) and the regex matches 0 of them, while it matches `new Function(`, a bare indirect-eval `Function(`, and `eval(`. Proof: temporarily appending `new Function(` to the committed bundle turns the committed security test RED while the freshly-built security test stays GREEN — so re-adding an executor to the source fails CI even when committed==fresh.

**P2 #2 (byte-exact brittleness).** Kept the byte-exact freshness check but demoted it to a secondary staleness nudge (documented in-file: 'regenerate on toolchain bump'), because the executor-absence assertion above is now the real security line and does not depend on byte-equality. Chose keep-byte-exact-plus-docs over normalize/strip because it is strictly simpler and a cosmetic Vite/esbuild/Rollup output change is a maintenance chore, not a vulnerability.

Gate: `@vendoai/ui` build + `vitest run test/mcp-shim-bundle.test.ts` green (3 passed).

yousefh409:
Narrowed the executor-absence claim in `packages/ui/test/mcp-shim-bundle.test.ts` (2391441ae) per this thread.

The comments and the `expect()` failure message no longer frame the `EXECUTOR` regex scan as a complete adversarial proof of "no dynamic execution." They now say plainly: it's a targeted regression tripwire for the specific forms the removed native executor used (`new Function`, bare `Function(`, `eval(`) — and it does not catch deliberately obfuscated/indirect forms (`globalThis.Function(`, `obj.Function(`, `["Function"](...)`, `.constructor(...)`).

The actual load-bearing guarantee is now stated explicitly: the committed bundle byte-matches a fresh build of the current source (the third test), and the current source (`renderer.tsx`) refuses `source:"generated"` execution — so a dynamic executor can only reappear in the bundle if one is reintroduced in source, which the source-level refusal and its review are what prevent, not this scan.

The regex, the assertions, and the test behavior are unchanged (still zero false positives against the bundled QuickJS interpreter's identifiers) — broadening the scan to catch the obfuscated forms is intentionally avoided since it would false-positive against those identifiers. Only the wording changed.

PR #1654, opened 2026-08-27
feat(telemetry): send operational events to PostHog Logs, not analytics

yousefh409:
## What

`doctor_run`, `command_run` and `agent_run` are records of work — "this ran, here is how it went" — not funnel steps. They now go to PostHog's **Logs** product (OTLP over HTTP, `/i/v1/logs`), where retention is enforced at 30 days by the org's `logs_retention_30d` add-on, instead of the product-analytics stream that keeps events indefinitely.

Everything else about them is unchanged: same public write-only project key, same closed allowlist, same `errorDetail` scrubbing, same cloud/internal lane markers, same opt-outs, same `VENDO_POSTHOG_HOST` override. Nothing is dual-written — `LOG_EVENTS` in `events.ts` is the whole switch. `init_started`, `init_completed`, `init_failed`, `star_prompt` and `error_class` still go to `/capture/`.

Existing historical rows stay in analytics and age out on their own; that is accepted.

## Why these three

Checked every consumer in project 397743 before touching code: 18 insights, 5 dashboards, 1 action, 2 cohorts, 15 destinations. **None** references `doctor_run`, `command_run` or `agent_run`. Nothing breaks.

## Contract

`TELEMETRY.md` is the published contract and is updated with the destination table, the OTLP shape, and the retention. Credential stays public-by-design (the same `phc_` key, which the Logs endpoint accepts as `?token=` — its documented alternative to a bearer header, so both transports stay header-free).

## Seam proof

Write and read are independent — no stub on either side:

1. Ran the real CLI: `vendo doctor` from `packages/vendo/bin/vendo.mjs`, real socket transport, temp `HOME`, consent granted.
2. Read it back through PostHog's HogQL `logs` table by the run's `distinct_id`:

```
timestamp     2026-08-27T04:24:47Z
service_name  vendo-sdk
event_name    doctor_run
attributes    { event: doctor_run, distinct_id: 018f9688-…, failures: "6",
                warnings: "5", wired: "false", vendoVersion: "0.51.1",
                osPlatform: darwin, nodeVersion: v24.2.0, projectIdHash: 2c57a3e6…,
                internal: "true" }
```

3. Queried the analytics stream for the same `distinct_id` over the same window: **zero rows**. The event moved; it is not duplicated.

## Test changes (stated out loud)

`client.test.ts` used `agent_run` as its generic carrier for the base-props, lane-marker and allowlist assertions. Since `agent_run` now takes the logs lane, those 15 uses move to `error_class` — an analytics-lane event with the same base-only shape — so each assertion keeps its exact meaning. A new `logs lane` block covers the new destination: endpoint, OTLP shape, `service.name`, the `eventName`/`event`/`distinct_id` fields, the routing split, and that the allowlist, lane markers and consent behave identically there.

The J11 wire seam (`fixtures/integration`) decodes both wire shapes and now asserts the **path** the real composed client posts to — the only place the destination is observable end to end.

Verified the new tests can fail: removing `doctor_run` from `LOG_EVENTS` turns 2 of them red; restored.

## Gate

`pnpm build` ✅ · `pnpm typecheck` ✅ (43/43) · `pnpm lint` ✅ (0 errors) · touched-scope tests ✅ — `packages/vendo-telemetry` 133/133, `fixtures/integration` telemetry-wire 5/5.

⚠️ The full `pnpm test:affected` could not complete on this machine: the disk is at 100% (~120 MiB free) and several unrelated packages failed with `No space left on device` (PGlite could not create its data dir, corpus-harness could not write a git config). CI is the gate of record.

---
## Summary by cubic
Moves `doctor_run`, `command_run`, and `agent_run` from PostHog's product-analytics stream to PostHog's Logs product, so operational "this ran" records land in a store with enforced 30-day retention instead of being kept indefinitely. They're now sent as OTLP/JSON to `/i/v1/logs` with the same write-only project key, allowlist, scrubbing, opt-outs, and `VENDO_POSTHOG_HOST` override; no event is dual-written, and `init_started`, `init_completed`, `init_failed`, `star_prompt`, and `error_class` keep going to `/capture/`.

**Notes**
- `TELEMETRY.md` now documents both destinations, the OTLP shape, and the retention.
- The logs lane renders every property value as a string (`ok` arrives as `"true"`, not `true`); the CLI command-test seam decodes both shapes so tests assert in the right spelling.
- No consumer in project 397743 references the moved events, so nothing breaks.
- Historical rows stay in analytics and age out on their own; no migration needed.

<sup>Written for commit 90bff89f6a1780ea8162617f49fc6c2c6ae31637. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1654?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/telemetry-operational-events-to-logs.md
+++ b/.changeset/telemetry-operational-events-to-logs.md
@@ -1,0 +1,5 @@
+---
+"@vendoai/telemetry": minor
+---
+
+Send the operational events — `doctor_run`, `command_run`, `agent_run` — to PostHog's Logs product instead of the product-analytics stream, so a record of "this ran" lands in a store with enforced 30-day retention rather than being kept indefinitely. Same project key, same allowlist, same scrubbing, same opt-outs, same `VENDO_POSTHOG_HOST` override; only the destination and the retention change, and no event is sent to both. `init_started`, `init_completed`, `init_failed`, `star_prompt` and `error_class` are unchanged.
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -85,7 +85,16 @@ Set `"optedOut": false` in `~/.vendo/telemetry.json` to clear the local opt-out 
 
 ## Where Data Goes
 
-Product events are sent to PostHog US Cloud using a write-only project key. Set `VENDO_POSTHOG_HOST` to send them to your own PostHog instead (`VENDO_POSTHOG_KEY` sets the project key); the path is always `/capture/`.
+Events are sent to PostHog US Cloud using a write-only project key. Set `VENDO_POSTHOG_HOST` to send them to your own PostHog instead (`VENDO_POSTHOG_KEY` sets the project key).
+
+There are two destinations, chosen by event, and nothing is ever sent to both:
+
+| Destination | Events | Path | Retention |
+| --- | --- | --- | --- |
+| Product analytics | `init_started`, `init_completed`, `init_failed`, `star_prompt`, `error_class` | `/capture/` | Kept |
+| Logs | `doctor_run`, `command_run`, `agent_run` | `/i/v1/logs` | 30 days, enforced |
+
+`doctor_run`, `command_run`, and `agent_run` are operational records — "this ran, here is how it went" — not funnel steps, so they go to PostHog's Logs product, which enforces a 30-day retention window. They are sent as OTLP/JSON with `service.name` `vendo-sdk`; the event name rides as both the log record's `eventName` and an `event` attribute, the anonymous id as a `distinct_id` attribute, and each allowlisted property as an attribute of the same name. The allowlist, the scrubbing, and every opt-out below apply identically on both paths — moving an event changes only where it is stored and for how long.
 
 Network calls are fire-and-forget, use a short timeout, and failures are swallowed so telemetry cannot break builds or dev servers. The capture socket is unref'd the moment it exists, so a telemetry POST can never keep the CLI running after a command finishes — on a captive-portal network that accepts the connection and never answers, `vendo init` still exits as soon as it prints its summary.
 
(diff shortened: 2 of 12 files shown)


yousefh409:
Paired with runvendo/vendo-web#661, which moves the cloud half (`guard_decision`, `deployment_boot`, `agent_run`, `sdk_error`) to the same destination. Ship together: `agent_run` has a producer in **both** repos — the CLI/wire server here, and the SDK-forwarded one there — so the name only actually leaves the analytics stream when both land.

yousefh409:
### Follow-up: one more test seam the routing change reached

`pnpm test:affected` caught 6 failures the changed packages' own suites could not: `telemetryCapture()` in `packages/vendo/src/cli/telemetry.test-util.ts` is the shared seam every tracked CLI command asserts through, and it lives in `@vendoai/vendo` while the routing lives in `@vendoai/telemetry`. Fixed in 90bff89f6 — the helper now decodes both destinations, and the `command_run` assertions read the logs lane's spelling (OTLP renders every value a string, which is how PostHog stores it either way).

Worth stating plainly: this is the *third* test file in this change whose only assumption was "there is one body shape". That is the real blast radius of a destination split, and it is why the new `logs lane` block in `client.test.ts` asserts the endpoint explicitly rather than leaving it implied.

yousefh409:
Full local gate now green (the disk pressure that broke the first run cleared): `pnpm build` ✅ · `pnpm test:affected` ✅ **exit 0, 0 failing tests, 0 red packages, 0 disk errors** (`@vendoai/vendo` 2810 passed / 45 skipped, `@vendoai/telemetry` 133/133, genbench 604, corpus-harness 189, integration 47, automations-e2e 68, demo-bank 138, + the rest) · `pnpm typecheck` ✅ 43/43 · `pnpm lint` ✅ 0 errors, portability-gate all legs green.

PR #1661, opened 2026-08-27
fix(ui): the build consent card says what a build actually does

yousefh409:
## What was wrong

A person approving a **build machine** — the hero beat of the escalate-to-box flow — read the generic consequence line instead of the sentence Vendo authored for exactly that card:

> This changes something in your account, and it runs as you.

The authored copy lived at `packages/apps/src/server/doors/build-door.ts:49` and never reached the card. `consentAsk` (`packages/ui/src/chrome/build-beat.tsx:162`) reads only the host's `ToolMeta.description`, and `vendo_app_build` is a tool **Vendo** ships, so no host meta exists for it — the ladder fell straight through to `consentClassLine("write")`.

This was a real divergence, not a cosmetic one: `BUILD_CONSENT_ASK` (`build-door.ts:56`) ships the authored sentence to the surfaces that render no card (text, MCP, an outside agent), so the same ask said two different things depending on where you answered it. `PendingApproval`'s own contract (`packages/core/src/tools.ts:419-423`) says that pair is "never a second vocabulary for one ask".

The doc comment at `build-door.ts:61-64` claimed the card derived its words from the descriptor. It didn't — ruling 14 (`build-beat.tsx:145-151`) had already taken descriptor descriptions off the ladder at every rung, and the comment was left behind.

## The fix

`VENDO_TOOL_NOTES` in core, beside `VENDO_TOOL_TITLES`, for the same stated reason that table exists: two sides must say the same words and neither can read the other's copy. The consent ladder reads it **under** the host's own `ToolMeta.description` and above the consequence class.

Ruling 14 is untouched. It bans a *descriptor's* description because that sentence is authored for the model or minted by extraction; nothing extracted can land in a table Vendo writes by hand.

## Gate

`pnpm build` · `test` (core, ui, apps) · `typecheck` · `lint` — all EXIT 0.

---
## Summary by cubic
Fixes the build consent card so a person approving a build machine sees Vendo's authored sentence instead of the generic consequence line, and the card and the text-only surfaces now say one thing.

**Bug Fixes**
- The authored sentence now sits in `VENDO_TOOL_NOTES` in core and is read as a fallback under the host's own `ToolMeta.description`.
- A host that authors a blank description falls through to that note; the ladder uses `||` instead of `??` so an empty string can't suppress the rung below it.
- The seam test pins the card's words against `VENDO_TOOL_NOTES` itself and asserts the generic line never appears, so the two surfaces can't drift apart again.
- Ruling 14 is untouched: descriptor descriptions stay off the consent ladder, and only copy Vendo writes by hand can reach the card.

<sup>Written for commit fad2deccc95fe40fe0cbdc6d70820d1f4c6da612. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1661?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/build-consent-card-authored-copy.md
+++ b/.changeset/build-consent-card-authored-copy.md
@@ -1,0 +1,7 @@
+---
+"@vendoai/core": patch
+"@vendoai/ui": patch
+"@vendoai/apps": patch
+---
+
+fix: the build consent card says what a build actually does. Ruling 14 keeps a descriptor's description off the consent ladder, so the authored build sentence never reached the card and a person approving a build machine read the generic "This changes something in your account, and it runs as you." The copy Vendo writes by hand for its own tools now lives beside `VENDO_TOOL_TITLES` as `VENDO_TOOL_NOTES`, which the ladder reads under the host's own `ToolMeta.description` — so the card and the words-only surfaces (`BUILD_CONSENT_ASK`) say one thing, and nothing extracted can reach that rung.
--- a/packages/apps/src/server/doors/build-door.ts
+++ b/packages/apps/src/server/doors/build-door.ts
@@ -11,6 +11,7 @@
 import {
   VENDO_APP_BUILD_TOOL,
   VENDO_APP_FORMAT,
+  VENDO_TOOL_NOTES,
   VENDO_TOOL_TITLES,
   VendoError,
   type AppBuildProposal,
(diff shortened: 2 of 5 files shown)


yousefh409:
Triaged, two findings — one fixed, one incorrect.

**#1 (`build-beat.tsx:165`, blank host description suppresses the note) — real, and fixed in `fad2dec`.** This one is introduced by this PR: the line went from `meta?.description?.trim()` to `(meta?.description ?? presentation.note)?.trim()`, and `??` keeps an empty string, so a host that authored *no* sentence still outranked `VENDO_TOOL_NOTES` and the card fell back to the generic consequence line — the exact failure the new rung exists to remove. Now `||`, with a comment naming the gotcha.

**#2 (`core/tools.ts:179`, the seam test still asserts the old phrase and now fails) — incorrect.** The PR already updates it: `screen-escalate-consent.seam.test.ts:366` was changed in this diff to assert `VENDO_TOOL_NOTES[VENDO_APP_BUILD_TOOL]` and to add a negative assertion on the old generic line. That is why the suite is green.

PR #1673, opened 2026-08-27
fix(voice): keep tool identifiers and run-on sentences out of the answer

yousefh409:
Two defects photographed in one TaxDome answer, both legible in the payoff frame of the sales GIF.

## 1 · the model had no word for a tool but its identifier

`modelToolDescription` (`packages/core/src/tools.ts`) dropped the human label whenever the listing carried no authored `title` — and `ToolListing.title` falls back to the tool's own NAME (`packages/harnesses/src/turn-tools.ts:248`), so "no title" is the ordinary case for a host whose `.vendo/tools.json` has descriptions and nothing else. TaxDome is exactly that host: **zero** of its 15 tools author a title.

So the model was handed `description` alone, leaving the identifier as the only proper noun it held — and it printed `host_getClient`, `host_listJobs` and `host_getRevenueByMonth` to a user whose own `designRules` forbid showing an internal id.

The prompt already promised this could not happen (`packages/vendo/src/prompt.ts:44` — *"Each tool's description leads with its human title before an em dash; say the title, never the identifier — not even in backticks"*). The instruction was unfollowable: there was no title to say. This makes the promise true by walking the **same ladder the render layer already walks** (`toolTitle`, `packages/ui/src/chrome/humanize.ts:54`) — Vendo's own table, then the prettified id. The screen's beat already said "Get client"; now the model has the same word. Two sides, one ladder, so they cannot say different things. The CALL name is untouched.

## 2 · two of the model's own paragraphs ran together mid-sentence

`vendo()` forwarded `text-delta` but dropped the SDK's `text-start`/`text-end` block boundaries into its `default` case, and the wire's `TextChannel` opens a fresh transcript part only when a **tool call** is mirrored. Two adjacent text blocks therefore landed in one part with no separator: `"…exposed here.No matching tool exists…"`. A block boundary now travels as a paragraph break, which keeps the `HarnessEvent` union closed (§1.5).

## Proof on the real path

New seam test `packages/harnesses/tests/answer-voice.test.ts` drives the **real** `vendo()` loop through the **real** runtime and reads the **real** wire parts — no stub on either side; only the model provider is scripted. Both assertions fail on `main` with exactly the photographed symptoms:

```
AssertionError: expected 'Look up one client account by id.' to be 'Get client — Look up one client accou…'
AssertionError: expected 'There\'s no document upload/filing ca…' not to contain 'here.No matching'
```

## Test change, stated out loud

`modelToolDescription`'s case *"leaves the description alone when there is no title to add"* asserted the **defect** — it pinned the behaviour that leaked the identifier. It is rewritten to assert the fallback, plus one new case pinning Vendo's own table for Vendo's own tools.

---
## Summary by cubic
Fixes two defects that appeared in one live TaxDome answer: the model named tools by their internal identifiers (like `host_getClient`) instead of human titles, and its own adjacent text blocks ran together without a separator.

**Bug Fixes**
- `modelToolDescription` now falls back to a human title from Vendo's own title table, then the prettified id, when a host authors no tool title; the CALL name is unchanged.
- `vendo()` now renders model text-block boundaries as paragraph breaks instead of dropping them.
- A new seam test drives the real loop and wire, and the old test that pinned the leaking behavior now asserts the fallback.

<sup>Written for commit c16b566ed35ee7668548afb1a546f816a0e14e61. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1673?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/tool-ids-out-of-the-answer.md
+++ b/.changeset/tool-ids-out-of-the-answer.md
@@ -1,0 +1,10 @@
+---
+"@vendoai/core": patch
+"@vendoai/harnesses": patch
+---
+
+fix: keep internal tool identifiers and run-on sentences out of the answer a user reads.
+
+`modelToolDescription` dropped the human label whenever a host authored no `title` (or the listing's title had fallen back to the tool's own name), so on a host whose `.vendo/tools.json` carries descriptions but no titles the identifier was the only proper noun the model held — and it printed `host_getClient`, `host_listJobs` and `host_getRevenueByMonth` in a live answer, on a host whose own design rules forbid showing an internal id. The label now falls back down the same ladder the render layer already walks (Vendo's own title table, then the prettified id), so the beat on screen and the model's vocabulary say the same words instead of the screen saying "Get client" while the model has nothing but `host_getClient`. Nothing about the CALL name changes.
+
+`vendo()` also dropped the model's own text-block boundaries, and the wire opens a fresh transcript part only when a tool call is mirrored — so two adjacent blocks ran together mid-sentence ("…exposed here.No matching tool exists…"). A block boundary now travels as a paragraph break.
--- a/packages/core/src/tools.ts
+++ b/packages/core/src/tools.ts
@@ -200,16 +200,22 @@ export function humanizeToolName(raw: string): string {
  * identifier stays the CALL name; this is the one place the human label enters
  * the model's vocabulary.
  *
- * A title equal to the name adds nothing (`ToolListing.title` falls back to the
- * name) and would teach exactly the wrong vocabulary, so it is dropped.
+ * A title equal to the name is no title at all (`ToolListing.title` falls back to
+ * the name), and neither is a missing one — so the label falls back down the SAME
+ * ladder the render layer walks (`toolTitle`, ui/humanize.ts): our own table, then
+ * the prettified id. Dropping the label instead is what leaked
+ * `host_getClient` into a TaxDome answer — the screen's beat said "Get client"
+ * while the model, told to say the title and given none, had only the identifier.
+ * Both sides now read one ladder, so they cannot say different words.
  */
 export function modelToolDescription(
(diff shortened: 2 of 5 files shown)


yousefh409:
Triaged, four findings:

**#2 (`humanizeToolName` returns the raw id) and #3 (`constructor` / `toString` hit `Object.prototype`) — dismissed as out of scope.** Both describe `VENDO_TOOL_TITLES[name] ?? humanizeToolName(name)`, which is the *existing* ladder in `ui/chrome/humanize.ts:54` (`toolTitle`), shipping on `main` today. This PR's whole point is that the model reads the **same** ladder as the render layer, so hardening one copy and not the other would re-open exactly the drift being closed. If the prototype lookup is worth guarding it should be guarded once, in both places, as its own change — noted as a follow-up.

**#1 (the comment overclaims) — fair, and the claim is narrowed rather than the signature widened.** `toolTitle` honours `meta.label` at a rung `modelToolDescription` genuinely cannot see: it receives `{name, title, description}` and no `ToolMeta`. So for a host that enriches a label but authors no title the two surfaces can still differ. Threading `meta` into `modelToolDescription` is a real API change and not worth it here; the behaviour is strictly better than the `undefined`-drop it replaces. Logged so the comment gets the caveat.

**#4 (`blockEnded` survives a mirrored tool call) — dismissed, P3 and cosmetic.** A leading `\n\n` on a part the wire has *already* opened is invisible once rendered as markdown; the failure it prevents (two blocks fused mid-sentence) is not.

None of these is a correctness blocker; all four required checks are green.

PR #1674, opened 2026-08-27
fix(apps): a vendo_make EDIT hands back the builder's own words, not the app's name

yousefh409:
## What was wrong

The `vendo_make` create arm has relayed the assembling agent's own closing
sentence since the front door stopped composing one out of the app's title —
`make-receipt.ts` law 2: *"only the thing that built the screen knows what is on
it… A sentence composed here from the app's name alone gave the calling agent a
title and no facts, and it invented the rest."*

**The edit arm never got that fix.** Three places, in order:

- `packages/apps/src/server/persistence/edit-journal.ts:393` — `assembleEdit`
  drops `ScreenOutcome.say` on the floor.
- `packages/apps/src/server/runtime/types.ts:260` — `EditResult` has nowhere to
  put it.
- `packages/apps/src/server/doors/make-tool.ts:270` — so the receipt is composed
  from the name: `"<name> is updated."` when it landed,
  `"I couldn't make that change to <name>."` when it did not. The floor's own
  reason was already computed and carried as far as `EditResult.issues`, and was
  discarded at the door too.

## What it cost

Captured verbatim off a real TaxDome turn (2026-08-27), from the tool-result
stream the calling agent reads:

```json
{"id":"app_3002…","title":"Firm ops","status":"failed",
 "say":"I couldn't make that change to Firm ops."}
```

Told only that, the agent told the person *"The document-collection portal
itself is still intact"* — over a stage-by-assignee table with no document
tracking on it at all. Six consecutive runs on that host had narration and DOM
disagree 6 out of 6 times.

## The fix

`say` travels the edit path the same way it travels the create path, and a
refused edit says why in the floor's own sentences. Same real turn, after:

```json
{"status":"failed",
 "say":"I couldn't make that change to Firm dashboard — line 72: prop
        \"columns\" on <DataTable> takes a list of rows, but this value is a
        list of rows — bind a value whose type matches the prop; …"}
```

(That refusal contradicting itself is a separate, already-fixed bug — #1662.
This change is what makes it visible to the caller at all.)

## Proof

- New seam test in `packages/vendo/tests/screen-route.e2e.test.ts` — a real edit
  through the real store, real checks floor, real paint and the real front door,
  read back off the receipt the calling agent actually got. Nothing stubbed on
  either side; only the model's words are scripted, as everywhere else in that
  file.
- Negative control: with the door change reverted, the new test fails with
  `expected 'Spending is updated.' to be 'The month's total is on it now…'`.
- Live: the two JSON receipts above are from a real browser turn against a real
  TaxDome host with a real model, before and after.

---
## Summary by cubic
Fixes the `vendo_make` EDIT arm so it hands back the builder's own words instead of a sentence composed from the app's name. Landed edits previously answered `"<name> is updated."` and refused ones `"I couldn't make that change to <name>."`; a calling agent handed only a title filled in the facts itself, producing narration that disagreed with the screen. Now `ScreenOutcome.say` travels the edit path exactly as it does the create path, and a refused edit says why in the floor's own sentences.

- Threads `say` from the edit journal through `EditResult` and into the receipt, keeping the name-based fallback when the builder says nothing.
- Attaches the floor's refusal issues to the failure receipt instead of discarding them at the door.
- Adds an e2e seam test covering a landed and a refused edit through the real store, floor, and front door.

<sup>Written for commit dcbc341fa9870b4a146bfbf769bef4aeaa245add. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/runvendo/vendo/pull/1674?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/edit-receipt-carries-the-builders-words.md
+++ b/.changeset/edit-receipt-carries-the-builders-words.md
@@ -1,0 +1,20 @@
+---
+"@vendoai/apps": patch
+---
+
+A `vendo_make` EDIT hands back the builder's own words, not the app's name.
+
+The create arm has relayed the assembling agent's closing sentence since the
+front door stopped composing one out of the app's title. The edit arm never did:
+every landed change answered `"<name> is updated."` and every refused one
+`"I couldn't make that change to <name>."` — a title and no facts. The floor's
+own reason for the refusal was already computed and carried as far as
+`EditResult.issues`, then dropped at the door.
+
+Handed a title and no facts, a calling agent invents the rest. Across six
+consecutive runs on a demo host, a chat agent told the person a per-client
+document tracker was "still intact" over a stage-by-assignee table it had never
+seen, and reported features that were never on the screen.
+
+`ScreenOutcome.say` now travels the edit path the same way it travels the create
+path, and a refused edit says why in the floor's own sentences.
--- a/packages/apps/src/server/doors/make-tool.ts
+++ b/packages/apps/src/server/doors/make-tool.ts
@@ -267,13 +267,25 @@ const changeExistingApp = async (
   if (result.automation !== undefined && stream !== undefined) {
     publishAutomationCard(stream, result.automation);
   }
+  // THE BUILDER'S OWN WORDS on this arm too (`make-receipt.ts` law 2). The create
+  // arm has relayed them since the front door stopped composing from the app's
+  // name alone; this one never did, so every landed edit answered
+  // "<name> is updated." and every refused one "I couldn't make that change to
+  // <name>" — a title and no facts, which is precisely what the calling agent
+  // invents around. Live 2026-08-27 (TaxDome): told only that, it reported a
+  // per-client document tracker "still intact" over a stage-by-assignee table.
+  // The refusal keeps its own sentence — the floor's lines say what is wrong with
+  // the change, and the builder's last words describe the screen it did not
+  // change.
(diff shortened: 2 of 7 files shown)


yousefh409:
Triaged — real observation, but out of scope for this PR, and not a regression it introduces.

`issues` is already documented on `main` as **"Caller-facing sentences"** (`runtime/types.ts:328`), and the two raw rungs cubic names — `edit-journal.ts:378` and `:399` on `main` — are pre-existing; this PR does not add or touch them (its only edit-journal change threads `say` through the `assembled` arm). So the raw-`safeErrorMessage` rung is a pre-existing gap between that contract and two throw paths.

Weighed against what the PR fixes: the status quo is a refusal carrying a title and **no facts**, which is precisely what made a calling agent invent a "still intact" document tracker over a table it had never seen, six runs running. A raw internal sentence is strictly more truthful than an invented one.

Curating those two throw paths is a real follow-up, but it is a behaviour change in `edit-journal`'s failure classification and belongs in its own change rather than riding a receipt fix. Logged for follow-up.

PR #1677, opened 2026-08-27
fix(vendo): the build brief tells the truth about the box it runs in

yousefh409:
Escalated builds stopped finishing. Four in a row died at **15.2–15.4 min** against the 15-minute message budget, on asks as small as *"show a QR code"* — so ask weight was never the variable.

The cause was not capacity, not the model, and not Vendo Cloud: a trivial box turn answers in **13.5s** on the same account. It was the brief. A build box is told to *"test what you built against reality"* and to reach host data through `callHost`, and it was never told which reality it is standing in. It reaches `registry.npmjs.org` and the inference host, and nothing else.

So an agent that took the brief seriously went looking for the reality it knows. Three separate dead ends, each measured from the box's own transcript — captured by temporarily restoring the events the lane discards at `build-agent.ts:184` (`emit: () => undefined`), which is why this was invisible for four runs:

1. **A browser, then a native canvas.** *"No network access for browser binaries… Node-canvas isn't installable here."* It re-architected the app between attempts.
2. **The baked `@vendoai/ui`.** `/opt/vendo-box` ships a copy that predates the frame protocol, so the build spent the time discovering that and switching to npm.
3. **A host tool list.** `callHost` was named as the way to reach host data, so it hunted `/app/.vendo` and `/opt/vendo-box/template/.vendo` for schemas that are deliberately absent — this box gets no `toolDoor` (`build-agent.ts:179`).

### The change

The brief now names the egress it lives inside (Node, a pure-JS DOM, no browser, no native binaries), says to install `@vendoai/ui` from npm rather than the stale baked copy, and puts `startFrameProtocol`/`callHost` on the **runtime** side of the line: they speak to the embedding page, so they answer for the shipped app and never inside the box.

**Nothing is withdrawn.** The instruction to verify stands — shipping an untested bundle is the failure this brief already guards against. `callHost` stays: it is a `postMessage` to the host page, and dropping it would take the capability with it. The egress allowlist is untouched because it is the security boundary. `MESSAGE_BUDGET_MS` is unchanged because a timeout is a hang-detector, not a speed limit.

### Proof — the real path

`appBuilder` itself, its own brief, its own egress, its own 15-minute budget, against a real Vendo Cloud box.

**BUILT in 836.6s**, sealing `dist/app.js` at **257,426 bytes**, alongside `src/App.tsx`, `src/QrCode.tsx`, `src/index.tsx` and `package-lock.json`. The first completed escalated build since 2026-08-24.

Progression across runs, all live:

| | outcome |
|---|---|
| before any fix | 910s, 646s, 905s — budget exhausted / box vanished |
| after this PR | **836.6s — built and sealed** |

### Known risk, stated plainly

836.6s against a 900s budget is **~64s of headroom**, and observed run-to-run spread is wide. A slower run can still exhaust the budget. This PR does not change that budget, and should not.

Related: run 1 above independently confirmed #1676 on the live lane — the failure reads `the build ran out of its time budget`, `retryable=false`, instead of "busy, try again shortly".

### DIFF
--- a/.changeset/build-brief-names-its-reality.md
+++ b/.changeset/build-brief-names-its-reality.md
@@ -1,0 +1,24 @@
+---
+"@vendoai/vendo": patch
+---
+
+The build brief tells the box what "reality" is, so builds stop hunting for a browser.
+
+A build box is told to "test what you built against reality, and fix what
+fails" — and it was never told what reality it is standing in. The machine
+reaches the npm registry and the inference host, and nothing else. So an agent
+that took the instruction seriously went looking for the reality it knows: a
+real browser to drive Playwright against, then a native canvas to render into.
+Both are unreachable past the box's allowlist, and neither failure is one it can
+be talked out of, so it re-architected the app and tried again.
+
+Measured live on 2026-08-27: four escalated builds died at 15.2–15.4 minutes
+against the 15-minute message budget, on asks as small as "show a QR code". Ask
+weight was never the variable — every build reached the same cul-de-sac, and the
+cul-de-sac costs the same whatever was asked for.
+
+The test step now names the egress it has to live inside: Node, a pure-JS DOM,
+no browser and no native binaries. The instruction to verify is unchanged — a
+build that ships an untested bundle is the failure this brief already guards
+against — and the allowlist is untouched, because it is the security boundary
+and the honest fix is to stop lying to the agent about it.
--- a/.changeset/build-brief-names-its-reality.md
+++ b/.changeset/build-brief-names-its-reality.md
@@ -17,8 +17,23 @@ against the 15-minute message budget, on asks as small as "show a QR code". Ask
 weight was never the variable — every build reached the same cul-de-sac, and the
 cul-de-sac costs the same whatever was asked for.
 
-The test step now names the egress it has to live inside: Node, a pure-JS DOM,
-no browser and no native binaries. The instruction to verify is unchanged — a
-build that ships an untested bundle is the failure this brief already guards
-against — and the allowlist is untouched, because it is the security boundary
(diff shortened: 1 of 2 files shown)


yousefh409:
Reviewer triage before merge — both P2s dismissed, reasons below.

**cubic, `build-agent.ts:160` — "nothing with a native binary will build" vs. "Bundle it with esbuild".**
Technically sharp, and dismissed on evidence. The sentence says nothing native will *build* — i.e. compile against system libs, which is the node-canvas dead end that caused the stall. esbuild is a prebuilt binary fetched from the registry, so it is not "built" and the brief does not exclude it.

The decisive part is empirical, not semantic: the 836.6s run that completed and sealed a 257 KB `dist/app.js` read **this exact wording** and reached for esbuild without hesitating. The rabbit hole this comment predicts is the one thing we actually measured not happening. Rewording now would replace proven brief text with unproven brief text and invalidate the only evidence this PR rests on — the merged brief must be the brief that was proved. Not worth a re-proof cycle for a hypothesis the run already contradicts.

**greptile, changeset vs. PR description.**
Correct when written, now stale — and it called its own resolution. The 836.6s run was measured *after* both the `@vendoai/ui`-from-npm and `callHost`-runtime-side additions, exactly as the comment supposed; the description has since been rewritten and now agrees with the changeset. Nothing to walk back: the changeset claim is the verified one, and the changeset is the only one of the two that reaches the changelog.

Merging. The headroom risk the description states plainly — 836.6s against a 900s budget, ~64s of margin, wide run-to-run spread — is real and is not closed by this PR. It is a large improvement over a 100% failure rate, not a guarantee, and the remaining budget question stays open.

PR #1680, opened 2026-08-27
feat(storage): one SQL database per app, with mine./shared. as the whole permission model

yousefh409:
## One SQL database per app

A generated app keeps its data in a real SQL database of its own, reached through
ONE agent tool — `vendo_apps_sql` — which runs one statement and whose description
states the live dialect. Two table namespaces are the entire permission model:

- **`shared.<table>`** — one table, all of the app's users.
- **`mine.<table>`** — ordinary SQL, per user.
- A bare `CREATE TABLE x` is refused with what happened · why · fix:
  `"x" is not a table this app can reach. Every table lives in shared. (all users) or mine. (per-user). Did you mean mine.x?`

New adapter slot `createVendo({ appDatabase })`, standard adapter rule: an
explicitly passed adapter always wins. Unset, every app gets its own fenced
schema inside the Postgres the host already wired — **zero new configuration**.
A store with no SQL handle composes no adapter and the tool is not offered.

## Why the fence is name resolution and not a database privilege

`mine.` is enforced platform-side, at the door, never by generated SQL — which is
treated as hostile input throughout. It deliberately does **not** rest on Postgres
RLS: a BYO `DATABASE_URL` is very often a superuser or the table owner, and
`FORCE ROW LEVEL SECURITY` is bypassed by both. A boundary that evaporates when
the operator's role is too strong is not a boundary.

Four independent layers, each holding alone:

1. `mine.x` / `shared.x` are the ONLY table addresses that exist. They become
   physical names carrying a `":"` — a character no identifier the grammar admits
   can contain — so a physical table has no spelling inside an app's SQL, and one
   person's tables have no spelling in another's.
2. Every statement runs with `search_path` set to the app's own schema, so an
   unqualified name resolves inside the app or **nowhere**.
3. Any qualifier that is not `mine`/`shared` is checked against `pg_namespace`:
   table aliases pass, `public.vendo_records` does not.
4. One statement, from a short allowlist, with no session, role, schema or catalog
   verb anywhere in it.

`mine.x` is a **separate physical table per person**, not a predicate over a shared
one. That is what keeps ordinary SQL ordinary — a `PRIMARY KEY` is unique per
person, a `UNIQUE` is per person — and it means the whole delete-then-write-by-key
class (`REPLACE INTO`, `MERGE`, upsert on someone else's key) has no reachable
target across the fence. One per-app DDL log, replayed lazily per person, keeps
everyone's schema identical.

## Deleted, whole

`vendo_apps_data_list` / `_put` / `_delete`; `AppDocument.storage` and
`StorageDecl`; the `declaredStorage` allow-list gate; `createAppData` /
`AppDataAccess` (records and blobs); `StoreOps["appData"]` and `AppDataTarget`;
the `app:<id>:<collection>` namespaces; the 256 KB record and 5 MB file caps; the
app-data owner backfill. Migration is fix-forward — chat-built apps could never
save through the old path, and its blob half had no callers.

Three isolation hazards die with it, confirmed gone:

- the unowned façade that gave every user one shared drawer on a store with no
  ops surface;
- `hostedStore`'s `owner: "user_local"` default, which put a whole multi-user
  deployment in one drawer with nothing refusing it;
- the un-allowlisted `ops.blobs` namespace, which let a caller write into another
  owner's app files.

The `appData.*` wire paths keep **RETIRED slots** in `STORE_WIRE_PATHS`: removing
eight entries from the middle would silently shift `STORE_WIRE_APPEND_MESSAGES_OPS`
and `STORE_WIRE_TURN_OPS` onto ops that are not those ops — the console #468 shape
the warning in that file was written about.

## Proof

- **Real Postgres 16** (docker), through the **real agent tool handler**, nothing
  stubbed on either side: 28/28. Includes a fresh-process read-back and a negative
  control — a real, populated `public.host_probe` table reachable by the same
  credentials is refused through every address (`public.x`, bare, UNION, CTE,
  scalar subquery).
- **In-suite seam test** `packages/store/tests/app-database.test.ts` — 27/27 over a
  real embedded Postgres. It cannot skip itself, unlike the `app-data-put-race`
  test it replaces.
- **Adversarial guard suite** `packages/apps/tests/security/app-sql-guard.test.ts`
  — 30/30.

## Scope

BYO Postgres only. The Cloud adapter and its wire route are the next lane.

## Decisions recorded (three-way tandem with vendo-web #670 + the app-data retirement)

**Cloud wire path** is `/api/v1/store/app-db/{appId}/{run,tables,drop}` — under the
store mount, so it inherits its auth, envelope, body caps and existing zone route.

**`drop()` holds an ADMIN key.** Not a new requirement: `apps-surface.delete` calls
`sql.drop(appId)` and `ops.lifecycle.erase({ appId })` thirteen lines apart on the
same key, and erase is already admin-only — so app deletion has always needed an
admin-class credential. `run` takes the write cap (the door cannot tell SELECT from
DELETE without a parser, and deliberately has none).

**A failed statement must answer 400 with SQLite's exact text** through the
`DataPlaneError` `validation` tunnel. This is load-bearing here, not cosmetic:
`app-sql.ts`'s `explain()` parses that text to produce "Did you mean mine.todos?".
Wrapped into a generic 503 it degrades to the raw engine error — pinned by
`app-database.test.ts`'s workerd-message cases.

**app-db is deliberately NOT in `STORE_WIRE_PATHS`.** That table is the HTTP profile
of the *StoreOps* contract and its `/status` count is a frozen LEVEL; `AppDatabase`
is a different adapter, not a StoreOps family, so appending it would make the level
index ops that are not StoreOps ops — the exact semantics this PR restores by
keeping the eight retired `appData.*` slots. app-db therefore sits outside the
mount-parity guarantee, which is safe because a spelling drift between the two
halves 404s loudly rather than failing silently. If it is ever wanted inside that
guarantee, it needs its own path constant, never an append to `STORE_WIRE_PATHS`.


yousefh409:
## Merge-lane triage — rebased onto `main`, **not merged**: two blocking findings

Rebased `feat/app-database` onto `main` (`2779dbd69`) and pushed as `dc50b33f6`. The PR is now `MERGEABLE` and CI is green so far (17 checks passed, none failing). I am **not** merging, because two of the AI reviewers' P1s are real and I verified both from source.

---

### BLOCKING 1 — the SQL guard's reserved-name check is bypassed by double-quoting

`packages/apps/src/server/persistence/app-sql-guard.ts` has two identifier branches that do not share validation:

- bare identifiers, **line 187**: `if (DENIED_PREFIX.test(word) || DENIED_WORDS.has(word)) refuse(...)`
- double-quoted identifiers, **lines 162-173**: the only gate is `if (!NAME.test(inner))` at line 165, then `push("quoted", ...)` at line 171.

`DENIED_PREFIX` and `DENIED_WORDS` are referenced exactly once in the tree, at line 187. So `query_to_xml` — which is in `DENIED_WORDS` at line 53 and matches `NAME` (`/^[a-z][a-z0-9_]{0,27}$/`) — is refused unquoted and **admitted quoted**:

```sql
SELECT "query_to_xml"('SELECT * FROM "m:<other-persons-digest>:notes"', true, true, '')
```

`query_to_xml` is `PUBLIC`-executable arbitrary SQL — no superuser needed. The inner statement is a `'…'` literal, swallowed opaquely by `tokenize` at lines 156-161, so the `m:` fence characters are never inspected. `"pg_tables"` is admitted the same way and enumerates every physical `m:<digest>:<table>` name, which is how you'd obtain the digest.

This defeats the `mine.`/`shared.` boundary that is the PR's entire permission model. Defence in depth does not cover it: `search_path` (`packages/store/src/app-database.ts:46`) cannot help, because Postgres implicitly searches `pg_catalog` regardless — the comment at `app-database.ts:44-45` says "the guard refuses every `pg_` identifier", which is true only for the bare branch. The connection also runs as the host's own store role (`app-database.ts:34-36`); there is no `SET ROLE` anywhere in `packages/store/src`.

Tests miss it precisely: `packages/apps/tests/security/app-sql-guard.test.ts:100-112` tests `pg_read_file`, `query_to_xml`, `set_config`, `current_setting` — **all bare**. Adding two quote characters defeats every one.

Fix is the one cubic proposed: apply `DENIED_PREFIX`/`DENIED_WORDS` to `inner` before line 171.

*(cubic's own example, `"_vendo_owner"`, is wrong — `NAME` already rejects a leading underscore. The severity is understated, not overstated.)*

---

### BLOCKING 2 — erase no longer removes an app's own data

- `eraseStore().bySubject()` (`packages/store/src/erase.ts:216-219`) enumerates owned apps and calls `eraseAppScoped` (`erase.ts:176-197`), which deletes only `vendo_records`, `vendo_app_grants`, `vendo_quarantine`. No SQL-database call; `eraseStore` never receives an `AppDatabase`.
- `eraseStore().byApp()` / `ops.lifecycle.erase({ appId })` (`erase.ts:298-333`, `packages/store/src/ops.ts:886-891`) delete the app row but never call `drop`. `createStoreOps` has no app-database parameter (`ops.ts:278`), so the ops surface structurally cannot drop a schema.
- Exhaustive grep: the only production caller of `drop` is `packages/apps/src/server/doors/apps-surface.ts:205`. `AppSqlAccess.forget` — declared at `app-sql.ts:117` as *"Erase cascade, subject leg — every `mine.` table this person holds in this app"* — has **zero production callers**. It was written, documented as part of the cascade, and never wired.

The comment at `erase.ts:172-175` asserts the apps runtime drops the schema, but that call goes the other way: `apps-surface.ts:218` calls *into* `ops.lifecycle.erase`.

This is a regression, not a pre-existing gap. The old `eraseAppData` removed the app's records and blobs from both cascades and had a purpose-built leg for a departing member's data inside a promoted **org** app (`vendo_blobs … namespace LIKE 'app:%'`, removed in `d4ac0a0b0`). What survives a `bySubject` now is an orphaned, permanently unreachable Postgres schema holding the person's actual rows.

The coverage that would have caught it was removed in the same commit: the `bySubject`/`byApp` assertions in `packages/store/tests/erase.conformance.test.ts`, and the `appData` assertion inside the core conformance case still named *"lifecycle.erase removes one app's data, and only that app's"* (`packages/core/src/conformance/store-ops.ts:1440`). The only tests touching `drop`/`forget` (`packages/store/tests/app-database.test.ts:398-408`) call the adapter methods directly and prove nothing about any erase path.

Note the org-app half is not a one-line fix: `SELECT id FROM vendo_apps WHERE subject = $1` (`erase.ts:216`) does not enumerate apps the subject merely *used*.

---

### Dismissed

- **`format-reference.ts:194` — "This teaches the wrong hook shape" (cubic P1): incorrect, and acting on it would break every generated screen.** `useQuery` returns `data[key]` verbatim on a hit (`packages/apps/src/contract/genui/component/vm-program.ts:581`), and the host stores the tool result unwrapped: `queries[queryKey(entry)] = await opts.runQuery(...)` (`packages/apps/src/server/checking/component-screen.ts:689`). `MISS = { data: undefined }` (`vm-program.ts:575`) is the *pending-read sentinel*, not a react-query wrapper on answers. So `pets.rows` is right and `pets.data?.rows ?? []` would render empty forever. The file as written is correct.
- **`compose-apps.ts` / `compose-context.ts` / `compose-surfaces.ts` — `vendo_apps_sql` missing from the registry (cubic P1 ×3, greptile P2):** addressed by `355cafdfa` ("compose the SQL tool where the dialect is actually known"); GitHub has since marked two of the three outdated.
- **`app-sql.ts` aggregate-read-returns-empty (P3), `CREATE TABLE IF NOT EXISTS` at the table cap (P3), concurrent `CREATE TABLE mine.x` retry loop (greptile P1), mixed shared/mine DDL replay (P1), schema-qualifier TOCTOU (P1), `args: null` dereference (P2 ×2):** all real-shaped but they are behaviour *inside a feature that did not previously exist*, not regressions against a path anything used — per the changeset, chat-built apps could never save through the old path at all. Follow-up material, not merge gates. The concurrent-`CREATE TABLE` one is the strongest of them and deserves its own issue.
- **`app-first-build.seam.test.ts:190` (cubic P2) — the negative control:** fair. The stranger's query runs after `settleBuild` has cleared `buildsInFlight`, so it cannot distinguish a subject-scoped `buildingFor` from an unscoped one. The positive direction was verified by reverting the fix, so the exemption itself is proven; the isolation property is not. Worth tightening with the rest.
- **`vendo_apps_sql` has no `outputSchema`:** knowingly deferred — it touches the settled tool descriptor and is a separate decision.
- Remaining P2/P3 style and doc-wording notes: not merge gates.

---

The rebase itself needed five conflict resolutions, all mechanical except one worth flagging: `packages/mcp/src/shim/shim-html.gen.ts` is generated and **both** sides had regenerated it (main via #1667's `VENDO_CONSOLE_URL`/`VENDO_BASE_URL` rename), so neither side's copy was the bundle of the merged tree. Regenerated from source rather than hand-merged (`dc50b33f6`).

yousefh409:
### CI is red on `typecheck-lint` — one line, unrelated to the rebase

```
tests/screen-route.test.ts(111,43): error TS2345:
  Property 'buildingFor' is missing in type '{ screen?; requireOwned; claimSlot; markUnbuilt }'
  but required in type 'AgentToolsDataDependencies'.
```

`buildingFor` became a required member of `AgentToolsDataDependencies` (`packages/apps/src/server/doors/agent-tools.ts:234`) in the mid-build-exemption commit. The deps literal at `packages/apps/tests/screen-route.test.ts:111` was not updated with it. Nothing to do with the rebase — the file is untouched by it, and the same error reproduces on the pre-rebase tip.

Leaving it to the author rather than editing a test file from the merge lane. Everything else CI has finished is green: `audit`, `conformance`, all five `integration` legs, `changeset-required`, `cone`, `gitleaks`, `opengrep`, and the `store`/`ui` shards.

